PR #17, opened 2020-10-29
Use `packageJson`

ehmicky:
This adds the new `packageJson` argument passed to plugin handlers, which helps simplifying code and handle any edge case (for example, the current code did not work when using a `base` directory with a `package.json` several directories above it).

Note: you will need to upgrade Netlify Build to `5.1.0` (the version is shown in build logs) to try this locally. Until we make a release of the CLI, in the repository which uses `netlify-cli`, this can be done by removing your `package-lock.json` and `node_modules`, then running `npm install` again. If installed globally, this can be done by re-installing `netlify-cli` globally.

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,7 +15,7 @@ const isStaticExportProject = require('./helpers/isStaticExportProject');
 // - Between the build and postbuild steps, any functions are bundled
 
 module.exports = {
-  async onPreBuild({ constants, netlifyConfig, utils }) {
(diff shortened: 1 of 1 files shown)


lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/15

lindsaylevine (approved):
ty for this!!!!!!!

Issue #8, opened 2020-10-27
Plugin load if `next` is not available


ehmicky:
`next-on-netlify` requires `next` as a peerDependency:

https://github.com/netlify/next-on-netlify/blob/7d1fd5e08969455035bf9918a6fb2d707984c4a1/package.json#L37-L39

This repository requires `next-on-netlify` as a production dependency:

https://github.com/netlify/netlify-plugin-nextjs/blob/ef4529016ed0f87ae89d5f312dc8f8f08d1ef7b6/package.json#L24-L29

If a site is using this plugin, but forgot to install `next`, this plugin will thrown when being `require()`'d, as opposed to when `onPreBuild` is run. This results in an error message which might be confusing:

```
Uncaught Error: Cannot find module 'next/constants'
Require stack:
- /home/ether/Desktop/node_modules/next-on-netlify/lib/helpers/getNextDistDir.js
- /home/ether/Desktop/node_modules/next-on-netlify/lib/config.js
- /home/ether/Desktop/node_modules/next-on-netlify/lib/steps/prepareFolders.js
- /home/ether/Desktop/node_modules/next-on-netlify/index.js
- /home/ether/Desktop/node_modules/@netlify/plugin-nextjs/index.js
- <repl>
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:900:15)
    at Function.Module._load (node:internal/modules/cjs/loader:745:27)
    at Module.require (node:internal/modules/cjs/loader:972:19)
    at require (node:internal/modules/cjs/helpers:88:18) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/home/ether/Desktop/node_modules/next-on-netlify/lib/helpers/getNextDistDir.js',
    '/home/ether/Desktop/node_modules/next-on-netlify/lib/config.js',
    '/home/ether/Desktop/node_modules/next-on-netlify/lib/steps/prepareFolders.js',
    '/home/ether/Desktop/node_modules/next-on-netlify/index.js',
    '/home/ether/Desktop/node_modules/@netlify/plugin-nextjs/index.js',
    '<repl>'
  ]
}
```

To fix this: 
  - `next` and `next-on-netlify` should be `require()'d` only inside the `if (isNextProject) { ... }` block
  - `_isNextProject()` should try to `require('next')`, and return `false` if that throws an error 

We might want to also wonder: are we ok with this plugin requiring users to install Next.js as a dependency (as opposed to doing it for them)?

[...]

PR #20, opened 2020-11-06
add some initial lightweight testing

lindsaylevine:
this also addresses https://github.com/netlify/netlify-plugin-nextjs/issues/8 and https://github.com/netlify/netlify-plugin-nextjs/issues/12

1. we see that exact error in issue #8 running tests if next isnt installed in the project. should i try to mock it? :/
2. seems we don't need hasFramework anymore with the next require try/catch. also when this is upgraded to a core plugin, won't the framework check be done outside of the plugin?
3. tried to mock `fs` in the standard way i mock the other dependencies and ran into a handful of issues. i now use the mock-fs package but calling mockFs broke the internals of `hasFramework` despite efforts to pass a correct config to mockFs that would appease hasFramework ☹️

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,8 +1,11 @@
-const { existsSync, readFileSync, appendFileSync } = require('fs')
+const fs = require('fs')
+const { promisify } = require('util')
+const exists = promisify(fs.exists)
 const path = require('path')
 const nextOnNetlify = require('next-on-netlify')
 const { PHASE_PRODUCTION_BUILD } = require('next/constants')
 const { default: loadConfig } = require('next/dist/next-server/server/config')
+const makef = require('makef')
 const makeDir = require('make-dir')
 const cpx = require('cpx')
 const isStaticExportProject = require('./helpers/isStaticExportProject')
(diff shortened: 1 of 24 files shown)


index.js
```diff
@@ -61,7 +53,7 @@ module.exports = {
             target: 'serverless'
           }
         `
-      await appendFile('next.config.js', nextConfig)
+      appendFileSync('next.config.js', nextConfig)
```
ehmicky, line 56: Note: if you want to keep it async while maintaining old Node versions (i.e. without `fs.promise`), you could also use:

```js
const { promisify } = require('util')
const appendFile = promisify(fs.appendFile)
```

index.js
```diff
@@ -15,13 +12,8 @@ const isStaticExportProject = require('./helpers/isStaticExportProject')
 // - Between the build and postbuild steps, any functions are bundled
 
 module.exports = {
-  async onPreBuild({ netlifyConfig, packageJson: { scripts = {}, dependencies = {} }, utils }) {
+  async onPreBuild({ netlifyConfig = {}, packageJson: { scripts = {}, dependencies = {} }, utils }) {
     const { failBuild } = utils.build
-
-    if (!(await hasFramework('next'))) {
```
ehmicky, line 21: Until this is made a core plugin, if a user is installing this plugin but they forgot to install `next`, wouldn't it be better to print this error message?

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "format": "prettier --write .",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "jest"
```
ehmicky, line 8: Should we also add some GitHub actions, so tests are run when someone creates a PR?

package.json
```diff
@@ -23,19 +23,27 @@
   },
   "homepage": "https://github.com/netlify/netlify-plugin-nextjs#readme",
   "dependencies": {
-    "@netlify/framework-info": "^0.2.0",
     "cpx": "^1.5.0",
     "make-dir": "^3.1.0",
+    "next": "^10.0.1",
```
ehmicky, line 28: What if the site is using a different version of next, e.g. `9.0.0`?
In that case, this plugin would assume the site is using a different version.

test/index.js
```diff
@@ -0,0 +1,140 @@
+const path = require('path')
```
ehmicky, line 1: Nice! :clap:

package.json
lindsaylevine: oof true

lindsaylevine: this version basically just needs to be aligned with next-on-netlify, which hardcodes its version as  `"next": "9.x"` 

in the future NoN will likely just live in here and so we won't have to manage in multiple places

lindsaylevine:
now fixes https://github.com/netlify/netlify-plugin-nextjs/issues/12

PR #31, opened 2020-11-13
Fix deprecated method

### LINKED ISSUE #30: Remove usage of deprecated `fs.exists()`
`fs.exists()` is [deprecated](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback).
We should use alternatives such as [`path-exists`](https://github.com/sindresorhus/path-exists) instead.

ehmicky:
Fixes #30.

`fs.exists()` is [deprecated](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback).
We should use alternatives such as [`path-exists`](https://github.com/sindresorhus/path-exists) instead.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,12 +1,10 @@
-const fs = require('fs')
-const { promisify } = require('util')
-const exists = promisify(fs.exists)
 const path = require('path')
 const nextOnNetlify = require('next-on-netlify')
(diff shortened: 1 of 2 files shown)


lindsaylevine (approved):
noted!!

ehmicky:
To be fair, considering how careful Node.js is about breaking changes, it is possible this deprecated method will never be removed :D

PR #34, opened 2020-11-13
Use test fixtures

ehmicky:
I am hitting two issues with `mock-fs`: [here](https://github.com/tschaub/mock-fs/issues/239) and [there](https://github.com/tschaub/mock-fs/issues/213).

When I try to add more tests, I get:

```
RUNS  test/index.js
node:fs:1732
      handleErrorFromBinding(ctx);
      ^

Error: ENOENT: no such file or directory, lstat '/home/ether/netlify/netlify-plugin-nextjs/node_modules/jest/node_modules/chalk'
    at Object.realpathSync (node:fs:1732:7)
    at toRealPath (node:internal/modules/cjs/loader:368:13)
    at tryFile (node:internal/modules/cjs/loader:364:10)
    at tryExtensions (node:internal/modules/cjs/loader:376:22)
    at tryPackage (node:internal/modules/cjs/loader:322:5)
    at Function.Module._findPath (node:internal/modules/cjs/loader:540:18)
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:887:27)
    at Function.Module._load (node:internal/modules/cjs/loader:745:27)
    at Module.require (node:internal/modules/cjs/loader:972:19)
    at require (node:internal/modules/cjs/helpers:88:18) {
  syscall: 'lstat',
  code: 'ENOENT',
  errno: -2,
  path: '/home/ether/netlify/netlify-plugin-nextjs/node_modules/jest/node_modules/chalk'
}
```

Apparently, the issue is the following: after filesystem has been mocked, `jest` lazily try to `require('chalk')` by using `fs` instead. But because mocking is in place, this crashes.

This new lazy loading behavior was added to `mock-fs@4`. I tried downgrading to `mock-fs@3` but an error message shows me it won't run on Node `15`.

This PR implements an alternative: using a fixtures directory with static files committed to `git`, then do `process.chdir()` inside tests that need filesystem access.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -1,9 +1,9 @@
 const path = require('path')
+const process = require('process')
 const nextOnNetlify = require('next-on-netlify')
 const makef = require('makef')
 const makeDir = require('make-dir')
 const cpx = require('cpx')
-const mockFs = require('mock-fs')
 const plugin = require('../index')
 
 const utils = {
@@ -32,6 +32,14 @@ jest.mock('cpx')
 console.log('Initializing tests')
 
 const DUMMY_PACKAGE_JSON = { name: 'dummy', version: '1.0.0' }
+
+const FIXTURES_DIR = `${__dirname}/fixtures`
+
+const useFixture = function (fixtureName) {
+  const originalCwd = process.cwd()
+  process.chdir(`${FIXTURES_DIR}/${fixtureName}`)
+  return process.chdir.bind(process, originalCwd)
+}
 
 describe('preBuild()', () => {
   test('fail build if the app has static html export in npm script', async () => {
@@ -98,11 +106,7 @@ describe('preBuild()', () => {
   })
 
   test(`fail build if the app's next config has an invalid target`, async () => {
-    mockFs({
-      'next.config.js': {
-        target: 'nonsense',
-      },
-    })
+    const restoreCwd = useFixture('invalid_next_config')
 
     await plugin.onPreBuild({
       netlifyConfig: {},
@@ -111,7 +115,7 @@ describe('preBuild()', () => {
       constants: { FUNCTIONS_SRC: 'out_functions' },
     })
 
-    mockFs.restore()
+    restoreCwd()
 
     const acceptableTargets = ['serverless', 'experimental-serverless-trace']
     expect(utils.build.failBuild.mock.calls[0][0]).toEqual(
--- a/test/index.js
+++ b/test/index.js
@@ -1,10 +1,13 @@
 const path = require('path')
 const process = require('process')
+
 const nextOnNetlify = require('next-on-netlify')
-const makef = require('makef')
-const makeDir = require('make-dir')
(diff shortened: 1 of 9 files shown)


lindsaylevine (approved):
whoa this is clever, very awesome !!!!!!!

thought: i was originally using appendFile to create the next config if it doesn't exist in the project so in this test `'create next.config.js with correct target if file does not exist'` i first tried using mock-fs or trying to mock appendFile and it was such a pain that i switched to make-file. would this fixture approach work for mocking appendFile and if so, how?

ehmicky:
Good point!
I have added a new commit to use a temporary directory in each test instead, so that mocking is not necessary.
The less we mock, the closer we are to the production code, which will help increase our confidence in the tests.
Please let me know what you think!

test/index.js
```diff
@@ -15,32 +18,42 @@ const utils = {
...
+})
+
+afterEach(async () => {
   utils.build.failBuild.mockReset()
   utils.run.command.mockReset()
   jest.clearAllMocks()
   jest.resetAllMocks()
+
+  this.restoreCwd()
+  await this.cleanup()
```
lindsaylevine, line 50: what does cleanup do and why is it async?

lindsaylevine (approved):
this is so awesome, i def have a lot more to learn and a lot more experience to gain with node 🙏 thank you for doing this!!!!!!!

test/index.js
ehmicky: Added some comment :+1:

ehmicky:
Quick run-through of the setup:
  - Before each test, we create an empty temporary directory and change `cwd` to it (https://github.com/netlify/netlify-plugin-nextjs/blob/ea07d970ff35d5b73e1b588eaf2eb973977a6489/test/index.js#L37)
  - Since each test runs in this temporary directory, the code can now create/modify/delete files. Which means tests can now directory assert things like "Does this file exist?" (https://github.com/netlify/netlify-plugin-nextjs/blob/ea07d970ff35d5b73e1b588eaf2eb973977a6489/test/index.js#L159)
  - After each test, we remove the temporary directory and change `cwd` back to its original value (https://github.com/netlify/netlify-plugin-nextjs/blob/ea07d970ff35d5b73e1b588eaf2eb973977a6489/test/index.js#L49

Then, some tests might need to use a specific fixture directory instead of an empty temporary directory. To do this:
  - Fixture directories are in `test/fixtures/{fixtureName}`
  - Using `useFixture(fixtureName)` at the beginning of a test changes `cwd` to the fixture directory (https://github.com/netlify/netlify-plugin-nextjs/blob/ea07d970ff35d5b73e1b588eaf2eb973977a6489/test/index.js#L124)
  - It returns `restoreCwd` which can be called to restore the original `cwd` (which is the temporary directory) (https://github.com/netlify/netlify-plugin-nextjs/blob/ea07d970ff35d5b73e1b588eaf2eb973977a6489/test/index.js#L131)

PR #36, opened 2020-11-13
Fix `failBuild()` testing

ehmicky:
Under the hood, `utils.build.failBuild()` is throwing an error.
However, in tests, it is mocked with an empty function. This means the control flow continues because the code does not have any `return` statement after `failBuild()`. This means tests end up in odd situation where they are reaching lines of code that should never been hit in production.

This PR tries to fix this by:
  - Adding `return` statements after `utils.build.failBuild()`
  - Instead of mocking `utils.build.failBuild()`, replacing it with a simple function that throws an error. The tests then need to assert that an error is thrown from the `async onPreBuild` function, which is done in Jest with `await expect(onPreBuild(...)).rejects.toThrow('error message')`, as opposed to looking for mock calls.

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,8 +22,7 @@ module.exports = {
     const { failBuild } = utils.build
 
     if (Object.keys(packageJson).length === 0) {
-      failBuild(`Could not find a package.json for this project`)
-      return
+      return failBuild(`Could not find a package.json for this project`)
     }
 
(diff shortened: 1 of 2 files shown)


lindsaylevine (approved):
this is awesome, much to learn <3 a suggestion to be made possibly to nimbella since i took the mock failBuild logic from them 🌚

PR #37, opened 2020-11-13
Improve fixtures setup

ehmicky:
This PR improves the fixtures setup a little.
In the current setup, it is not possible to modify files when the current directory is a fixture directory, since we are not in the temporary directory anymore. This PR copies the files from the fixture to the temporary directory instead.

In a nutshell:
  - Before each test, we create a temporary directory and change `cwd` to it. That temporary directory is empty by default.
  - A test can use a fixture directory by calling `await useFixture(fixtureName)`. This will copy all the files from `./test/fixtures/{fixtureName}/**` to that temporary directory (used as `cwd`). In essence this create a "fake repository" before running the plugin.
  - After each test, we remove the temporary directory and change `cwd` to its original value.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -1,11 +1,15 @@
 const path = require('path')
 const process = require('process')
+const { promisify } = require('util')
 
+const { copy } = require('cpx')
 const nextOnNetlify = require('next-on-netlify')
 const pathExists = require('path-exists')
 const { dir: getTmpDir } = require('tmp-promise')
 
 const plugin = require('..')
+
+const pCopy = promisify(copy)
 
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -32,6 +32,7 @@
   },
   "devDependencies": {
     "@netlify/eslint-config-node": "^0.3.0",
+    "cpy": "^8.1.1",
```
lindsaylevine, line 35: does cpx (aka what we already had installed) not work for this use case?

lindsaylevine (approved):
1 question otherwise ⭐

package.json
ehmicky: Yes, it would. Fixed! :+1:

PR #38, opened 2020-11-16
Fix support for Node 8

### LINKED ISSUE #12: Node 8 support
Our Node.js support is currently `>=8`.

This plugin is using `fs.promises`, which was introduced in Node 10. Using `require('util').promisify(fs.appendFile)` instead would make it work for Node 8 users.

Also, do you know whether `next-on-netlify` works on Node 8?

It'd be nice to add some `engines.node` like [this](https://github.com/nimbella/netlify-plugin-nimbella/blob/0b38edd2df42e9bedd9377a99120af05c0523d35/package.json#L43). Our build system detects this property and prints a warning message when a site is using a Node.js version incompatible with a plugin.

Finally, our CI tests should run on Node 8 to ensure support.

ehmicky:
Fixes #12.

This adds tests for Node 8.
Jest has to be downgraded because the latest version does not support Node 8.
This seems to currently fail because `next-on-netlify` does not seem to support Node 8.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,12 +12,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        node-version: [10.x, 14.x]
+        node-version: [8.17.0, 14.x]
         exclude:
           - os: macOS-latest
-            node-version: 10.x
+            node-version: 8.17.0
           - os: windows-latest
-            node-version: 10.x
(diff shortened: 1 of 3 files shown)


ehmicky:
Next.js 9 does not support Node 8. More information in https://github.com/netlify/netlify-plugin-nextjs/issues/12#issuecomment-728294624

ehmicky:
Closing since we only need to support Node `10.13.0`.

PR #39, opened 2020-11-16
Use the site's Functions directory

### LINKED ISSUE #27: Do not configure to set `build.command`, `build.functions` and `build.publish`
In order to use this plugin, users have to set `build.command`, `build.functions` and `build.publish` in their `netlify.toml`.

We should instead re-use the configuration properties set the site:
  - If `build.functions` is defined, we should use this. If not, we should default to the default functions directory. This value can be retrieved using `constants.FUNCTIONS_SRC` (see #11)
  - Same for `build.publish`. This value can be retrieved using `constants.PUBLISH_DIR`.
  - The main reason we require setting `build.command` seem to be to ensure that `next-on-netlify` is run after `next build`. This should already be covered by this plugin considering it calls `next-on-netlify` programmatically during `onPostBuild`.

This would require updating `next-on-netlify` so it accepts custom functions and publish directories.

The main pros would be:
  - This allows users to define custom functions and publish directories
  - This removes those configuration steps, so we can eventually provide a 0-steps use of this plugin

ehmicky:
Fixes #27.

We just released in production the possibility for Build plugins to create a Functions directory even if the site does not have one. 
Thanks to this, we can now use `constants.FUNCTIONS_SRC` in order to use the site's Functions directory. 

It none was configured, the default Functions directory is used: this is named `netlify-automatic-functions` for the moment, but we will probably eventually rename it to `netlify-functions` (see [issue](https://github.com/netlify/build/issues/2016#issuecomment-726755971)).

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,12 +7,12 @@ const { PHASE_PRODUCTION_BUILD } = require('next/constants')
 const { default: loadConfig } = require('next/dist/next-server/server/config')
 const findUp = require('find-up')
 const makeDir = require('make-dir')
-const pathExists = require('path-exists')
-const cpx = require('cpx')
+const { copy } = require('cpx')
 
 const isStaticExportProject = require('./helpers/isStaticExportProject')
 
 const pWriteFile = util.promisify(fs.writeFile)
+const pCopy = util.promisify(copy)
 
 // * Helpful Plugin Context *
 // - Between the prebuild and build steps, the project's build command is run
@@ -73,26 +73,15 @@ module.exports = {
       console.log(`** Adding next.config.js with target set to 'serverless' **`)
     }
   },
-  async onBuild({ constants }) {
+  async onBuild({ constants: { PUBLISH_DIR, FUNCTIONS_SRC = DEFAULT_FUNCTIONS_SRC } }) {
     console.log(`** Running Next on Netlify package **`)
     nextOnNetlify()
 
     // Next-on-netlify puts its files into out_functions and out_publish
     // Copy files from next-on-netlify's output to the right functions/publish dirs
-
-    // TO-DO: use FUNCTIONS_DIST when internal bug is fixed
-    const { PUBLISH_DIR } = constants
-    // if (!(await pathExists(FUNCTIONS_DIST))) {
-    //   await makeDir(FUNCTIONS_DIST)
-    // }
-    const hasPublishDir = await pathExists(PUBLISH_DIR)
-    if (!hasPublishDir) {
-      await makeDir(PUBLISH_DIR)
-    }
-
-    // TO-DO: make sure FUNCTIONS_DIST doesnt have a custom function name conflict
-    // with function names that next-on-netlify can generate
-    // cpx.copySync('out_functions/**/*', FUNCTIONS_SRC);
-    cpx.copySync('out_publish/**/*', PUBLISH_DIR)
+    await makeDir(PUBLISH_DIR)
+    await Promise.all([pCopy('out_functions/**', FUNCTIONS_SRC), pCopy('out_publish/**', PUBLISH_DIR)])
   },
 }
+
+const DEFAULT_FUNCTIONS_SRC = 'netlify-automatic-functions'
--- a/test/index.js
+++ b/test/index.js
@@ -137,7 +137,7 @@ describe('onBuild()', () => {
   test('runs next on netlify', async () => {
     await plugin.onBuild({
       constants: {
-        PUBLISH_DIR: '',
+        PUBLISH_DIR: '.',
       },
     })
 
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -73,26 +73,15 @@ module.exports = {
...
     // Copy files from next-on-netlify's output to the right functions/publish dirs
-
-    // TO-DO: use FUNCTIONS_DIST when internal bug is fixed
-    const { PUBLISH_DIR } = constants
-    // if (!(await pathExists(FUNCTIONS_DIST))) {
-    //   await makeDir(FUNCTIONS_DIST)
-    // }
-    const hasPublishDir = await pathExists(PUBLISH_DIR)
-    if (!hasPublishDir) {
-      await makeDir(PUBLISH_DIR)
```
ehmicky, line 90: `make-dir` creates the directory if it does not exist.

index.js
```diff
@@ -73,26 +73,15 @@ module.exports = {
...
-    }
-
-    // TO-DO: make sure FUNCTIONS_DIST doesnt have a custom function name conflict
-    // with function names that next-on-netlify can generate
-    // cpx.copySync('out_functions/**/*', FUNCTIONS_SRC);
-    cpx.copySync('out_publish/**/*', PUBLISH_DIR)
+    await makeDir(PUBLISH_DIR)
+    await Promise.all([pCopy('out_functions/**/*', FUNCTIONS_SRC), pCopy('out_publish/**/*', PUBLISH_DIR)])
```
ehmicky, line 83: Parallelizing for better performance.

index.js
```diff
@@ -73,26 +73,15 @@ module.exports = {
...
   },
 }
+
+const DEFAULT_FUNCTIONS_SRC = 'netlify-automatic-functions'
```
ehmicky, line 87: We will need to change this value once the final filename has been decided.

test/index.js
```diff
@@ -155,7 +155,7 @@ describe('onBuild()', () => {
     expect(await pathExists(PUBLISH_DIR)).toBeTruthy()
   })
 
-  test('calls copySync with correct args', async () => {
+  test('copy files to the publish directory', async () => {
```
ehmicky, line 158: Some unused fixture files for this test were accidentally committed to `git` in a previous PR.

test/index.js
```diff
@@ -137,7 +137,7 @@ describe('onBuild()', () => {
   test('runs next on netlify', async () => {
     await plugin.onBuild({
       constants: {
-        PUBLISH_DIR: '',
+        PUBLISH_DIR: '.',
```
ehmicky, line 140: This is passed to `make-dir`. Using `.` is closer to production behavior than an empty string (which can never happen in production).

test/index.js
```diff
@@ -166,4 +166,19 @@ describe('onBuild()', () => {
 
     expect(await pathExists(`${PUBLISH_DIR}/subdir/dummy.txt`)).toBeTruthy()
   })
+
+  test.each([
+    { FUNCTIONS_SRC: 'functions', resolvedFunctions: 'functions' },
+    { FUNCTIONS_SRC: undefined, resolvedFunctions: 'netlify-automatic-functions' },
```
ehmicky, line 172: The first case is when a site configured a Functions directory. The second is when a site did not.

index.js
lindsaylevine: oh so it does nothing if it already exists? cool

index.js
```diff
@@ -73,26 +73,15 @@ module.exports = {
...
-    if (!hasPublishDir) {
-      await makeDir(PUBLISH_DIR)
-    }
-
-    // TO-DO: make sure FUNCTIONS_DIST doesnt have a custom function name conflict
-    // with function names that next-on-netlify can generate
-    // cpx.copySync('out_functions/**/*', FUNCTIONS_SRC);
-    cpx.copySync('out_publish/**/*', PUBLISH_DIR)
+    await makeDir(PUBLISH_DIR)
+    await Promise.all([pCopy('out_functions/**', FUNCTIONS_SRC), pCopy('out_publish/**', PUBLISH_DIR)])
```
lindsaylevine, line 83: 🖖

test/index.js
```diff
@@ -166,4 +166,19 @@ describe('onBuild()', () => {
...
+  ])('copy files to the functions directory', async ({ FUNCTIONS_SRC, resolvedFunctions }) => {
+    await useFixture('functions_copy_files')
+    await plugin.onBuild({
+      constants: {
+        FUNCTIONS_SRC,
+        PUBLISH_DIR: '.',
+      },
+    })
+
+    expect(await pathExists(`${resolvedFunctions}/function/index.js`)).toBeTruthy()
```
lindsaylevine, line 182: huge nit but could possibly be helpful for any devs new to the project:

could be helpful to have the fixture `functions_copy_files` be 

`out_functions/next_random/next_random.js`
and 
`out_functions/next_random/nextJsPage.js`

just to mimic exactly what NoN generates

lindsaylevine (approved):
🥇 ⭐ ❗

index.js
lindsaylevine: wait does this need to be defined before FUNCTIONS_SRC is defaulted to it on 76? 🤔

ehmicky: No because `onPostBuild()` is executed after the top-level scope has been executed.

test/index.js
ehmicky: Good point! Fixed.

PR #45, opened 2020-11-17
Make `next` a peer dependency

### LINKED ISSUE #26: Better error handling when the site does not use Next.js
If a site that does not use Next.js mistakenly uses this plugin, an error message should be printed.

ehmicky:
Fixes #26 and https://github.com/netlify/netlify-plugin-nextjs/issues/8
Also part of https://github.com/netlify/netlify-plugin-nextjs/issues/25

This makes `next` a peer dependency, which allows users to define which Next.js version they prefer (among the versions we support).
It a site is not using `next`, an error message is shown.
This also changes the version range of `next` from `^9.5.3` to `>=9.5.3` so that Next.js 10 users can use this, although support is experimental and we will print a warning message (to be done in separate PR).

Testing this turns out to be quite complicated. Making `require('next')` fail in automated tests is pretty hard to achieve without mocking `require()` which could bring its own sets of problems. Tests need `next` as a `devDependency` to run properly, so `require('next')` is guaranteed to succeed otherwise, even when the current directory changes (since `require()` is based on `__filename` instead). So I think we might need to leave out automated testing for this specific validation check (and also for the upcoming Next.js versions checks).

### DIFF
--- a/helpers/validateNextUsage.js
+++ b/helpers/validateNextUsage.js
@@ -1,0 +1,22 @@
+// Ensure Next.js is available.
+// We use `peerDependencies` instead of `dependencies` so that users can choose
+// the Next.js version. However, this requires them to install "next" in their
+// site.
+const validateNextUsage = function (failBuild) {
+  if (!hasPackage('next')) {
+    return failBuild(
+      'This site does not seem to be using Next.js. Please run "npm install next" or "yarn next" in the repository.',
+    )
+  }
+}
+
+const hasPackage = function (packageName) {
+  try {
+    require(packageName)
+    return true
+  } catch (error) {
+    return false
+  }
+}
+
+module.exports = validateNextUsage
--- a/index.js
+++ b/index.js
@@ -2,14 +2,12 @@ const fs = require('fs')
 const path = require('path')
 const util = require('util')
 
-const nextOnNetlify = require('next-on-netlify')
-const { PHASE_PRODUCTION_BUILD } = require('next/constants')
-const { default: loadConfig } = require('next/dist/next-server/server/config')
 const findUp = require('find-up')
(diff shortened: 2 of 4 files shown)


helpers/validateNextUsage.js
```diff
@@ -0,0 +1,22 @@
...
+const hasPackage = function (packageName) {
+  try {
+    require(packageName)
+    return true
+  } catch (error) {
+    return false
+  }
+}
+
+module.exports = { validateNextUsage }
```
lindsaylevine, line 22: curious why exported as object instead of just `module.exports = validateNextUsage`

index.js
```diff
@@ -46,6 +46,11 @@ module.exports = {
 
     const nextConfigPath = await findUp('next.config.js')
     if (nextConfigPath !== undefined) {
+      // We cannot load `next` at the top-level because we validate whether the
+      // site is using `next` inside `onPreBuild`.
```
lindsaylevine, line 50: 🖖

lindsaylevine (approved):

helpers/validateNextUsage.js
ehmicky: If this file exports new methods in the future, we don't need to change the places where `validateNextUsage()` was imported. 
I can change it to a default export if you prefer though. Which one do you prefer?

lindsaylevine: no preference! just used to the default non-object export consistent throughout NoN but i'm totally cool with being future-proof here, i think it makes sense if you think we might put another func/helper in here. up to you!

ehmicky: Let's go with consistency and use a default export :+1:

PR #48, opened 2020-11-17
Improve `npm install` fix

ehmicky:
Discussion at https://github.com/netlify/netlify-plugin-nextjs/issues/10

The `npm install next-on-netlify` can be simplified to `npm install next-aws-lambda` instead (which should be slightly faster).
@lindsaylevine can you please confirm this still works?


lindsaylevine (approved):
confirmed !

PR #49, opened 2020-11-18
pass constants FUNCTIONS_SRC and PUBLISH_DIR to nextOnNetlify

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/43

wasn't quite sure how to replace the tests now that the manual copying is no longer a part of the plugin. making a note to add a specific test for this in next-on-netlify.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,14 +25,14 @@
   "homepage": "https://github.com/netlify/netlify-plugin-nextjs#readme",
   "dependencies": {
     "chalk": "^4.1.0",
-    "cpx": "^1.5.0",
     "find-up": "^4.1.0",
     "make-dir": "^3.1.0",
-    "next-on-netlify": "^2.6.0",
+    "next-on-netlify": "^2.6.3",
     "semver": "^7.3.2"
   },
   "devDependencies": {
     "@netlify/eslint-config-node": "^0.3.0",
+    "cpx": "^1.5.0",
     "husky": "^4.3.0",
     "jest": "^26.6.1",
     "next": "^9.5.3",
--- a/test/index.js
+++ b/test/index.js
@@ -1,3 +1,4 @@
+const fs = require('fs')
 const path = require('path')
 const process = require('process')
 const { promisify } = require('util')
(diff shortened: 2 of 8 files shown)


lindsaylevine:
there's a "blocking" issue with this currently:
1. when you don't set a publish dir in your toml, it defaults to the root of your project, and NoN empties the directory of publishDir and functionsDir when it runs - aka it wipes the entire project and you just end up with your functions_dir lol

possible solutions:
1. force people to define a publish directory
2. don't empty the publish directory in NoN if the publishPath !== 'out_publish'
3???

package.json
```diff
@@ -28,7 +28,7 @@
     "cpx": "^1.5.0",
```
ehmicky, line 28: We can move `cpx` to `devDependencies`

test/index.js
```diff
@@ -121,51 +122,43 @@ describe('preBuild()', () => {
...
       constants: {
         PUBLISH_DIR,
+        FUNCTIONS_SRC,
       },
     })
 
-    expect(await pathExists(PUBLISH_DIR)).toBeTruthy()
+    const nextOnNetlifyOptions = nextOnNetlify.mock.calls[0][0]
+    expect(nextOnNetlifyOptions.functionsDir).toEqual(FUNCTIONS_SRC)
+    expect(nextOnNetlifyOptions.publishDir).toEqual(PUBLISH_DIR)
```
ehmicky, line 137: Sounds good :+1: 

Should we call the "real" `next-on-netlify` instead of mocking it? This would give a stronger confidence that the files are actually being copied to those directories. Generally speaking, this would be a stronger indicator that the plugin works as intended. What do you think? (Note: this would probably be better done in a separate PR)

ehmicky (approved):
Your solution `2.` sounds good :+1: (as fixed in https://github.com/netlify/next-on-netlify/pull/94)

I think this applies to the Functions directory too. If the user already has some Functions in their site, we don't want to remove them :)

test/index.js
lindsaylevine: yes, definitely. when i tried to call the "real" nextOnNetlify at first, i got errors due to the emptyDirSync stuff in the prev NoN version. now we can address this in a separate PR. here's an issue https://github.com/netlify/netlify-plugin-nextjs/issues/53 :)

ehmicky: Thanks for opening an issue for it! :heart:

PR #51, opened 2020-11-18
remove next-aws-lambda install since NoN no longer depends on it

lindsaylevine:
⚠️ TO DO BEFORE MERGING ⚠️ 
- publish NoN
- install latest NoN in here

Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/10


ehmicky (approved):
:tada:

Issue #53, opened 2020-11-18
Properly test nextOnNetlify functionality in onBuild part of plugin


lindsaylevine:
cc: @ehmicky

ehmicky:
To be explicit, this is about stopping to mock `next-on-netlify`, i.e. testing if it integrates properly with this plugin.

PR #65, opened 2020-11-25
#53 properly test next-on-netlify in onBuild

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/53


lindsaylevine:
thought: we could maybe try to get a dist from /sample

test/index.js
```diff
@@ -1,12 +1,12 @@
-const fs = require('fs')
+const fs = require('fs-extra')
```
ehmicky, line 1: `fs-extra` is missing from the `package.json`.
What about using [`cpy`](https://github.com/sindresorhus/cpy)? That library has no issues with dotfiles.

test/index.js
```diff
@@ -32,12 +32,22 @@ const changeCwd = function (cwd) {
   return process.chdir.bind(process, originalCwd)
 }
 
+// Move .next from sample project to current directory
+const moveNextDist = function () {
+  // Use copySync because cpx doesn't copy hidden files
```
ehmicky, line 37: Do you mean dotfiles by "hidden files"?
If so, I believe `cpx` does support it (see [this issue](https://github.com/mysticatea/cpx/issues/45)). The problem might come from the globbing pattern instead. Most globbing libraries will ignore dotfiles with `*` and `**` unless some `dotfile` option is passed. However, `.*` will work (see [this comment](https://github.com/mysticatea/cpx/issues/45#issuecomment-418163164)).

test/index.js
```diff
@@ -32,12 +32,22 @@ const changeCwd = function (cwd) {
...
+  fs.copySync(`${FIXTURES_DIR}/sample/.next`, `${process.cwd()}/.next`)
```
ehmicky, line 38:
```suggestion
  fs.copySync(`${FIXTURES_DIR}/sample/.next`, '.next')
```

test/index.js
```diff
@@ -32,12 +32,22 @@ const changeCwd = function (cwd) {
...
+}
+
 // Copy fixture files to the current directory
 const useFixture = async function (fixtureName) {
   const fixtureDir = `${FIXTURES_DIR}/${fixtureName}`
   await pCopy(`${fixtureDir}/**`, process.cwd())
 }
 
+beforeAll(async () => {
+  await npmRun('build', `${FIXTURES_DIR}/sample`)
```
ehmicky, line 48: Instead of running `npm run build` which runs `next build`, should we just run `next build` directly?

test/index.js
```diff
@@ -32,12 +32,22 @@ const changeCwd = function (cwd) {
...
+})
```
ehmicky, line 49: Should we merge the fixture's `package.json` with the top-level `package.json`?
However, we would need to update dependencies on several `package.json` and run `npm install` on several directories.

test/index.js
```diff
@@ -152,43 +160,36 @@ describe('preBuild()', () => {
...
-        PUBLISH_DIR,
-        FUNCTIONS_SRC,
-      },
-    })
-
-    const nextOnNetlifyOptions = nextOnNetlify.mock.calls[0][0]
-    expect(nextOnNetlifyOptions.functionsDir).toEqual(FUNCTIONS_SRC)
-    expect(nextOnNetlifyOptions.publishDir).toEqual(PUBLISH_DIR)
+  beforeEach(() => {
+    moveNextDist()
```
ehmicky, line 164: Instead of `beforeEach()`, could we repeat `moveNextDist()` in the beginning of each test function, like we do with `useFixture()`? This would make it clearer when looking at a specific file that this step is performed first.

test/index.js
lindsaylevine: oh huh i was going off [this one](https://github.com/mysticatea/cpx/issues/34) but gotchu

test/index.js
lindsaylevine: tru

test/index.js
lindsaylevine: with execa? currently this isn't even reliably working

test/index.js
lindsaylevine: next-on-netlify generally follows this model of having (full) sample projects within tests/fixtures, it's all needed to run the build from within the tests i think. i can move sample up one dir into `test/` but i think it better belongs there than top level, which is where it used to be. wherever it is, it's the convenient sample project to run the plugin against as a local plugin, so it needs its own package.json

test/index.js
ehmicky: If that does not work, [`cpy`](https://github.com/sindresorhus/cpy) could also be used.

test/index.js
ehmicky: Yes. I think removing that additional `npm` step might help make the behavior more reliable.

test/index.js
ehmicky: Having full test fixtures sound good. I am just wondering if we could keep the set of files to the minimum required for the test to work. For example, `.gitignore` might not be needed.

test/index.js
lindsaylevine: ive tried several things to copy .next with cpx and none have worked for me, feel free to pull n try n prove me wrong !!!

ehmicky: Did it work with `cpy` instead?

lindsaylevine: nope

.github/workflows/test.yml
```diff
@@ -27,4 +27,6 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm install
+      - run: cd test/fixtures/sample && npm install
+      - run: cd test/fixtures/sample && npm run build
```
lindsaylevine, line 31: even though the beforeAll in test/index.js is supposed to run the build (this works locally), this was required for tests to pass in github actions

lindsaylevine:
re: the conversation around the `sample` project/fixture - i hear what youre saying but dont see nearly a big enough issue to justify why certain files need to be removed. next-on-netlify has a ton of sample projects embedded in its cypress and test directories, with respective package.json filees and .gitignore and files. when /sample was on the root level of the project, it existed with these same files (package.json required to, well, run the project to test the plugin as a local plugin and .gitignore to avoid committing files relevant to that project). if you'd really like, we can combine the .gitignore with the top-level, but because the sample project requires different dependencies to run as a local plugin, i don't think it makes sense to combine package.json files. i hope this makes sense!

test/index.js
ehmicky: The following commit fixes it with `cpy`: b099569
I haven't pushed it so you can review it first.

ehmicky:
What you say makes sense. However, when it comes to having multiple `package.json`, we do need to consider the following drawbacks:
  - This requires adding `cd ... && npm install` in the CI, which makes CI much slower. It also makes local development setup (since multiple directories `node_modules` would now need to be installed)
  - Dependencies upgrades would now need to be done in multiple `package.json`
  - At the moment, this is a single `sample` project. But new tests might add more of them, which would increase that maintenance cost.

Separating which `dependencies` are needed in specific samples/fixtures from the main `dependencies` is nice, but it feels to me that benefit would be outweighed by those drawbacks. Just adding samples/fixtures to the top-level `package.json` `devDependencies` would just work without any of those drawbacks. What do you think?

lindsaylevine:
removing the package.json from the sample dir seems to interfere with testing the local plugin within the /sample project, which is and has been its primary purpose (see TESTING.md from original commit of the sample dir). we cant test with netlify cli without a package.json and the standard Next.js essentials. this is essential to plugin development and so unless you have a way to run `netlify build` and `netlify deploy` from sample without a package.json and the files you claim to be unnecessary, i'm going to assert the need to keep things as is.

test/index.js
lindsaylevine: i was def using cpy wrong, the word `dot` literally isnt even included in the readme

Screen Shot 2020-12-01 at 5 58 31 AM

thanks for fixing!!!!!

lindsaylevine: ahhhh i cherry-picked and your commit fails the test. im gonna rebase it out for now; could you update your branch and make sure the tests pass there and PR it into this branch?

lindsaylevine:
separate issue: the plugin won't install as a local plugin with the tests, it seems. it complains about the jest keywords being undefined. unclear at this moment if this is just local plugin specific or an issue with installing the plugin in prod as well (that wouldn't make sense, though, since we have tests in `main`/published). what's the fix here for local plugin installation? not sure if it's obvious or a plugin issue, happy to open an issue as well.

test/index.js
ehmicky: `dot` is part of `globby` which is mentioned [there](https://github.com/sindresorhus/cpy#options) :)

I fixed the bug and pushed to this PR :+1:

ehmicky:
> not sure if it's obvious or a plugin issue, happy to open an issue as well.

Yes, that might be good (including build logs and steps to reproduce), thanks!

lindsaylevine:
i think this should be good now!!!!!!!!! @ehmicky

test/sample/plugins/netlify-plugin-nextjs/index.js
```diff
@@ -0,0 +1,2 @@
+const plugin = require('../../../..')
+module.exports = plugin
```
ehmicky, line 2: This file can be removed thanks to `netlify.toml` using `../..` directly.

ehmicky (approved):
:tada: :tada: :tada: 

(Note: [one small comment](https://github.com/netlify/netlify-plugin-nextjs/pull/65/files#r534259373) about a file which can be removed)

PR #68, opened 2020-12-07
README edits to installation process

rstavchansky:
- Added information instructing users to use a Node.js package manager to install the plugin as a dependency in the site's/app's `package.json`.

- Minor copy edits


lindsaylevine (approved):
thank you!!!!

Issue #67, opened 2020-12-05
Having "next export" in npm script fails build, even if unused. 


racedale:
Error message `** Static HTML export next.js projects do not require this plugin **` was caused by `"export": "next export"` existing in my package.json file. I removed "npm run export" from my netlify.toml file when converting my site from SSG to adding some SSR pages, but had left that command in package.json (not documented in any posts that I would need to make sure to remove it). 

Was very _very_ confusing to figure out. I think the easiest fix is updating the error message to be more specific to what is detected. But maybe there is another thing to detect off of? I'm new to Next so not sure. 

[File with detection logic](https://github.com/netlify/netlify-plugin-nextjs/blob/921c95ed10fc381cb919c3e0fa0debc5ed2ab42b/helpers/isStaticExportProject.js)

Also off topic, but for a new NextJS dev the docs around next-on-netlify vs this plugin were confusing. Is next-on-netlify legacy? Is this one the only way to also have custom functions?

lindsaylevine:
hey @racedale thanks so much for bringing this up!!! really value your insight/feedback :). 

1. would love to hear what an ideal error message would've been for you to read/diagnose that build failure! if you don't have one, no worries! i can definitely make it more specific and point devs in the right direction.

2. yes, we need to clarify that :( it's been tough because internally our mission is to make next.js "just work" on netlify. that means in the not so distant future, this plugin will be _automatically installed on all new next.js sites_, without needing to add it yourself. for that reason, we've been unfortunately ambiguous about NoN vs plugin. NoN will likely always exist as its own repo, but using it as a manual install will become "legacy"/deprecated/the uncommon path.

3. yes this is the only way right now to use custom functions and publish dirs!

chriscoyier:
Noting that this just super confused me. I removed the `export` from the `.toml` file, but just that existing at all in `package.json` triggered the failed, even totally unused.

gianpaj:
Hi,

At the bottom of the announcement [blog post](https://www.netlify.com/blog/2020/12/07/announcing-one-click-install-next.js-build-plugin-on-netlify/) it says to deploy:
@cassidoo's https://github.com/cassidoo/next-netlify-starter

I get this error:
```
1:04:57 PM:   Plugin "@netlify/plugin-nextjs" failed                        
1:04:57 PM: ────────────────────────────────────────────────────────────────
1:04:57 PM: ​
1:04:57 PM:   Error message
[...]

PR #72, opened 2020-12-13
#67 next export unused script mistakenly failed builds

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/67

cc @racedale @bjonamu @olegrjumin @ahaasler @henningmu

### DIFF
--- a/helpers/isStaticExportProject.js
+++ b/helpers/isStaticExportProject.js
@@ -4,9 +4,14 @@
 
 const isStaticExportProject = ({ build, scripts }) => {
   const NEXT_EXPORT_COMMAND = 'next export'
-  const isSetInNetlifyConfig = build && build.command && build.command.includes(NEXT_EXPORT_COMMAND)
+
+  if (!build.command) return false
+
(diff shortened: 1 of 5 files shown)


helpers/isStaticExportProject.js
```diff
@@ -4,9 +4,14 @@
 
 const isStaticExportProject = ({ build, scripts }) => {
   const NEXT_EXPORT_COMMAND = 'next export'
-  const isSetInNetlifyConfig = build && build.command && build.command.includes(NEXT_EXPORT_COMMAND)
+
+  if (!build || !build.command) return
```
ehmicky, line 8: `build` is always defined, so we might be able to only check `if (!build command)`.

lindsaylevine: cool, will add. just wondering, is there any harm in checking for something even if you expect it to always exist?

ehmicky: It does not do any harm, except to code simplicity (adding some code that is unused).

Issue #25, opened 2020-11-10
Next.js version support


ehmicky:
`next-on-netlify` requires users to install `next` themselves, i.e. it is a peerDependency.

https://github.com/netlify/next-on-netlify/blob/a48735f625cc7ad1a080b322673d3e9a11f42d27/package.json#L37-L39

Furthermore, it requires them to install a specific Next.js version (`v9` even though `v10` is the latest).

In this plugin, we are using the `next` package to load the configuration file

https://github.com/netlify/netlify-plugin-nextjs/blob/2849dc5f7c57e9fd827a939e067a871e4cb487b1/index.js#L7

https://github.com/netlify/netlify-plugin-nextjs/blob/2849dc5f7c57e9fd827a939e067a871e4cb487b1/package.json#L30

We might be missing a clear strategy on Next.js versioning compatibility:
  - When a new release of Next.js is made (e.g. Next 10), users of that new release would not be able to use it on both `next-on-netlify` and this plugin, because both are pinning a specific Next.js major version
  - Users with older Next.js versions (e.g. Next 8) might not be able to use both `next-on-netlify` and this plugin

As new versions of Next.js will be released, the lack of a clear strategy might become a bigger problem.

I would recommend the following:
  - Enforcing a minimum Next.js version with this plugin
  - Using a looser `>=` range for both this plugin and `next-on-netlify` in the peerDependency version range, so that we can support multiple Next.js major releases
  - Using the user's Next.js version to load the configuration file, i.e. making `next` a `peerDependencies` instead of a `dependencies` in this plugin's `package.json`
  - Ensure Next 10 is supported
  - When time comes to make this plugin opt-out instead of opt-in for all Netlify users, builds should check the site's Next.js version to ensure it is recent enough to use this plugin

What do you think?

ehmicky:
Based on discussions with @lindsaylevine and @cassidoo on Slack, our current solution would be to only support Next.js `^9.5.3` users. 
This excludes Next.js 8 users and Next.js `<9.5.3` users. 
This also excludes Next.js 10 users for the moment, due to the current unstability of Next.js 10. However, we should do a follow-up to support that major version as soon as possible.

Enabling the plugin by default in new sites should only happen if the site is using one of our officially supported Next.js versions. We could allow Next.js 10 users to use this plugin providing they explicitly install it though.

This would result in the following actionable items:
[...]

PR #73, opened 2020-12-14
Use site's Next.js version

ehmicky:
Part of #25.

Instead of always using Next.js `9.5.3` to require `next/constants` and `next/dist/next-server/server/config`, this PR re-uses the version of Next.js used in the site.
It the site does not have Next.js installed, or if its version is `<9.5.3`, the build will fail with an error message.

This was previously reverted by #45. However, thanks to the new plugins installation flow, plugins can now require site's Node.js dependencies, which makes this PR work.


lindsaylevine (approved):
amazing!!!!!!!

PR #83, opened 2021-01-14
upgrade NoN to 2.8.2, explain CLI usage workflow in README, add jimp

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/74
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/85
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/84

next-on-netlify 2.8.1 now has file tracking per https://github.com/netlify/next-on-netlify/pull/134. this is definitely a bandaid but a decent one to get people using the CLI who want to use it. i updated the README to recommend a workflow for managing the git file tracking until we can come up with a better solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ This build plugin is a utility for supporting Next.js on Netlify. To enable serv
 ## Table of Contents
 
 - [Installation and Configuration](#installation-and-configuration)
+- [CLI Usage](#cli-usage)
 - [Custom Netlify Functions](#custom-netlify-functions)
 - [Publish Directory](#publish-directory)
 - [Custom Netlify Redirects](#custom-netlify-redirects)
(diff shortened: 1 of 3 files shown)


lindsaylevine:
cc @finnwoehm

lindsaylevine:
cc @FinnWoelm

lindsaylevine:
hm. tests failing here in next-on-netlify:

`  const cacheDir = findCacheDir({ name: "next-on-netlify", create: true });`
`const trackingFilePath = join(cacheDir, ".nonfiletracking");`

tests pass when i delete the `join` line. assuming findCacheDir `create: true` will always output the "created" cacheDir, i'm not sure why it's coming back as undefined. i suppose that assumption is wrong 🤪 .

lindsaylevine:
i'm guessing this is because of our strange node_modules/test setup in this repo? wonder if there's a workaround without having to update NoN but you probably know best @ehmicky

ehmicky:
It looks like `find-cache-dir` can return `undefined` if either:
  - No `package.json` can be found in `cwd` or any parent directory
  - There is no `node_modules` directory next to `package.json`

PR #87, opened 2021-01-21
Use new default functions directory

eduardoboucas:
With https://github.com/netlify/build/pull/2188, the default functions directory will be `netlify/functions`. This PR updates the plugin accordingly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 # Test
 test/sample/node_modules
 test/sample/out_functions
-test/sample/netlify-automatic-functions
+test/sample/netlify/functions
 test/sample/my-publish-dir
 test/sample/.next
 test/sample/.netlify
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ If you'd like to build and deploy your project using the [Netlify CLI](https://d
 2. Run any number of builds and deploys freely (i.e. `netlify build`, `netlify deploy --build`, `netlify deploy --prod`)
 3. Run `git stash --include-unstaged` to easily ignore plugin-generated files
 
-Plugin-generated files will output into either (a) the default functions and publish directories (`netlify-automatic-functions` and `.`, respectively) or (b) whichever custom functions and publish directories you configure. See below for custom directory configuration. It's important to note that, in both cases (a) and (b), the CLI may mix your project's source code and plugin-generated files; this is why we recommend committing all project source files before running CLI builds.
+Plugin-generated files will output into either (a) the default functions and publish directories (`netlify/functions` and `.`, respectively) or (b) whichever custom functions and publish directories you configure. See below for custom directory configuration. It's important to note that, in both cases (a) and (b), the CLI may mix your project's source code and plugin-generated files; this is why we recommend committing all project source files before running CLI builds.
 
 **Debugging CLI builds:**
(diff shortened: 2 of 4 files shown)


lindsaylevine:
hey @eduardoboucas ! will this be backwards compatible by chance? i.e. if build releases first and the plugin tries to run on a site before the plugin can release aka while the plugin still has the old const defined. coordinating releases might be tricky though shouldnt be too risky if we're in sync enough :)

eduardoboucas:
Hey @lindsaylevine! As far as I know, there are no plans of making this backwards-compatible (i.e. to still honour the `netlify-automatic-functions` directory if it exists), but I'm pinging @ehmicky because he has more context on this than I do.

I think we _could_ make this backwards-compatible, but it would add some complexity, so perhaps let's discuss how best to sync the release process to avoid it?

Thanks! 🙌🏻

lindsaylevine:
@eduardoboucas gotcha ok sounds good! just occurred to me as well that some people may have the plugin installed/locked at a specific version in which case their builds will still have `netlify-automatic-functions` and build will be out of sync

ehmicky:
Good point @lindsaylevine!

This PR requires `@netlify/plugin-nextjs` to be run with a version of `@netlify/build` (not released yet) using the new `netlify/functions/` name. Otherwise, Function files will be created but not bundled nor deployed.

The initial idea was to release `@netlify/plugin-nextjs` (i.e. merge PR in `netlify/plugins`) and `@netlify/build` (i.e. release the buildbot) at approximately the same time. However, there might be some shortcomings:
  - Those two events won't happen at the exact same time, which will make build fails for few minutes
  - Releasing `@netlify/plugin-nextjs` in `netlify/plugins` will not upgrade it for users who installed it in their `package.json`

It seems to me the only solution to this would be keep supporting `netlity-automatic-functions` (in addition to `netlify/functions`) in `@netlify/build`, for backward compatibility. After releasing `@netlify/build`, then we could make this change in `@netlify/plugin-nextjs`.

A separate potential problem: the new value `netlify/functions` includes a sub-directory, which might not exist yet. Would that be a problem with `next-on-netlify`?

eduardoboucas:
Good point indeed, I didn't consider that.

> It seems to me the only solution to this would be keep supporting netlity-automatic-functions (in addition to netlify/functions) in @netlify/build

I agree. I'm happy to handle that.

lindsaylevine:
>A separate potential problem: the new value netlify/functions includes a sub-directory, which might not exist yet. Would that be a problem with next-on-netlify?

jeez i hope not 😂 i _think_ our unit tests would have caught this though, right?

verythorough:
Do you think we could set an "end of support" date for the plugin versions that require `netlify-automatic-functions`? It doesn't have to be particularly soon, but it would be nice if we could set a date when we could remove that backwards-compatibility.

eduardoboucas:
> Do you think we could set an "end of support" date for the plugin versions that require netlify-automatic-functions? It doesn't have to be particularly soon, but it would be nice if we could set a date when we could remove that backwards-compatibility.

Sorry @verythorough, I missed your comment somehow. I created an issue for this at https://github.com/netlify/build/issues/2252, since removing the backwards-compatibility would happen on the Build side.

ehmicky (approved):

PR #94, opened 2021-02-04
update failure points in plugin to do nothing instead

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/93

for the test cases that previously tested the failure points, i'm thinking i could just test that it _doesnt_ copy the expected files over in onBuild()? curious if yall have a better idea for testing that the plugin does nothing lol

### DIFF
--- a/helpers/doesNotNeedPlugin.js
+++ b/helpers/doesNotNeedPlugin.js
@@ -1,11 +1,14 @@
+const findUp = require('find-up')
+
 // Checks all the cases for which the plugin should do nothing
 const isStaticExportProject = require('./isStaticExportProject')
 const doesSiteUseNextOnNetlify = require('./doesSiteUseNextOnNetlify')
 const hasCorrectNextConfig = require('./hasCorrectNextConfig')
 
-const doesNotNeedPlugin = ({ netlifyConfig, packageJson, nextConfigPath }) => {
+const doesNotNeedPlugin = async ({ netlifyConfig, packageJson }) => {
   const { build } = netlifyConfig
   const { name, scripts = {} } = packageJson
+  const nextConfigPath = await findUp('next.config.js')
 
   return (
     isStaticExportProject({ build, scripts }) ||
--- a/helpers/doesNotNeedPlugin.js
+++ b/helpers/doesNotNeedPlugin.js
@@ -1,0 +1,17 @@
+// Checks all the cases for which the plugin should do nothing
+const isStaticExportProject = require('./isStaticExportProject')
+const doesSiteUseNextOnNetlify = require('./doesSiteUseNextOnNetlify')
+const hasCorrectNextConfig = require('./hasCorrectNextConfig')
+
+const doesNotNeedPlugin = ({ netlifyConfig, packageJson, nextConfigPath }) => {
+  const { build } = netlifyConfig
+  const { name, scripts = {} } = packageJson
+
+  return (
+    isStaticExportProject({ build, scripts }) ||
+    doesSiteUseNextOnNetlify({ packageJson }) ||
+    !hasCorrectNextConfig(nextConfigPath)
+  )
+}
+
+module.exports = doesNotNeedPlugin
--- a/index.js
+++ b/index.js
@@ -1,8 +1,8 @@
 const fs = require('fs')
 const path = require('path')
 const util = require('util')
-const findUp = require('find-up')
 const makeDir = require('make-dir')
+const findUp = require('find-up')
 
 const validateNextUsage = require('./helpers/validateNextUsage')
 const doesNotNeedPlugin = require('./helpers/doesNotNeedPlugin')
@@ -16,7 +16,6 @@ const pWriteFile = util.promisify(fs.writeFile)
 module.exports = {
   async onPreBuild({ netlifyConfig, packageJson, utils }) {
     const { failBuild } = utils.build
-    const nextConfigPath = await findUp('next.config.js')
 
     validateNextUsage(failBuild)
 
@@ -25,10 +24,11 @@ module.exports = {
       return failBuild('Could not find a package.json for this project')
     }
 
-    if (doesNotNeedPlugin({ netlifyConfig, packageJson, nextConfigPath })) {
+    if (await doesNotNeedPlugin({ netlifyConfig, packageJson })) {
       return
     }
 
+    const nextConfigPath = await findUp('next.config.js')
     if (nextConfigPath === undefined) {
       // Create the next config file with target set to serverless by default
       const nextConfig = `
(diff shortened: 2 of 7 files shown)


helpers/doesNotNeedPlugin.js
```diff
@@ -0,0 +1,23 @@
...
+const doesSiteUseNextOnNetlify = require('./doesSiteUseNextOnNetlify')
+const hasCorrectNextConfig = require('./hasCorrectNextConfig')
+
+const doesNotNeedPlugin = ({ netlifyConfig, packageJson, nextConfigPath }) => {
+  const { build } = netlifyConfig
+  const { name, scripts = {}, dependencies = {} } = packageJson
+
+  const hasNoPackageJson = Object.keys(packageJson).length === 0
+  if (hasNoPackageJson) {
+    console.log('Could not find a package.json for this project')
```
erezrokah, line 12: Should we keep this as a failure?

erezrokah (changes requested):
This looks good per a small comment.

For the tests, will returning `true/false` in `onPreBuild/onBuild` if the plugin `did something/did nothing` and assert that in the tests work?

test/index.js
```diff
@@ -61,18 +61,18 @@ const DUMMY_PACKAGE_JSON = { name: 'dummy', version: '1.0.0' }
...
+  //     plugin.onPreBuild({
+  //       netlifyConfig: { build: { command: 'npm run build' } },
+  //       packageJson: { ...DUMMY_PACKAGE_JSON, scripts: { build: 'next export' } },
+  //       utils,
+  //       constants: { FUNCTIONS_SRC: 'out_functions' },
+  //     }),
+  //   ).rejects.toThrow(
+  //     `Static HTML export Next.js projects do not require this plugin. Check your project's build command for 'next export'.`,
+  //   )
+  // })
```
ehmicky, line 75: Should this be removed instead of being commented out?
We can always retrieve the old code by using `git checkout` or going back to this GitHub PR.

index.js
```diff
@@ -18,61 +16,29 @@ const pWriteFile = util.promisify(fs.writeFile)
...
             target: 'serverless'
           }
         `
       await pWriteFile('next.config.js', nextConfig)
-      console.log(`** Adding next.config.js with target set to 'serverless' **`)
     }
   },
-  async onBuild({ constants: { PUBLISH_DIR, FUNCTIONS_SRC = DEFAULT_FUNCTIONS_SRC } }) {
+  async onBuild({ netlifyConfig, packageJson, constants: { PUBLISH_DIR, FUNCTIONS_SRC = DEFAULT_FUNCTIONS_SRC } }) {
+    if (doesNotNeedPlugin({ netlifyConfig, packageJson })) {
```
ehmicky, line 38: Should this pass `nextConfigPath` too?
Otherwise, it will not check for the `target` (unlike `onPreBuild()`). So `onPreBuild()` might be skipped, but not `onBuild()`.

test/index.js
```diff
@@ -61,18 +61,18 @@ const DUMMY_PACKAGE_JSON = { name: 'dummy', version: '1.0.0' }
...
-        netlifyConfig: { build: { command: 'npm run build' } },
-        packageJson: { ...DUMMY_PACKAGE_JSON, scripts: { build: 'next export' } },
-        utils,
-        constants: { FUNCTIONS_SRC: 'out_functions' },
-      }),
-    ).rejects.toThrow(
-      `Static HTML export Next.js projects do not require this plugin. Check your project's build command for 'next export'.`,
-    )
-  })
+  // test('fail build if the app has static html export in npm script', async () => {
```
ehmicky, line 64: Instead of removing those tests, should those be updated to check that the plugin does not throw, nor call `next-on-netlify` under those scenarios?

test/index.js
lindsaylevine: i left them commented per my question/thoughts in the PR body re: these tests :)

test/index.js
lindsaylevine: see my reply above

helpers/doesNotNeedPlugin.js
lindsaylevine: that sounds like a solid idea erez thanks! will implement :) @ true/false

and yeah we can keep it is a failure!

index.js
lindsaylevine: ah yes you're right!!! thank you :) 

at first i was trying to decide if i should put the `findUp` call inside `doesNotNeedPlugin` but didnt know if i wanted to make that async so i'll just call it again in onBuild i suppose

test/index.js
```diff
@@ -61,71 +61,58 @@ const DUMMY_PACKAGE_JSON = { name: 'dummy', version: '1.0.0' }
...
-      `Static HTML export Next.js projects do not require this plugin. Check your project's build command for 'next export'.`,
-    )
+  test('do nothing if the app has static html export in npm script', async () => {
+    const output = await plugin.onPreBuild({
+      netlifyConfig: { build: { command: 'npm run build' } },
+      packageJson: { ...DUMMY_PACKAGE_JSON, scripts: { build: 'next export' } },
+      utils,
+      constants: { FUNCTIONS_SRC: 'out_functions' },
+    })
+    expect(output).toBe(undefined)
```
erezrokah, line 71: [sand] FYI, Jest has a [`toBeUndefined`](https://jestjs.io/docs/en/expect#tobeundefined) matcher.

erezrokah (approved):

index.js
```diff
@@ -82,6 +56,8 @@ module.exports = {
     // inside `onPreBuild`.
     const nextOnNetlify = require('next-on-netlify')
     nextOnNetlify({ functionsDir: FUNCTIONS_SRC, publishDir: PUBLISH_DIR })
+
+    return true
```
ehmicky, line 60: Build plugin methods should not return values. Returning values is reserved for future features.

index.js
```diff
@@ -18,61 +16,37 @@ const pWriteFile = util.promisify(fs.writeFile)
...
-    const { build } = netlifyConfig
-    const { name, scripts = {}, dependencies = {} } = packageJson
-
-    if (isStaticExportProject({ build, scripts })) {
-      return failBuild(
-        `Static HTML export Next.js projects do not require this plugin. Check your project's build command for 'next export'.`,
-      )
+    const hasNoPackageJson = Object.keys(packageJson).length === 0
+    if (hasNoPackageJson) {
+      failBuild('Could not find a package.json for this project')
```
ehmicky, line 25:
```suggestion
      return failBuild('Could not find a package.json for this project')
```

ehmicky (changes requested):
Looks good, except for one small issue with returning values in plugins.

index.js
erezrokah: This is done per my suggestion to be able to test that the method skips artifacts generation based on return value.
Another option is indeed to verify that nothing was generated in the publish/functions directory (or remove those tests completely)

ehmicky: It seems to me checking for the side effect of the function would give a stronger guarantee that it works as intended.

The problem with plugins returning values in methods is that they would prevent us from using return values in future features, which we might want to do.

ehmicky (approved):

PR #99, opened 2021-02-17
fail build if plugin cant load next.config.js

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/78
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/79

not 100% sure about trying to pass in the exact error from the catch into `failBuild`, open to thoughts!

### DIFF
--- a/helpers/doesNotNeedPlugin.js
+++ b/helpers/doesNotNeedPlugin.js
@@ -5,7 +5,7 @@ const isStaticExportProject = require('./isStaticExportProject')
 const doesSiteUseNextOnNetlify = require('./doesSiteUseNextOnNetlify')
 const hasCorrectNextConfig = require('./hasCorrectNextConfig')
 
-const doesNotNeedPlugin = async ({ netlifyConfig, packageJson }) => {
+const doesNotNeedPlugin = async ({ netlifyConfig, packageJson, utils }) => {
   const { build } = netlifyConfig
   const { name, scripts = {} } = packageJson
   const nextConfigPath = await findUp('next.config.js')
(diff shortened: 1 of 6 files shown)


helpers/hasCorrectNextConfig.js
```diff
@@ -12,7 +12,12 @@ const hasCorrectNextConfig = (nextConfigPath) => {
 
   // If the next config exists, log warning if target isnt in acceptableTargets
   const acceptableTargets = ['serverless', 'experimental-serverless-trace']
-  const nextConfig = loadConfig(PHASE_PRODUCTION_BUILD, path.resolve('.'))
+  let nextConfig
+  try {
+    nextConfig = loadConfig(PHASE_PRODUCTION_BUILD, path.resolve('.'))
+  } catch (e) {
+    return failBuild('Error loading your next.config.js.')
+  }
```
erezrokah, line 20:
```suggestion
  } catch (error) {
    return failBuild('Error loading your next.config.js.', { error })
  }
```

helpers/doesNotNeedPlugin.js
```diff
@@ -5,15 +5,15 @@ const isStaticExportProject = require('./isStaticExportProject')
 const doesSiteUseNextOnNetlify = require('./doesSiteUseNextOnNetlify')
 const hasCorrectNextConfig = require('./hasCorrectNextConfig')
 
-const doesNotNeedPlugin = async ({ netlifyConfig, packageJson }) => {
+const doesNotNeedPlugin = async ({ netlifyConfig, packageJson, utils }) => {
```
erezrokah, line 8: [dust] should we pass only `failBuild`

erezrokah (changes requested):
👍 For more error handling.

I think we should pass the original error as suggested [here](https://docs.netlify.com/configure-builds/build-plugins/create-plugins/#error-reporting) to make it easier for users to self serve

ehmicky (commented):
Thanks @lindsaylevine!

helpers/doesNotNeedPlugin.js
lindsaylevine: i was trying to keep it on the same level as netlifyConfig and packageJson, as in they're all things that get passed to the different event handlers (i.e onBuild) so i think it's ok? will keep in mind tho going forward

PR #100, opened 2021-02-17
update README with info about deploy contexts

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/97

would love your feedback as well @amuttsch!

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,14 @@ Read more about [UI-based plugin installation](https://docs.netlify.com/configur
       package = "@netlify/plugin-nextjs"
     ```
 
+ You can also add context-specific properties and environment variables to your `netlify.toml`. Read more about [deploy contexts](https://docs.netlify.com/configure-builds/file-based-configuration/#deploy-contexts) in our docs. For example:
+
+    ```toml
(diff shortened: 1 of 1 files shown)


amuttsch:
Sounds good to me :+1:

README.md
```diff
@@ -47,6 +47,14 @@ Read more about [UI-based plugin installation](https://docs.netlify.com/configur
       package = "@netlify/plugin-nextjs"
     ```
 
+ You can also add context-specific properties and environment variables to your `netlify.toml`. Read more about [deploy contexts](https://docs.netlify.com/configure-builds/file-based-configuration/#deploy-contexts) in our docs. For example:
```
erezrokah, line 50: Do you think we can fit a link to https://docs.netlify.com/cli/get-started/#run-builds-locally in here somewhere?
This is still 🚀 without it

erezrokah (approved):

README.md
lindsaylevine: we have a link to cli/get-started right below the installation steps so i think that should be fine? can always add later tho :)

PR #105, opened 2021-03-03
fix: next 10.0.8 changes loadConfig to be async

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/104

this means that sites using =>10.0.8 will be incompatible with plugin <= 2.0.1
BUT
sites using <=10.0.7 using plugin > 2.0.1 will still work at least 🙃


erezrokah (approved):

ehmicky:
I think you implied this @lindsaylevine, but just to confirm:
  - Both `next <=10.0.7` and `next >=10.0.8` should work after this PR, since awaiting a non-Promise is valid.
  - `next >=10.0.8` won't work without this PR, i.e. `@netlify/plugin-nextjs <=2.0.1` won't work for them.

src/lib/helpers/isRouteInPrerenderManifest.js
```diff
@@ -1,20 +1,21 @@
...
+    Object.entries(routes).forEach(([staticRoute, { srcRoute }]) => {
+      // This is because in i18n we set the nakedRoute to be the srcRoute in the manifest
+      if (route === srcRoute) isStaticRoute = true
+    })
+    return isStaticRoute || route in dynamicRoutes
+  }
 
-// Return true if the route is defined in the prerender manifest
-const isRouteInPrerenderManifest = (route) => {
   if (i18n.defaultLocale) return isRouteInManifestWithI18n(route)
```
lindsaylevine, line 19:
```suggestion
  if (i18n.defaultLocale) return isRouteInManifestWithI18n()
```

doesnt rly matter tho

src/lib/helpers/getPrerenderManifest.js
```diff
@@ -26,9 +26,11 @@ const transformManifestForI18n = (manifest) => {
   return { ...manifest, routes: newRoutes }
```
jlengstorf, line 26: not for now, but if we Promise.all these we'll get slightly better perf

```js
const promises = items.map(async (item) => {
  return await thing(item);
})

await Promise.all(items);

// continue
```

not a dealbreaker by any means, but something that might help us eke some extra ms out of this

jlengstorf (approved):

src/index.js
```diff
@@ -54,6 +54,7 @@ const watch = (functionsPath, publishPath) => {
     },
   )
 
+  // TO-DO: uhhhhhhhh
```
lindsaylevine, line 57:
```suggestion
```

PR #114, opened 2021-03-11
chore: separate `getNextConfig()`

ehmicky:
Part of https://github.com/netlify/netlify-plugin-nextjs/issues/79

This separates `getNextConfig()` so it can be re-used (at the moment, this code is present in two different places).

In a follow-up PR, I will remove the duplicate code at https://github.com/netlify/netlify-plugin-nextjs/blob/main/src/lib/helpers/getNextConfig.js

### DIFF
--- a/helpers/getNextConfig.js
+++ b/helpers/getNextConfig.js
@@ -1,0 +1,26 @@
+const { resolve } = require('path')
+
+let nextConfig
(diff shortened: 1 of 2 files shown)


jlengstorf (approved):
thank you!

Issue #79, opened 2021-01-08
Handle errors when loading `next.config.js`


ehmicky:
There are many reasons why loading `next.config.js` might fail such as:
  - Invalid configuration value ([example](https://app.bugsnag.com/netlify/netlify-build/errors/5fdf345b5c974300192886ec))
  - Syntax error ([example](https://app.bugsnag.com/netlify/netlify-build/errors/5ff228c670d83600174cac05))
  - General error in the JavaScript logic, throwing an exception ([example](https://app.bugsnag.com/netlify/netlify-build/errors/5fe1a057808a0e0018bce5cd))

Those errors are currently reported as plugin bugs. We should report those as user errors instead by adding a `try`/`catch` + `utils.build.failBuild()` block.

ehmicky:
Other example [here](https://app.bugsnag.com/netlify/netlify-build/errors/60073abf129fb70017badeed):

```
Specified i18n.defaultLocale should be included in i18n.locales.
See more info here: https://err.sh/next.js/invalid-i18n-config 
```

ehmicky:
This bug was fixed in #99 for: 

https://github.com/netlify/netlify-plugin-nextjs/blob/61b61f57d58ae095e3522c86013c75823911179b/helpers/hasCorrectNextConfig.js#L17

However, the configuration is loaded also here:

https://github.com/netlify/netlify-plugin-nextjs/blob/61b61f57d58ae095e3522c86013c75823911179b/src/lib/helpers/getNextConfig.js#L11

We should use a similar `try`/`catch`/`failBuild` block there as well.

[Bugsnag error](https://app.bugsnag.com/netlify/netlify-build/errors/604a23fdfea8af0007fb09f6)

ehmicky:
First part at https://github.com/netlify/netlify-plugin-nextjs/pull/114

ehmicky:
Fixed by #124.

PR #124, opened 2021-03-11
fix: fix errors while loading `next.config.js`

### LINKED ISSUE #117: Fix validation of sites without Next.js
When a site does not have Next.js, the following error message should be shown, so it is clear to users what should be fixed:

https://github.com/netlify/netlify-plugin-nextjs/blob/430569c6421f4447228ada44fe5821707f796d98/helpers/validateNextUsage.js#L9-L13

This requires importing `next` dynamically. This used to rely on `next-on-netlify` being dynamically imported. With `next-on-netlify` being merged, this stopped working and the following is shown instead (marked as a plugin bug in the build logs):

wrong_next

This should be fixed by importing `next` dynamically.

ehmicky:
Fixes #117 and #79.


jlengstorf (approved):
thank you!

PR #128, opened 2021-03-11
fix: fix memoization

ehmicky:
This fixes memoization of `getNextConfig()`.

Note: the tests seem to be failing. Work in progress.

### DIFF
--- a/helpers/getNextConfig.js
+++ b/helpers/getNextConfig.js
@@ -1,30 +1,27 @@
+'use strict'
+
 const { resolve } = require('path')
 
-let nextConfig
+const moize = require('moize')
 
 // Load next.config.js
 const getNextConfig = async function (failBuild = defaultFailBuild) {
-  // Memoizes `nextConfig`
-  if (nextConfig !== undefined) {
-    return nextConfig
-  }
-
   // We cannot load `next` at the top-level because we validate whether the
   // site is using `next` inside `onPreBuild`.
   const { PHASE_PRODUCTION_BUILD } = require('next/constants')
   const loadConfig = require('next/dist/next-server/server/config').default
 
   try {
-    nextConfig = await loadConfig(PHASE_PRODUCTION_BUILD, resolve('.'))
+    return await loadConfig(PHASE_PRODUCTION_BUILD, resolve('.'))
   } catch (error) {
     return failBuild('Error loading your next.config.js.', { error })
   }
-
-  return nextConfig
 }
+
+const moizedGetNextConfig = moize(getNextConfig, { maxSize: 1e3, isPromise: true })
 
 const defaultFailBuild = function (message, { error }) {
   throw new Error(`${message}\n${error.stack}`)
 }
 
-module.exports = getNextConfig
+module.exports = moizedGetNextConfig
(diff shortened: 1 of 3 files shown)


jlengstorf (approved):
🎉

jlengstorf:
I'm thinking this test failure is due to the addition of a `defaultFailBuild` — the test can't override `utils` now, so it's throwing an actual error

jlengstorf:
my new theory is that the memoization means the different next configs used in testing are no longer loaded due to returning the same promise

I started debugging this, but I ran into an issue where tests failed due to promises never resolving

I've run out of time to keep working on this, but for the record, here's where I ended up (tests still fail at this point)

first, I disabled memoization during tests:
```js
  // Memoizes `nextConfig` except during test runs
  if (nextConfigPromise !== undefined && process.env.NODE_ENV !== 'test') {
    return await nextConfigPromise
  }
```

the code also moved the `await` out of the try/catch, so errors weren't throwing as expected. I moved it back in:
```js
  let config
  try {
    nextConfigPromise = loadConfig(PHASE_PRODUCTION_BUILD, resolve('.'))
    config = await nextConfigPromise
  } catch (error) {
  }

  return config
```

the tests at this point are failing with:
```
  ●  Cannot log after tests are done. Did you forget to wait for something async in your test?
    Attempted to log "warn  - Detected next.config.js, no exported configuration found. https://err.sh/vercel/next.js/empty-configuration".
```

so I'm either doing something very obviously wrong (quite possible) or there's something tricky happening here that we haven't found yet

ehmicky:
Thanks for the additional debugging @jlengstorf, this really helped!
It turns out the problem was that:
  - When the promise returned by `loadConfig()` was rejected, memoized calls would not be wrapped by the `try`/`catch` block. In other words, the `try`/`catch` block should have covered the whole function, including the memoized `if` block.
  - Rejected promises should not be memoized.

I fixed those by using [a proper memoization library](https://github.com/planttheidea/moize) instead, which correctly handles async functions. Tests are now passing :tada:

helpers/getNextConfig.js
```diff
@@ -1,28 +1,29 @@
+'use strict'
```
ehmicky, line 1: Needed by `moize` (otherwise, it throws).

helpers/getNextConfig.js
```diff
@@ -1,28 +1,29 @@
...
-    nextConfig = await loadConfig(PHASE_PRODUCTION_BUILD, resolve('.'))
-  } catch (error) {
-    return failBuild('Error loading your next.config.js.', { error })
-  }
-
-  return nextConfig
+  return loadConfig(PHASE_PRODUCTION_BUILD, resolve('.'))
 }
 
+const moizedGetNextConfig = moize(getNextConfigValue, { maxSize: 1e3, isPromise: true })
```
ehmicky, line 25: `maxSize` is required by `moize@6`. `1` would work in our case since the function has no arguments, but I used `1e3` so it covers future usage where we might add more arguments to this function.

jlengstorf (approved):
nice! I'm adding this to my list of things to look at and figure out why your original solution didn't work

PR #143, opened 2021-03-16
fix: fix minimum Next.js version

### LINKED ISSUE #130: Increment `next` `MIN_VERSION`
`next` `MIN_VERSION` should be updated to `10.0.6`.

https://github.com/netlify/netlify-plugin-nextjs/blob/ed9a8c8ff42ab83f9d6bedbd6e9a6088ba1e098a/helpers/validateNextUsage.js#L31

This should only be done once https://github.com/netlify/build/issues/2402 is fixed.

ehmicky:
Fixes #130.

`next <10.0.6` is not expected to work with the current version of this plugin. Please let me know if this is not the case.

### DIFF
--- a/helpers/validateNextUsage.js
+++ b/helpers/validateNextUsage.js
@@ -28,7 +28,7 @@ const validateNextUsage = function (failBuild) {
   }
 }
 
-const MIN_VERSION = '9.5.3'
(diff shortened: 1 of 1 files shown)


lindsaylevine (approved):
tru forgot about this ty <3

PR #146, opened 2021-03-16
fix: backport `update failure points in plugin to do nothing instead` to v1

erezrokah:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/144

I created 2 branches out of the `v1.1.1` tag:
1. `release_v1.1.1` to represent the release.
2. `fix/backport_no_op` to represent the fix.

This PR cherry picks https://github.com/netlify/netlify-plugin-nextjs/commit/b2718f299e27603d4def10e0ff2f3f831d8b8f6e and wants to merge it from `fix/backport_no_op` to `release_v1.1.1` so we can cut a release from that branch.

I also kept the experimental version as `10.0.0` (see https://github.com/netlify/netlify-plugin-nextjs/commit/723be5d3cf6d1bbef3b4cedcf438afda58035d21)

I tested this locally and it works (I used https://github.com/erezrokah/next-starter-jamstack which is the starter with the update to Next.js 10 reverted)

### DIFF
--- a/helpers/doesNotNeedPlugin.js
+++ b/helpers/doesNotNeedPlugin.js
@@ -1,0 +1,20 @@
+const findUp = require('find-up')
+
+// Checks all the cases for which the plugin should do nothing
+const isStaticExportProject = require('./isStaticExportProject')
+const doesSiteUseNextOnNetlify = require('./doesSiteUseNextOnNetlify')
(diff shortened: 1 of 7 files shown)


jlengstorf (approved):
this looks correct to me. let's roll it out! thank you!

PR #163, opened 2021-03-18
chore: add pr labeler

erezrokah:
This PR adds a GitHub action to auto label PRs based on the title

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,0 +1,26 @@
+name: Label PR
+on:
+  pull_request:
+    types: [opened, edited]
+
+jobs:
+  label-pr:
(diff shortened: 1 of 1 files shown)


.github/workflows/labeler.yml
```diff
@@ -0,0 +1,26 @@
+name: Label PR
+on:
+  pull_request:
+    types: [opened, edited]
+
+jobs:
+  label-pr:
+    if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false
```
erezrokah, line 8: We can't run this action on PRs created from forks since those don't have write access to the repo.

lindsaylevine (approved):
SWeEEET thank you

PR #167, opened 2021-03-21
chore: update README with note about version compatibility

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/166

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,11 @@ Read more about [Netlify redirects](https://docs.netlify.com/routing/redirects/)
 
 ### Versions
 
-You can check our `package.json` for supported Next and Node versions. Our support of Next 10 is currently experimental.
+You can check our `package.json` for supported Node versions.
+
+For breaking Next.js releases, all Next.js versions before the breaking release will be locked to the last working plugin version:
+
+Next.js < 10.0.6 is locked to plugin version 1.1.3. To use the latest plugin version, upgrade to Next.js >= 10.0.6.
 
 ### Fallbacks for Pages with `getStaticPaths`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -134,7 +134,10 @@ Read more about [Netlify redirects](https://docs.netlify.com/routing/redirects/)
 
 ### Versions
 
-You can check our `package.json` for supported Next and Node versions. Our support of Next 10 is currently experimental.
+You can check our `package.json` for supported Next and Node versions.
```
erezrokah, line 137: Where do our `package.json` says the supported Next version? Is it here https://github.com/netlify/netlify-plugin-nextjs/blob/ff311a06991f3fc98b4f026c138137ecdfff9ad8/package.json#L68

Maybe it's better to explicitly write the the min supported version?

README.md
```diff
@@ -134,7 +134,10 @@ Read more about [Netlify redirects](https://docs.netlify.com/routing/redirects/)
...
+
+For breaking Next.js releases, all Next.js versions before the breaking release will be locked to the last working plugin version.
+You can check dependency version compatibility [here](https://github.com/netlify/plugins/blob/master/site/plugins.json).
```
erezrokah, line 140: Similar comment - can we make it easier for our users so they don't have to search for the `compatibility` field in our `plugins.json` (how would they know to search for that specific field?)

erezrokah (changes requested):
I think it might make it easier for our users to write the supported versions explicitly in the README.

Please let me know what you think.

erezrokah (approved):

README.md
```diff
@@ -134,7 +134,11 @@ Read more about [Netlify redirects](https://docs.netlify.com/routing/redirects/)
 
 ### Versions
 
-You can check our `package.json` for supported Next and Node versions. Our support of Next 10 is currently experimental.
+You can check our `package.json` for supported Node versions.
+
+For breaking Next.js releases, all Next.js versions before the breaking release will be locked to the last working plugin version:
+
+Next.js < 10.0.6 is locked to plugin version 1.1.3. To use the latest plugin version, upgrade to Next.js >= 10.0.6.
```
ehmicky, line 141: Should we replace `1.1.3` by `1.*.*` or `v1` to avoid updating the `README.md` when new backports are made to `v1`?

lindsaylevine: how often are new backports gonna be made? i feel like it should be rare enough that it's not a big deal, people should probably just be upgrading

ehmicky: :+1:

Issue #115, opened 2021-03-11
Deployment broken with Nx


ebrehault:
**Describe the bug**
A typical Next.js app built into an Nx monorepo will be located in a subfolder like `./apps/my-site`.
When the `next.config.js`file is not in the project root folder, the`onBuild` command fails with

```
Error: No static assets found in .next dist (aka no /.next/static). Please check your project configuration. Your next.config.js must be one of `serverless` or `experimental-serverless-trace`. Your build command should include `next build`.

```

Using `base` in `netlify.toml` [build] section does not fix the problem (but it is not the proper approach according the documentation as it is supposed to target a folder containing `package.json`, while an Nx project contains its `package.json` in its root, not in each app folder).

**To Reproduce**
1. Create a monorepo project with Nx
2. Add a Next.js site
3. Publish it on Netlify

**Expected behavior**
The deployment should work

**Versions**
- Next.js: ^10.0.8
- plugin (if installed at fixed version): ^3.0.0

**If you're using the CLI to build**
 - OS: [e.g. macOS, Windows]
 - netlify/cli version:

**If you're using file-based installation**
- netlify.toml:

```
[build]
  command   = "npm run build"
  publish   = "dist/apps/basic"

[[plugins]]
  package = "@netlify/plugin-nextjs"
```

jlengstorf:
thanks, @ebrehault — we're digging into this!

rayriffy:
@jlengstorf I believe that `next.config.js` should be resolve in publish directory, not directly point to the root of the repository

https://github.com/netlify/netlify-plugin-nextjs/blob/2f564b151552358689ea578563a52f2713345ccd/helpers/getNextConfig.js#L15

rayriffy:
Good news, I found a workaround. What you have to do is to build Next application to root repository instead by passing `--output-path` arguments. [More detail](https://nx.dev/latest/react/guides/nextjs#deploying-to-vercel)

```
yarn nx build basic --output-path=./
```

simonpeters:
@rayriffy your fix doesn't work for me. 

I keep getting this error.

`NX   ERROR  Running target "app:build" failed`

anyone else found a work around?

lindsaylevine:
hey all! thanks for your patience on this. we're finally getting to work on these nx/greater monorepo issues. we're consolidating everyone's issues into #433, so we'll close this in favor of that. please follow along there, but we'll try to circle back here, too, whenever we release support! 🙏

[...]

PR #178, opened 2021-03-23
refactor: locate next.config.js by root publish directory, instead of root of repository

rayriffy:
Description
---

This PR should solves #115 to handle edge cases when user build Netlify site and export `.next` and `next.config.js` to somewhere else. In case of Nx after build is succeed, `next.config` and `.next` is generated in `dists/apps/<application name>` which is nessesary to use this plugin  but current code still always looking `next.config.js` in root repository causing builds to failed due to plugin error.

https://github.com/netlify/netlify-plugin-nextjs/blob/2f564b151552358689ea578563a52f2713345ccd/helpers/getNextConfig.js#L15

### DIFF
--- a/helpers/getNextConfig.js
+++ b/helpers/getNextConfig.js
@@ -2,26 +2,23 @@
 
 const { resolve } = require('path')
 
+const { defaultFailBuild } = require('./defaultFailBuild')
 const moize = require('moize')
 
 // Load next.config.js
-const getNextConfig = async function (failBuild = defaultFailBuild) {
+const getNextConfig = async function (failBuild = defaultFailBuild, publishPath = '.') {
   // We cannot load `next` at the top-level because we validate whether the
   // site is using `next` inside `onPreBuild`.
   const { PHASE_PRODUCTION_BUILD } = require('next/constants')
   const loadConfig = require('next/dist/next-server/server/config').default
 
   try {
-    return await loadConfig(PHASE_PRODUCTION_BUILD, resolve('.'))
+    return await loadConfig(PHASE_PRODUCTION_BUILD, resolve(publishPath))
   } catch (error) {
     return failBuild('Error loading your next.config.js.', { error })
   }
 }
 
 const moizedGetNextConfig = moize(getNextConfig, { maxSize: 1e3, isPromise: true })
-
-const defaultFailBuild = function (message, { error }) {
-  throw new Error(`${message}\n${error.stack}`)
-}
 
 module.exports = moizedGetNextConfig
--- a/helpers/getNextConfig.js
+++ b/helpers/getNextConfig.js
@@ -21,6 +21,6 @@ const getNextConfig = async function (params = {}) {
   }
 }
 
-const moizedGetNextConfig = moize(getNextConfig, { maxSize: 1e3, isPromise: true })
+const moizedGetNextConfig = moize(getNextConfig, { maxSize: 1e3, isPromise: true, isDeepEqual: true })
 
 module.exports = moizedGetNextConfig
--- a/helpers/getNextConfig.js
+++ b/helpers/getNextConfig.js
@@ -6,7 +6,9 @@ const { defaultFailBuild } = require('./defaultFailBuild')
 const moize = require('moize')
 
 // Load next.config.js
-const getNextConfig = async function (failBuild = defaultFailBuild, publishPath = '.') {
+const getNextConfig = async function (params = {}) {
+  const { failBuild = defaultFailBuild, publishPath = '.' } = params
+
   // We cannot load `next` at the top-level because we validate whether the
   // site is using `next` inside `onPreBuild`.
   const { PHASE_PRODUCTION_BUILD } = require('next/constants')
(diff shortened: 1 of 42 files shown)


helpers/getNextConfig.js
```diff
@@ -2,26 +2,23 @@
 
 const { resolve } = require('path')
 
+const { defaultFailBuild } = require('./defaultFailBuild')
 const moize = require('moize')
 
 // Load next.config.js
-const getNextConfig = async function (failBuild = defaultFailBuild) {
+const getNextConfig = async function (failBuild = defaultFailBuild, publishPath = '.') {
```
ehmicky, line 9: What are your thoughts on using named parameters instead?

```js
const getNextConfig = async function ({ failBuild = defaultFailBuild, publishPath = '.' }) {
```

Like this, you would not need to specify the `failBuild` argument in `getNextDistDir.js`. 
This, in turn, might remove the need to separate `defaultFailBuild()` to its own file.

ehmicky (commented):
Thanks a lot @rayriffy!

For the Nx specific logic, I will delegate to @lindsaylevine, but the PR looks good otherwise from a code standpoint, except for one small suggestion :+1:

src/lib/helpers/getNextDistDir.js
```diff
@@ -1,9 +1,11 @@
 // Get the NextJS distDir specified in next.config.js
 const { join } = require('path')
+
 const getNextConfig = require('../../../helpers/getNextConfig')
+const { defaultFailBuild } = require('../../../helpers/defaultFailBuild')
 
-const getNextDistDir = async () => {
-  const nextConfig = await getNextConfig()
+const getNextDistDir = async (publishPath) => {
+  const nextConfig = await getNextConfig(defaultFailBuild, publishPath)
```
lindsaylevine, line 8:
```suggestion
  const nextConfig = await getNextConfig({ publishPath })
```

ya like mickael said if you're defaulting failBuild in getNextConfig as an obj param then youd only have to pass publishPath here :)

lindsaylevine (commented):
thanks so much for this!!! +1 to what mickael said about getNextConfig/getNextDistDir being called in several places :)

rayriffy:
I've make changes that being requested and my god my eyes now in tears.

Yes, I just realized that `getNextConfig()` is being called everywhere that turns out when I finished traced back everything. I ended up with up to 40 file changes (which is almost entire plugin) and 3 test suites failed regards to copying files suites.

I already reverted those changes because clearly this is way out of my control. Might try again tmr if you wanted to.

helpers/getNextConfig.js
```diff
@@ -2,26 +2,25 @@
...
 
   try {
-    return await loadConfig(PHASE_PRODUCTION_BUILD, resolve('.'))
+    return await loadConfig(PHASE_PRODUCTION_BUILD, resolve(publishPath))
   } catch (error) {
     return failBuild('Error loading your next.config.js.', { error })
   }
 }
 
 const moizedGetNextConfig = moize(getNextConfig, { maxSize: 1e3, isPromise: true })
```
ehmicky, line 24:
```suggestion
const moizedGetNextConfig = moize(getNextConfig, { maxSize: 1e3, isPromise: true, isDeepEqual: true })
```

Memoization [requires `isDeepEqual: true`](https://github.com/planttheidea/moize#isdeepequal) due to using an object as parameter.

src/lib/helpers/getNextDistDir.js
```diff
@@ -1,9 +1,11 @@
 // Get the NextJS distDir specified in next.config.js
 const { join } = require('path')
+
 const getNextConfig = require('../../../helpers/getNextConfig')
+const { defaultFailBuild } = require('../../../helpers/defaultFailBuild')
 
-const getNextDistDir = async () => {
-  const nextConfig = await getNextConfig()
+const getNextDistDir = async (publishPath) => {
+  const nextConfig = await getNextConfig({ failBuild: defaultFailBuild, publishPath })
```
ehmicky, line 8: Passing `defaultFailBuild` explicitly should not be required anymore.

lindsaylevine:
oh my goodness @rayriffy

lindsaylevine:
@rayriffy i think we might want to rethink how our config gets passed around after seeing all the work this has taken 😲 we really appreciate it though and will keep you updated

rayriffy:
I am thinking about some kind of file-based configuration generated in `onPreBuild` and have some functions (`getPluginConfig` etc.) to read configuration from file systems.

PS. Yeah, it quite hard to track down all functions that needs to be modified, I would like to suggest to enforce named exports stack and refactor code to TypeScript.

lindsaylevine:
hey @rayriffy! i'm so sorry for neglecting this PR. i think i felt bad about closing it after all of your hard work! :( here's what happened overall though:

your last comment stemmed a big conversation internally about the right way we could do this.

>I am thinking about some kind of file-based configuration generated in onPreBuild and have some functions (getPluginConfig etc.) to read configuration from file systems.

we were lightly exploring ideas of global config, file-based config like you suggested, passing down the config everywhere like you did in your work (with publishPath), etc. ultimately we knew it'd take a lot of work/thought before we could really prioritize this, and so here we are 😅 .

i'm opening a new issue shortly with our plan going forward. it'll be very close to what you've done, but passing the nextConfig down. i will make sure that, in this work, we will also check for a config in the publish dir if we cannot find one at the root.

thank you so much for all your hard work on this. i/we truly appreciate it <3. let me know if you have any other feedback. i'll keep the exports/typescript note in mind though that would be difficult to prioritize any time soon :/.

PR #182, opened 2021-03-28
feat: unstableNetlifyFunctionsSupport.includeDirs configuration

lindsaylevine:
`unstableNetlifyFunctionsSupport` i think should be fine? `functions` was v unclear and this is just gonna be taken out anyways and prob only communicated to people who find the open issue so 🤷

### DIFF
--- a/helpers/copyUnstableIncludedDirs.js
+++ b/helpers/copyUnstableIncludedDirs.js
@@ -1,0 +1,27 @@
+const path = require('path')
+const fs = require('fs')
+const AdmZip = require('adm-zip')
+const getNetlifyFunctionName = require('../src/lib/helpers/getNetlifyFunctionName')
+
+// This is UNSTABLE support for special use cases needing files served with their Next.js
+// pages that become Netlify functions. Eventually this will be supported internally.
+const copyUnstableIncludedDirs = ({ nextConfig, functionsDist }) => {
+  for (const name in nextConfig.unstableNetlifyFunctionsSupport || {}) {
+    const includeDirs = nextConfig.unstableNetlifyFunctionsSupport[name].includeDirs || []
+    console.log('Processing included dirs for ', name)
+
+    const zipName = path.join(functionsDist, getNetlifyFunctionName(name) + '.zip')
+    const zip = new AdmZip(zipName)
+    includeDirs.forEach((includes) => {
+      if (fs.lstatSync(includes).isDirectory()) {
+        // We add the files at the root of the ZIP because process.cwd()
+        // points to `/` in serverless functions
+        zip.addLocalFolder(includes, includes)
+        console.log(`Added ${includes} to ${zipName}`)
+      }
+    })
+    zip.writeZip(zipName)
+  }
+}
+
+module.exports = copyUnstableIncludedDirs
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@sls-next/lambda-at-edge": "^1.5.2",
+        "adm-zip": "^0.5.4",
         "chalk": "^4.1.0",
         "chokidar": "^3.5.1",
         "commander": "^7.1.0",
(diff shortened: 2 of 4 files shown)


lindsaylevine:
unclear about that ubuntu 10.13 failure

package-lock.json
```diff
@@ -5661,32 +5662,6 @@
...
-      "cpu": [
-        "x64"
-      ],
-      "dev": true,
-      "optional": true,
-      "os": [
-        "linux"
-      ]
-    },
-    "node_modules/@netlify/traffic-mesh-agent-win32-x64": {
```
lindsaylevine, line 5677: but why

jlengstorf (approved):

helpers/copyUnstableIncludedDirs.js
```diff
@@ -0,0 +1,26 @@
...
+    const zip = new AdmZip(zipName)
+    includeDirs.forEach((includes) => {
+      if (fs.lstatSync(includes).isDirectory()) {
+        // We add the files at the root of the ZIP because process.cwd()
+        // points to `/` in serverless functions
+        zip.addLocalFolder(includes, includes)
+        console.log(`Added ${includes} to ${zipName}`)
+      }
+    })
+    zip.writeZip(zipName)
```
ehmicky, line 22: Note: we should remove the `mtime` when creating the archive.

Otherwise, every zip archive of the same Function will have a different checksum. This has created some issues in the past which ended up increasing our AWS Lambda usage weekly rate significantly.

See https://github.com/netlify/zip-it-and-ship-it/blob/9143e4e572963dc85c96e69c216160f9c77f045d/src/archive.js#L21

lindsaylevine: hmm i’ve been reading the adm-zip docs (that are v unclear lol) and i dont quite understand how i’d do this :neutral_face: do yk exactly how this would be done with this package or should i switch to whatever zisi is doing?

ehmicky: From checking their source code, it does not look like `adm-zip` allows setting the `mtime`.

Based on this, I think copying the logic from `zip-it-and-ship-it` might be good. Eventually, I think we might want to fix this in `zip-it-and-ship-it` so `unstableNetlifyFunctionsSupport` should hopefully be removed in the future? (cc @eduardoboucas) If so, duplicating the logic might be ok IMHO.

lindsaylevine: yes this will be removed in the future! :D

helpers/copyUnstableIncludedDirs.js
```diff
@@ -0,0 +1,26 @@
...
+// pages that become Netlify functions. Eventually this will be supported internally.
+const copyUnstableIncludedDirs = ({ nextConfig, functionsDist }) => {
+  for (const name in nextConfig.unstableNetlifyFunctionsSupport || {}) {
+    const includeDirs = nextConfig.unstableNetlifyFunctionsSupport[name].includeDirs || []
+    console.log('Processing included dirs for ', name)
+
+    const zipName = path.join(functionsDist, getNetlifyFunctionName(name) + '.zip')
+    const zip = new AdmZip(zipName)
+    includeDirs.forEach((includes) => {
+      if (fs.lstatSync(includes).isDirectory()) {
```
ehmicky, line 15: `require('fs')` is missing.

Note: we might want [to add `eslint`](https://github.com/netlify/eslint-config-node) to prevent this?

helpers/copyUnstableIncludedDirs.js
cassidoo: Maybe this specific part can be a separate PR since it's a lower visibility issue?

helpers/copyUnstableIncludedDirs.js
lindsaylevine: https://github.com/netlify/netlify-plugin-nextjs/issues/210 :D

helpers/copyUnstableIncludedDirs.js
lindsaylevine: i'm gonna defer the archiver/adm swap for if issues arise - as of right now there will be v few users using this and merging this is mainly to keep the temporary logic/branch up to date with main. want to unblock the cache PR and a lot of the other work i have to do and promised this to another user 😢  @eduardoboucas would you mind opening and/or sharing the issue to track the toml-side version of this work? 🙏 🙏

PR #185, opened 2021-03-29
feat: cache .next/cache between builds

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/63

questions:
1. how will this behave with the pizzafox plugin? answer: works
2. how can this be tested within the plugin, if at all? answer: mock/spy

### DIFF
--- a/helpers/cacheBuild.js
+++ b/helpers/cacheBuild.js
@@ -1,0 +1,32 @@
+const path = require('path')
+
+const DEFAULT_DIST_DIR = '.next'
+
+// Account for possible custom distDir
+const getPath = (distDir, source) => {
+  return path.join(distDir || DEFAULT_DIST_DIR, source)
+}
+
+const restoreCache = async ({ cache, distDir }) => {
+  const cacheDir = getPath(distDir, 'cache')
+  if (await cache.restore(cacheDir)) {
+    console.log('Next.js cache restored.')
+  } else {
+    console.log('No Next.js cache to restore.')
+  }
+}
+
+const saveCache = async ({ cache, distDir }) => {
+  const cacheDir = getPath(distDir, 'cache')
+  const buildManifest = getPath(distDir, 'build-manifest.json')
+  if (await cache.save(cacheDir, { digests: [buildManifest] })) {
+    console.log('Next.js cache saved.')
+  } else {
+    console.log('No Next.js cache to save.')
+  }
+}
+
+module.exports = {
+  restoreCache,
+  saveCache,
+}
--- a/index.js
+++ b/index.js
@@ -28,7 +28,7 @@ module.exports = {
       return failBuild('Could not find a package.json for this project')
     }
 
-    const nextConfig = await getNextConfig(process.cwd(), utils.failBuild)
+    const nextConfig = await getNextConfig(utils.failBuild)
     await restoreCache({ cache: utils.cache, distDir: nextConfig.distDir })
 
     if (await doesNotNeedPlugin({ netlifyConfig, packageJson, failBuild })) {
@@ -70,7 +70,7 @@ module.exports = {
       return
     }
 
-    const nextConfig = await getNextConfig(process.cwd(), utils.failBuild)
+    const nextConfig = await getNextConfig(utils.failBuild)
     await saveCache({ cache: utils.cache, distDir: nextConfig.distDir })
     copyUnstableIncludedDirs({ nextConfig, functionsDist: FUNCTIONS_DIST })
   },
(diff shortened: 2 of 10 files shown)


helpers/cacheBuild.js
```diff
@@ -0,0 +1,33 @@
+const path = require('path')
+
+const DEFAULT_DIST_DIR = '.next'
+
+// Give advanced/monorepo users the ability to point to their custom build & dist dirs
```
ehmicky, line 5: Instead of using a `buildDir` input, couldn't users use our `base` configuration property instead?
This changes the current directory of the plugin's process, which is used by default to resolve the file paths passed to `utils.cache.save()` and `utils.cache.restore()`.

helpers/cacheBuild.js
```diff
@@ -0,0 +1,33 @@
...
+const getPath = ({ buildDir, distDir, source }) => {
+  const mainPath = path.join(distDir || DEFAULT_DIST_DIR, source)
```
ehmicky, line 7: Out of curiosity, when users Next.js directory is not `{buildDir}/.next`, is it because:
  - The directory is not named `.next`?
  - Or the directory is named `.next` but is not located in `buildDir`?

helpers/cacheBuild.js
lindsaylevine: hm i suppose so, though i know *as is* (with current monorepo support) the base configuration doesn't work for all monorepo use cases - which i guess begs the qestion, is that why pizzafox's plugin accepts buildDir as input? 🤔

helpers/cacheBuild.js
lindsaylevine: the first: directory is not named .next

https://nextjs.org/docs/api-reference/next.config.js/setting-a-custom-build-directory

thats why i wrote in my the PR desc that i technically should get distDir from nextConfig, but i just want to wait for #182 to be merged first (because these 2 PRs conflict)

helpers/cacheBuild.js
ehmicky: The `base` property is definitely more of a patch than a real solution for many monorepo setups.

In this case though, I think the `buildDir` input would use the same strategy as the `base` property. In essence, it would behave the same by changing the base directory to find `.next`.

However, please feel free to discard this comment if you think this might be a useful input. :+1: 

If we do though, I would personally suggest renaming it to `base` because `buildDir` is a somewhat internal name, and to reflect the similar configuration property with the same name.

helpers/cacheBuild.js
ehmicky: Thanks for the explanation @lindsaylevine 

> i technically should get distDir from nextConfig, but i just want to wait for #182 to be merged first (because these 2 PRs conflict)

After we do this, would there be any use cases left for the `distDir` input?

index.js
```diff
@@ -58,6 +61,9 @@ module.exports = {
 
     await nextOnNetlify({ functionsDir: FUNCTIONS_SRC, publishDir: PUBLISH_DIR })
   },
+  async onPostBuild({ inputs }) {
```
ehmicky, line 64: We should add the new `inputs` to `manifest.yml`.

lindsaylevine: RIGHT DUH

helpers/cacheBuild.js
lindsaylevine: nope, none :)

helpers/cacheBuild.js
lindsaylevine: ok no i can get rid of buildDir (or `base`) so we won't have any inputs after all! if people end up needing it, we can always add it back in :)

lindsaylevine:
@ehmicky any specific thoughts on these questions in the PR desc? 🙏 

how will this behave with the pizzafox plugin?
how can this be tested within the plugin, if at all?

ehmicky:
The cache utility deletes the destination file/directory before both saving and restoring. So I believe it should be idempotent, and running both `netlify-plugin-cache-nextjs` and the logic in this PR should work.

However, I would suggest manually testing it before merging this PR. The production behavior of the caching utility is different from local builds, so I would suggest testing both (or at least production). If it turned out to be incompatible, I would suggest:
 - Detecting whether `netlify-plugin-cache-nextjs` is used. This could be done by using `require('netlify-plugin-cache-nextjs')`.
 - If detected, the logic of this PR could be a noop.

When it comes to testing it, I would suggest using a fixture directory then:
  - Delete `.next/` if already exists (from running the test previously)
  - Create `.next/` with a `.next/dummy` in it
  - Run the plugin
  - Delete `.next/` and `.next/dummy`
  - Run the plugin again
  - Assert that `.next/dummy` exists
  - Delete `.next/` and `.next/dummy`

lindsaylevine:
i manually tested. as for automated tests, i'm not sure how we'd "run the plugin" with the real valid utils.cache hook?

ehmicky:
The above was assuming the tests would be run using `@netlify/build` programmatically. This is more robust, but also more complicated to implement.

Another option would be to spy on the mocked `utils.cache.*` methods, for example using `jest.spyOn()`. This would be simpler to implement, but would also not catch every type of bug.

ehmicky (approved):

PR #213, opened 2021-04-11
feat: add eslint

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/210

had to push --no-verify because it was hanging?? not sure why, checked/compared to `cli` to see if i was missing a dep or something 🤔

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,29 @@
+const { overrides } = require('@netlify/eslint-config-node')
+
+module.exports = {
+  extends: '@netlify/eslint-config-node',
+  rules: {
+    complexity: 0,
+    'fp/no-let': 0,
+    'fp/no-loops': 0,
+    'fp/no-mutation': 0,
+    'fp/no-mutating-methods': 0,
+    'id-length': 0,
+    'max-statements': 0,
+    'no-await-in-loop': 0,
+    'node/exports-style': 0,
+    'node/global-require': 0,
+    'node/no-unpublished-require': 0,
+    'node/prefer-global/process': 0,
+    'no-magic-numbers': 0,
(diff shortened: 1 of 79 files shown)


lindsaylevine:
Screen Shot 2021-04-11 at 4 17 26 AM

lindsaylevine:
also not sure about the ci failures with `run-s` 🤡

.github/workflows/test.yml
```diff
@@ -26,5 +26,8 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
+      - name: Linting
```
erezrokah, line 29: This should be after the `npm install` step to fix the `run-s` errors.
Also in CI actions it's better to use `npm ci` to have reproducible builds (also it's faster than `npm install`).

erezrokah (changes requested):
> had to push --no-verify because it was hanging?? not sure why, checked/compared to cli to see if i was missing a dep or something 🤔

I thought it was hanging too, but then it reported over 2000 errors, so maybe it's just slow?

erezrokah (approved):

ehmicky (changes requested):
:tada: 
Note: CI tests are failing.

.eslintrc.js
```diff
@@ -0,0 +1,36 @@
+const { overrides } = require('@netlify/eslint-config-node')
+
+const customOverrides = [
+  {
+    files: ['*.js'],
+    rules: {
```
ehmicky, line 6: Note: instead of using overrides, it it possible to put those `rules` directly under the top-level `rules` property.

PR #264, opened 2021-05-03
fix: nextConfig was memoized too early

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/254

as discussed on slack, we can't/won't write a regression test for this (because we opted to not memoize in tests). instead, we will prioritize the refactor that removes memoization altogether.


ehmicky (approved):
Nice solution in the mean time! :+1:

PR #270, opened 2021-05-04
chore: lock sls-next

lindsaylevine:


package.json
```diff
@@ -54,7 +54,7 @@
   },
   "homepage": "https://github.com/netlify/netlify-plugin-nextjs#readme",
   "dependencies": {
-    "@sls-next/lambda-at-edge": "^1.5.2",
+    "@sls-next/lambda-at-edge": "1.7.0",
```
ehmicky, line 57: Shouldn't this be `1.8.0`?

lindsaylevine: sure! i did 1.7.0 i guess to make less of a version leap but doesnt rly matter to me! force-pushed up 👍

ehmicky (approved):

PR #294, opened 2021-05-11
fix: dynamic imports with webpack 5

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/209

why, yes, this IS a tragic hack..

now a serverless function generated for a page (that requires it) will look like:
Screen Shot 2021-05-11 at 4 10 16 PM

previously it was: `nextPage/index.js`
now it is: `nextPage/nextPage/index.js` to accommodate looking for chunks two levels up in webpack 5

this is still backwards compatible with webpack 4, but unfortunately with next installed @ ^10.2.0 now in our project, we only test the webpack 5 behavior (< 10.2.0 was still default webpack 4).

i spent a good deal of time exploring the next source and webpack source to see if there was anything we could possibly do upstream to avoid this work. there's really nothing; it's just a breaking change in a major webpack update. everything runs fine on vercel, and i somehow couldn't identify any changes in the next source made to accommodate dynamic imports with webpack 5 - it appears they _just work_ with the same source for wp4 and wp5? not sure.


lindsaylevine:
possibly related but very unclear without more info from author: https://github.com/vercel/next.js/issues/24739

lukeocodes (commented):
my big concern with this one is that next makes another change upstream and we end up with even more complex logic for imports. 

is this something we can help with upstream instead?

jlengstorf (approved):
approving this to get a fix out while we explore ways to make this more robust long-term

package.json
```diff
@@ -77,7 +77,7 @@
     "husky": "^4.3.0",
     "jest": "^26.6.1",
     "netlify-cli": "^3.29.14",
-    "next": "^10.0.8",
+    "next": "^10.2.0",
```
ehmicky, line 80: This is a breaking change for users of `next` `>=10.0.8 <10.2.0`. 
Should we break builds for those users, or should we support another backport release (like `v1`)?
Note: the first option (breaking builds) might be ok if we think not many users are using those versions (do we have data on this?).

Issue #155, opened 2021-01-13
Improve next/image performance


lindsaylevine:
This issue follows up https://github.com/netlify/next-on-netlify/issues/70. Initial support for next/image was merged in https://github.com/netlify/next-on-netlify/pull/138. However, that initial support is not performant since the Netlify function will re-run on every request without any caching.

Our next step is to improve this performance with caching. This will depend on internal work at Netlify. Please follow this issue for updates on next/image performance!! :)

juliefleet:
Hey @lindsaylevine,
Any update on a release date for this feature? 🙏
Thanks!

GBrachetta:
Looking forward to the release.

Just deployed a Next.js app to Netlify with 2 very small `<Image />` components (no more than 50KB each) and they take a ridiculous amount of time to render: up to 3s for 36Kb after `<Image />` optimisation!

lindsaylevine:
@GBrachetta @juliefleet and everyone - thanks so much for your patience! last week we had two lucky folks from this thread test out our solution. on/around april 13th, we'll be able to finally offer it to everyone. 👀 hang tight!

kpfromer:
I have been watching this thread for a bit and I am super excited! Thanks for all the hard work @lindsaylevine !

GBrachetta:
Thank you @lindsaylevine, looking forward to test it and compare it with Vercel's native solution.

Will this require any action from us once it's been released (rebuild/redeploy or reinstallation of the plugin)?

lindsaylevine:
@GBrachetta

if you have your plugin installed at a fixed version (i.e. you npm installed it), then you'll need to upgrade the plugin version with the new release (when it happens).

if you just have the plugin installed via your .toml file or the UI or it's auto installed on your site, then you're good to go and no action necessary!

lindsaylevine:
hello all! for those who haven't seen: today netlify announced our first step towards solving the problems ISR/ISG originally set out to solve.

https://www.netlify.com/blog/2021/04/14/faster-builds-for-large-sites-on-netlify-with-on-demand-builders-now-in-early-access/
https://www.netlify.com/blog/2021/04/14/distributed-persistent-rendering-a-new-jamstack-approach-for-faster-builds/

what i can offer you today is a tagged release of our plugin:
`npm install --save @netlify/plugin-nextjs@experimental-odb`

this release uses on-demand builders to dramatically improve next/image performance by caching next/image requests.

[...]

PR #295, opened 2021-05-11
feat: use ODB (fallback pages + next/image)

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/155
Addresses https://github.com/netlify/netlify-plugin-nextjs/issues/151

this was previously published as a tagged release to npm but now is safe for main!

### DIFF
--- a/src/lib/steps/setupRedirects.js
+++ b/src/lib/steps/setupRedirects.js
@@ -47,10 +47,16 @@ const setupRedirects = async (publishPath) => {
   const staticRedirects = nextRedirects.filter(({ route }) => !isDynamicRoute(removeFileExtension(route)))
   const dynamicRedirects = nextRedirects.filter(({ route }) => isDynamicRoute(removeFileExtension(route)))
 
-  // Add next/image redirect to our image function
+  // Add necessary next/image redirects for our image function
   dynamicRedirects.push({
     route: '/_next/image*  url=:url w=:width q=:quality',
-    target: `/.netlify/functions/${NEXT_IMAGE_FUNCTION_NAME}?url=:url&w=:width&q=:quality`,
+    target: `/nextimg/:url/:width/:quality`,
+    statusCode: '301',
+    force: true,
+  })
+  dynamicRedirects.push({
+    route: '/nextimg/*',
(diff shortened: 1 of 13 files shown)


lindsaylevine:
need to investigate this: https://github.com/netlify/netlify-plugin-nextjs/issues/155#issuecomment-821216071

lukeocodes:
> need to investigate this: [#155 (comment)](https://github.com/netlify/netlify-plugin-nextjs/issues/155#issuecomment-821216071)

is this a blocker to this being merged?

src/lib/steps/setupRedirects.js
```diff
@@ -47,10 +47,16 @@ const setupRedirects = async (publishPath) => {
   const staticRedirects = nextRedirects.filter(({ route }) => !isDynamicRoute(removeFileExtension(route)))
   const dynamicRedirects = nextRedirects.filter(({ route }) => isDynamicRoute(removeFileExtension(route)))
 
-  // Add next/image redirect to our image function
+  // Add necessary next/image redirects for our image function
   dynamicRedirects.push({
     route: '/_next/image*  url=:url w=:width q=:quality',
```
ascorbic, line 52: Rather than using a redirect, maybe we should emulate the URL structure of one of the built-in loaders (Cloudinary and Imgix) and then set that as the loader, with our root path. https://nextjs.org/docs/basic-features/image-optimization#loader  It looks like that's what Vercel does for sites they host: they switch the loader to Cloudinary, with `assets.vercel.com` as the path.

MaximeHeckel:
I hope my comment is not a blocker @lindsaylevine, honestly if gifs can make it that's enough to me 😄

The other image might be a very specific issue on my end due to maybe the use of transparent backgrounds

ascorbic (approved):
I think this is good to go, as-is. We can refactor to use a loader and add gif support later.

lindsaylevine:
@MaximeHeckel this PR does not include gif support. i've learned it's a lot more complicated/effort than i imagined. i'll open a new issue for you on the repo. i'll also open an issue about the lambda limit as i ran into it myself! ❤️

MaximeHeckel:
That's totally fine, not a problem at all I'll handle gifs separately for now! Thank you for letting me know

PR #305, opened 2021-05-14
fix: handle dynamic import chunks in Webpack 5

ascorbic:
When using Webpack 5, Next.JS builds serverless functions with async chunks in a `chunks` directory as a sibling of the `pages` and `api` directories in `.next/serverless`. Currently we move these functions to the top level in `netlify/functions/next_xxxx`, meaning that the relative imports of the dynamic chunks in them are incorrect. e.g. with a page like `.next/serverless/pages/deeply/nested/file.js` the relative import will be something like `../../../chunks/mychunkname.js`. Currently this would be moved up to `next_deeply_nested_file/nextPage/index.js`, with chunks in `next_deeply_nested_file/chunks`

This PR preserves the directory structure: `next_deeply_nested_file/nextPage/pages/deeply/nested/file.js`, with chunks in `next_deeply_nested_file/nextPage/chunks`, so relative imports will work.

In order for our handler to import the nested file, we now also write out an entry point at `nextPage/index.js` that refers to the next page, e.g. in this example it would contain `module.exports = require("./pages/deeply/nested/file.js")`

image


src/lib/helpers/copyDynamicImportChunks.js
```diff
@@ -31,7 +31,7 @@ const copyDynamicImportChunks = async (functionPath) => {
     // WP5 files are looked for two levels up (../../chunks) in runtime
     // This is a hack to make the file one level up i.e. with
     // nextPage/nextPage/index.js, the chunk is moved to outer nextPage in a /chunks dir
```
lindsaylevine, line 33: can we update this comment to reflect your changes p pls :)

src/lib/helpers/setupNetlifyFunctionForPage.js
```diff
@@ -39,13 +39,16 @@ const setupNetlifyFunctionForPage = async ({ filePath, functionsPath, isApiPage
...
+  const nextPageCopyPath = join(functionDirectory, 'nextPage', filePath)
   const nextDistDir = await getNextDistDir()
+
   copySync(join(nextDistDir, 'serverless', filePath), nextPageCopyPath, {
     overwrite: false,
     errorOnExist: true,
   })
+
+  // Write the import entry point
+  await writeFile(join(functionDirectory, 'nextPage', 'index.js'), `module.exports = require("./${filePath}")`)
```
lindsaylevine, line 51: ah, clever! templating wouldn't work because of the require

lindsaylevine (commented):
this looks like everything is together under nextPage at the same level? 🤔 correct me if i'm wrong though but this wouldn't solve the issue then i think. if you could take a screenshot of the functions dir output like i did in my first PR that'd be dope!

lindsaylevine (approved):
lgtm!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Issue #312, opened 2021-05-16
[next_image] Fails to load most of JPG images. (Input buffer contains unsupported image format)


TheFullResolution:
**Description**

I am trying to migrate a small project to Next.js to Netlfy, I was waiting for the release of new image support, and it works great when it works. However, now, the new version can't process half of the images on the page. I can't completely determine why. I even processed all images locally to optimize the image size.

1. Go to https://graweria-new.netlify.app
3. Scroll down to the image gallery and observe how most of the images failed to load. 
4. See 502 errors in the dev tools.

Here are the logs:

`5:36:27 AM: 433f157c ERROR  Invoke Error 	{"errorType":"Error","errorMessage":"Input buffer contains unsupported image format","stack":["Error: Input buffer contains unsupported image format"]}`

![Screenshot 2021-05-16 053928](https://user-images.githubusercontent.com/11708648/118384775-2a906880-b609-11eb-83e5-0fc864783e33.png)

Here is one of the images failing:
![shop_inside (2)](https://user-images.githubusercontent.com/11708648/118384804-5a3f7080-b609-11eb-8c51-46dc9fc02cf8.jpg)

**Expected behavior**
Images are loaded

**Versions**
- Next.js: "10.2.0"
- plugin (if installed at fixed version): latest

TheFullResolution:
I realized that GitHub is processing the image example
[uploads.zip](https://github.com/netlify/netlify-plugin-nextjs/files/6488673/uploads.zip)
 so here is a zip with original files

ascorbic:
Hi @TheFullResolution 
Thanks for the report. It seems to be caused by incorrect handling of redirects with spaces in them. I've reported this internally and will also be pushing a fix in the plugin with a workaround. In the meantime, renaming the image to remove the space fixes the error. I'll try and give more helpful error messages too: the "unsupported image format" was actually an HTML 404 page!

TheFullResolution:
Nice, thank you for a quick reaction! Looking forward to your PR @ascorbic

ascorbic:
Fixed in #317

TheFullResolution:
@ascorbic, do you know when this fix will be released?

PR #317, opened 2021-05-17
fix: refactor image function

### LINKED ISSUE #307: Support for GIF and SVG in next/image
cc @maximeheckel

ascorbic:
This is a rewrite of the image function. The main fixes are improved error handling and new logic to determine output format. The logic for choosing the output format is as follows:

1. Determine the format of the input image by examining the buffer
2. If the format is SVG or GIF then we do not process it at all, so we redirect to the source image
3. If the format is unsupported (i.e. not JPEG, PNG, WebP or AVIF) then we fall-back to JPEG output
4. If the env var `FORCE_WEBP_OUTPUT` is set, then set output format to WebP. 
5. Otherwise, output format is the same as input format.
Fixes #307 and #312

### DIFF
--- a/src/lib/templates/imageFunction.js
+++ b/src/lib/templates/imageFunction.js
@@ -1,52 +1,95 @@
-const path = require('path')
 const { builder } = require('@netlify/functions')
 const sharp = require('sharp')
 const fetch = require('node-fetch')
+const imageType = require('image-type')
+const isSvg = require('is-svg')
 
-// Function used to mimic next/image and sharp
+function getImageType(buffer) {
+  const type = imageType(buffer)
+  if (type) {
+    return type
+  }
+  if (isSvg(buffer)) {
+    return { ext: 'svg', mime: 'image/svg' }
+  }
+  return null
+}
+
+const IGNORED_FORMATS = new Set(['svg', 'gif'])
+const OUTPUT_FORMATS = new Set(['png', 'jpg', 'webp', 'avif'])
+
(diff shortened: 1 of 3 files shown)


src/lib/templates/imageFunction.js
```diff
@@ -1,52 +1,95 @@
...
+
+  const type = getImageType(bufferData)
+
+  if (!type) {
+    return { statusCode: 400, body: 'Source does not appear to be an image' }
+  }
+
+  let { ext } = type
+
+  // For unsupported formats (gif, svg) we redirect to the original
```
lindsaylevine, line 61: 🔥

src/lib/templates/imageFunction.js
```diff
@@ -1,52 +1,95 @@
...
+  if (process.env.FORCE_WEBP_OUTPUT) {
+    ext = 'webp'
+  }
+
+  if (!OUTPUT_FORMATS.has(ext)) {
+    ext = 'jpg'
+  }
+
+  // The format methods are just to set options: they don't
+  // make it return that format.
```
lindsaylevine, line 80: oh huh ok!!!!!

lindsaylevine (approved):
amazing!!!!!!! i think this might also fix https://github.com/netlify/netlify-plugin-nextjs/issues/133 which i forgot about until i saw the env var?! just added the label to it - lmk what you think and/or comment on it if you see fit!!!

PR #330, opened 2021-05-20
fix: use allowlist for remote images

ascorbic:
This PR adds an allowlist for remote images, to prevent the function being used to proxy arbitrary images. The standard way to support this is by setting a value in next.config.js, however we can't do this as we don't have access to this in the function. Instead we use an env variable and warn if the user sets the value in config instead.

This also adds docs for this and for the WebP env var

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,14 @@ Similarly, you can customize your publish directory in your `netlify.toml` file:
 
 Read more about [Netlify's build settings](https://docs.netlify.com/configure-builds/get-started/#basic-build-settings) in our docs.
 
+## Image handling
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -126,6 +126,14 @@ Similarly, you can customize your publish directory in your `netlify.toml` file:
 
 Read more about [Netlify's build settings](https://docs.netlify.com/configure-builds/get-started/#basic-build-settings) in our docs.
 
+## Image handling
```
lindsaylevine, line 129: do you wanna add an anchor to this with the other section headings at the top? :)

lindsaylevine (approved):
just the one link otherwise 👍

PR #332, opened 2021-05-20
fix: add sharp optimizations

ascorbic:
Adds some sharp settings that optimize filesize. sharp recently added these options which previous required something like imageoptim/pngquant


lindsaylevine (approved):
who knew not me

ascorbic:
It's new!

PR #343, opened 2021-05-24
fix: force serverless target

ascorbic:
Uses an env var to emulate Vercel, then another to force the target to serverless if it's not set as `serverless` or `experimental-serverless-trace`


lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/204 :D

helpers/hasCorrectNextConfig.js
```diff
@@ -12,9 +11,27 @@ const hasCorrectNextConfig = async ({ nextConfigPath, failBuild }) => {
...
+
+    // 🐉 We need Next to recalculate "isZeitNow" var so we can set the target, but it's
+    // set as an import side effect so we need to clear the require cache first. 🐲
+    // https://github.com/vercel/next.js/blob/canary/packages/next/telemetry/ci-info.ts
+
+    delete require.cache[require.resolve('next/dist/telemetry/ci-info')]
+    delete require.cache[require.resolve('next/dist/next-server/server/config')]
+
+    // Clear memoized cache
+    getNextConfig.clear()
```
lindsaylevine, line 32: why do we have to clear here?

index.js
```diff
@@ -28,21 +28,8 @@ module.exports = {
...
-              target: 'serverless'
-            }
-          `
-        await pWriteFile('next.config.js', nextConfig)
-      }
-    }
+    // Populates the correct config if needed
+    await doesNotNeedPlugin({ netlifyConfig, packageJson, failBuild })
 
     // Because we memoize nextConfig, we need to do this after the write file
```
lindsaylevine, line 34: we can prob remove this comment now i think?

helpers/hasCorrectNextConfig.js
```diff
@@ -12,9 +11,27 @@ const hasCorrectNextConfig = async ({ nextConfigPath, failBuild }) => {
   const isValidTarget = acceptableTargets.includes(target)
   if (!isValidTarget) {
     console.log(
-      `Your next.config.js must set the "target" property to one of: ${acceptableTargets.join(', ')}. Update the
-      target property to allow this plugin to run.`,
+      `The "target" config property must be one of "${acceptableTargets.join('", "')}". Setting it to "serverless".`,
```
lindsaylevine, line 14: "Settingit to serverless" might imply to some that we're actually mutating/updating/changing their config file (which i don't think we're doing? 🤔 ). maybe "forcing the build to use target serverless" or something?

lindsaylevine (commented):
just need to update the tests :) plus few comments

helpers/hasCorrectNextConfig.js
ascorbic: The function is memoized, so it will return the old value if we don't clear it

helpers/hasCorrectNextConfig.js
ascorbic: Good point

ascorbic:
@ehmicky Can you think of a way for us to detect if the plugin is returning early when we're in tests? You mentioned [here](https://github.com/netlify/netlify-plugin-nextjs/pull/94#discussion_r572071819) that we shouldn't return values in plugin methods.

ascorbic:
@lindsaylevine what do you think about disabling those tests for now until we can work out a better way of checking if it's run?

lindsaylevine:
@ascorbic we can def remove the next export one. how about setting an env variable at the end of each step and checking that? i dont see a problem w that 🤔

helpers/hasCorrectNextConfig.js
ehmicky: This works. 
In the long run, we probably should consider removing memoization on `getNextConfig()` and passing its return value around function calls instead.

helpers/hasCorrectNextConfig.js
```diff
@@ -12,9 +11,29 @@ const hasCorrectNextConfig = async ({ nextConfigPath, failBuild }) => {
...
+    process.env.NOW_BUILDER = true
+    // If no valid target is set, we use an internal Next env var to force it
+    process.env.NEXT_PRIVATE_TARGET = 'serverless'
+
+    // 🐉 We need Next to recalculate "isZeitNow" var so we can set the target, but it's
+    // set as an import side effect so we need to clear the require cache first. 🐲
+    // https://github.com/vercel/next.js/blob/canary/packages/next/telemetry/ci-info.ts
+
+    delete require.cache[require.resolve('next/dist/telemetry/ci-info')]
+    delete require.cache[require.resolve('next/dist/next-server/server/config')]
```
ehmicky, line 31: Considering `next` is performing some logic at `require()`-time, this approach makes sense.

The only alternative I can think of would be to call `getNextConfig()` in a separate process. However, this would add some significant complexity. Also, this would JSON-serialize the Next.js configuration, which would not work if the config includes non-JSON values (can it?), unless we use `serialization: 'advanced'` but that option is not available in Node 10.

Two potential problems we might have in the future though are:
  - I am unsure whether this would work with ES modules since the current Node.js implementation does not allow removing cached URLs (it allows passing query parameters as cache key, but this would not work here), so this might require mixing CommonJS `require()` and ES modules `import` in order to keep using `require.cache`.
  - This relies on `next` internal file structure. The cache deletion would be a noop if they choose to rename those files. Some things we could do to fix this:
     - Run some automated tests confirming those files exist with the latest version of `next`
     - `delete` all the `next` entries in `require.cache` instead of those specific files

lindsaylevine:
yes that's definitely coming up!!! @mickael

helpers/hasCorrectNextConfig.js
```diff
@@ -12,9 +11,29 @@ const hasCorrectNextConfig = async ({ nextConfigPath, failBuild }) => {
...
+        '", "',
+      )}". Building with "serverless" target.`,
     )
+
+    /* eslint-disable fp/no-delete, node/no-unpublished-require */
+
+    // We emulate Vercel so that we can set target to serverless if needed
+    process.env.NOW_BUILDER = true
+    // If no valid target is set, we use an internal Next env var to force it
+    process.env.NEXT_PRIVATE_TARGET = 'serverless'
```
ehmicky, line 24: I believe some automated tests might be using the same process, i.e. running one test would change `process.env` for other tests. We could probably fix this by either:
  - Spawning those tests in different processes
  - Loading `next` right away so it uses those environment variables, then deleting `process.env.NOW_BUILDER` and `process.env.NEXT_PRIVATE_TARGET` (or restoring them to their original value).

ehmicky:
> @ehmicky Can you think of a way for us to detect if the plugin is returning early when we're in tests? You mentioned here that we shouldn't return values in plugin methods.

I would personally suggest either:
  - Running a unit test instead of an integration test, i.e. importing the specific function to test instead of running the whole plugin
  - Testing a side effect produced by the plugin returning early.

ascorbic:
I worked it out. I extracted the target updating code into a different funciton, so it wasn't combined with the checks which I then run before. In the tests I check whether `process.env.NEXT_PRIVATE_TARGET` has been set, which will only happen if the checks have passed.

helpers/verifyBuildTarget.js
```diff
@@ -0,0 +1,38 @@
...
+  delete require.cache[require.resolve('next/dist/telemetry/ci-info')]
+  delete require.cache[require.resolve('next/dist/next-server/server/config')]
+
+  // Clear memoized cache
+  getNextConfig.clear()
+
+  /* eslint-enable fp/no-delete, node/no-unpublished-require */
+}
+
+module.exports = verifyBuildTarget
```
lindsaylevine, line 38: love this 🙏

helpers/verifyBuildTarget.js
```diff
@@ -0,0 +1,38 @@
...
+  // If the next config exists, log warning if target isnt in acceptableTargets
+  const acceptableTargets = ['serverless', 'experimental-serverless-trace']
+  const isValidTarget = acceptableTargets.includes(target)
+  if (isValidTarget) {
+    return
+  }
+  console.log(
+    `The "target" config property must be one of "${acceptableTargets.join(
+      '", "',
+    )}". Building with "serverless" target.`,
```
lindsaylevine, line 15: NICE

index.js
lindsaylevine: double ping on this one?
```suggestion
```

test/index.js
```diff
@@ -47,6 +49,11 @@ const useFixture = async function (fixtureName) {
 // In each test, we change cwd to a temporary directory.
 // This allows us not to have to mock filesystem operations.
 beforeEach(async () => {
+  delete process.env.NEXT_PRIVATE_TARGET
+  delete require.cache[require.resolve('next/dist/telemetry/ci-info')]
+  delete require.cache[require.resolve('next/dist/next-server/server/config')]
+
+  getNextConfig.clear()
```
lindsaylevine, line 56: :O omg. we couldve done this the whole time...................

test/index.js
```diff
@@ -89,14 +85,15 @@ describe('preBuild()', () => {
   })
 
   test('run plugin if the app has next export in an unused script', async () => {
+    process.env.hi = 'ok'
```
lindsaylevine, line 88: hi

test/index.js
```diff
@@ -89,14 +85,15 @@ describe('preBuild()', () => {
...
     await plugin.onPreBuild({
       netlifyConfig,
       packageJson: { ...DUMMY_PACKAGE_JSON, scripts: { export: 'next export' } },
       utils,
       constants: {},
     })
-
-    expect(await pathExists('next.config.js')).toBeTruthy()
+    expect(process.env.NEXT_PRIVATE_TARGET).toBe('serverless')
+    // expect(await pathExists('next.config.js')).toBeTruthy()
```
lindsaylevine, line 96:
```suggestion
```

test/index.js
```diff
@@ -133,7 +130,7 @@ describe('preBuild()', () => {
       utils,
     })
 
-    expect(await pathExists('next.config.js')).toBeFalsy()
+    expect(process.env.NEXT_PRIVATE_TARGET).toBeUndefined()
```
lindsaylevine, line 133: brilliant

lindsaylevine (commented):
👏 💯

test/index.js
ascorbic: hah. oops

lindsaylevine (approved):

PR #344, opened 2021-05-24
fix: handle image rotation

ascorbic:
Rotates the image automatically, using the exif rotation data (which we then strip from the image along with all metadata)


lindsaylevine (approved):
o
 tru

👏

PR #355, opened 2021-05-26
fix(watch): add missing bin property

erezrokah:
This allows running `npx netlify-plugin-nextjs watch` to run the plugin in watch mode so it executes `next build` and `netlify-on-netlify build` on src files changes.

This used to be supported in `netlify-on-netlify`.


src/index.js
```diff
@@ -43,7 +43,7 @@ const watch = (functionsPath, publishPath) => {
   const runBuild = debounceFn(
     async () => {
       try {
-        execa.sync('next', ['build'], { stdio: 'inherit' })
+        execa.sync('next', ['build'], { stdio: 'inherit', preferLocal: true })
```
erezrokah, line 46: `preferLocal` so it uses the locally installed `next` version

lindsaylevine (approved):
the best!

PR #363, opened 2021-05-30
ci: set check_latest to true

erezrokah:
Follow up on https://github.com/netlify/netlify-plugin-nextjs/pull/328

### DIFF
--- a/src/lib/helpers/handleFileTracking.js
+++ b/src/lib/helpers/handleFileTracking.js
@@ -1,7 +1,7 @@
 const { join } = require('path')
 
 const findCacheDir = require('find-cache-dir')
-const { existsSync, readdirSync, readFileSync, rmdirSync, removeSync, writeFileSync } = require('fs-extra')
+const { existsSync, readdirSync, readFileSync, removeSync, writeFileSync } = require('fs-extra')
 
 const { NETLIFY_PUBLISH_PATH, NETLIFY_FUNCTIONS_PATH } = require('../config')
 
@@ -22,15 +22,13 @@ const handleFileTracking = ({ functionsPath, publishPath }) => {
     const trackingFile = readFileSync(trackingFilePath, 'utf8')
     const [trackedFunctions, trackedPublish] = trackingFile.split(TRACKING_FILE_SEPARATOR)
     const cleanConfiguredFiles = (trackedFiles, dirPath) => {
-      trackedFiles.forEach((file) => {
-        const filePath = join(dirPath, file.trim('\r'))
-        if (file !== '') {
-          if (process.platform === 'win32') {
-            rmdirSync(filePath, { recursive: true })
-          }
+      trackedFiles
+        .map((file) => file.trim())
+        .filter(Boolean)
+        .forEach((file) => {
+          const filePath = join(dirPath, file)
           removeSync(filePath)
-        }
-      })
+        })
     }
 
     if (isConfiguredPublishDir) {
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
-          check-latest: '*'
+          check-latest: true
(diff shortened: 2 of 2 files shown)


erezrokah:
Windows tests are failing on Node.js 16, looking into those and it suggests the plugin is not compatible with Node.js 16 + Windows

src/lib/helpers/handleFileTracking.js
```diff
@@ -22,15 +22,13 @@ const handleFileTracking = ({ functionsPath, publishPath }) => {
     const trackingFile = readFileSync(trackingFilePath, 'utf8')
     const [trackedFunctions, trackedPublish] = trackingFile.split(TRACKING_FILE_SEPARATOR)
     const cleanConfiguredFiles = (trackedFiles, dirPath) => {
-      trackedFiles.forEach((file) => {
-        const filePath = join(dirPath, file.trim('\r'))
```
erezrokah, line 26: Not sure I understand the original code. `trim` doesn't accept an argument. Also `recursive: true` is not supported for Node.js 10.
Finally I'm not sure why we need `rmdirSync` on Windows since [removeSync](https://github.com/jprichardson/node-fs-extra/blob/master/docs/remove-sync.md) does the same without failing on non existent files.

src/lib/helpers/handleFileTracking.js
```diff
@@ -22,15 +22,13 @@ const handleFileTracking = ({ functionsPath, publishPath }) => {
...
-        if (file !== '') {
```
erezrokah, line 27: I'm not sure but the intention here might be to checked the trimmed version of `file`

src/lib/helpers/handleFileTracking.js
```diff
@@ -22,15 +22,13 @@ const handleFileTracking = ({ functionsPath, publishPath }) => {
...
-          if (process.platform === 'win32') {
-            rmdirSync(filePath, { recursive: true })
-          }
+      trackedFiles
```
erezrokah, line 25: @lindsaylevine can you approve this change?

src/lib/helpers/handleFileTracking.js
lindsaylevine: Screen Shot 2021-05-31 at 2 47 40 AM

basically this work was trial and error on windows when i first did it. it was just to get a bandaid solution out so plugin users could even SORTOF use the CLI, in lieu of a CLI-based solution like having plugins build into a sep dir (instead of changing their src). if all the tests pass still with this diff then thats great!!

src/lib/helpers/handleFileTracking.js
lindsaylevine: trackingFile.split would sometimes insert '' into trackedFiles; this is just to skip that

lindsaylevine: in these cases, with file === '', it removes the current dir

src/lib/helpers/handleFileTracking.js
```diff
@@ -22,15 +22,13 @@ const handleFileTracking = ({ functionsPath, publishPath }) => {
...
+        .map((file) => file.trim())
+        .filter(Boolean)
```
lindsaylevine, line 27: is this effectively the same thing as `if (file !== '')` ?

src/lib/helpers/handleFileTracking.js
erezrokah: Sorry I should have been more clear. `trim('\r')` is the same as `trim()` as `trim` doesn't accept an argument.

src/lib/helpers/handleFileTracking.js
erezrokah: In the original code `if (file !== '')` is the original non trimmed value (`trim` doesn't mutate the string), so the check should be `if (file.trim() !== '')`.
`filter(Boolean)` operates on the trimmed value.

lindsaylevine (approved):

erezrokah:
Thanks for the review @lindsaylevine

PR #371, opened 2021-05-31
chore: move README into docs/

lindsaylevine:


docs/custom-functions.js
```diff
@@ -0,0 +1,32 @@
+## Custom Netlify Functions
```
ascorbic, line 1: Shouldn't this be a `.md` file?

docs/custom-redirects.js
```diff
@@ -0,0 +1,10 @@
+## Custom Netlify Redirects
```
ascorbic, line 1: ...and this one

lindsaylevine: lol. yes.

lindsaylevine:
fixed 🤡

ascorbic (approved):
We could probably have fewer of these pages, but I think this is good for now. I think we can find somewhere between "everything in the readme" and "a page per paragraph".

PR #374, opened 2021-06-01
feat: support for basePath

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/131

note: my first stab at this work included touching every single pageType/redirects.js file (the way we did with i18n) to handle basePath there. with that flowing into several helpers and needing to import the config everywhere again, i opted for the performance cost of needing to loop over all the redirects after theyre collected (if there's a basePath configured) and building a new, basePath-supported collection of redirects. i have my old Logic Everywhere TM commit in a separate branch if anyone's curious lol

remaining issues:
- [x] catch-all redirects are not working as is (as seen in a run of cypress). not sure the best route here
- [x] there's a lot of specific logic in here that does stuff that isnt obvious why. could document this better or possibly find a better way

### DIFF
--- a/src/tests/__snapshots__/basePath.test.js.snap
+++ b/src/tests/__snapshots__/basePath.test.js.snap
@@ -2,12 +2,6 @@
 
 exports[`Routing creates Netlify redirects 1`] = `
 "# Next-on-Netlify Redirects
-/getStaticProps/withRevalidate/2  /foo/getStaticProps/withRevalidate/2  301!
-/getStaticProps/withRevalidate/1  /foo/getStaticProps/withRevalidate/1  301!
-/getStaticProps/with-revalidate  /foo/getStaticProps/with-revalidate  301!
-/getServerSideProps/static  /foo/getServerSideProps/static  301!
-/api/static  /foo/api/static  301!
-/api/hello-background  /foo/api/hello-background  301!
 /  /foo  301!
 /_next/data/%BUILD_ID%/getServerSideProps/static.json  /.netlify/functions/next_getServerSideProps_static  200
 /_next/data/%BUILD_ID%/getStaticProps/1.json  /.netlify/functions/next_getStaticProps_id  200!  Cookie=__prerender_bypass,__next_preview_data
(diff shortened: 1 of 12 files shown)


src/tests/__snapshots__/basePath.test.js.snap
```diff
@@ -0,0 +1,74 @@
...
+/foo/getStaticProps/withRevalidate/2  /.netlify/functions/next_getStaticProps_withRevalidate_id  200
+/shows/:params/*  /foo/shows/:params  301!
+/shows/:id  /foo/shows/:id  301!
+/getStaticProps/withRevalidate/withFallback/:id  /foo/getStaticProps/withRevalidate/withFallback/:id  301!
+/getStaticProps/withFallbackBlocking/:id  /foo/getStaticProps/withFallbackBlocking/:id  301!
+/getStaticProps/withFallback/:slug/*  /foo/getStaticProps/withFallback/:slug  301!
+/getStaticProps/withFallback/:id  /foo/getStaticProps/withFallback/:id  301!
+/getServerSideProps/:id  /foo/getServerSideProps/:id  301!
+/getServerSideProps/all  /foo/getServerSideProps/all/:[slug]  301!
+/getServerSideProps/all/*  /foo/getServerSideProps/all/:[slug]  301!
```
lindsaylevine, line 50: yeah. uhhhh

lindsaylevine:
you absolutely love to see failing tests on one os/node version 💯

lindsaylevine:
@ascorbic it has to be the basePathSortFunc on that node version, that's the only thing that could affect the order like this. as for what's wrong with that sorting or includes check on that version, i have no idea 🙃

lindsaylevine:
https://github.com/nodejs/node/issues/24294 hmm

lindsaylevine:
hm. yeah. its def this sorting issue, because this second try(/commit) failed in a different order than the first. will continue to see how i can eliminate it or make it independent of arg order (like that node issue suggests)

ascorbic (approved):
🎉

PR #375, opened 2021-06-02
feat: add etag to image function

### LINKED ISSUE #372: Use Etag in image function
The image function should generate and return an Etag from the image buffer, and send a 304 response if the client has sent the same one

ascorbic:
Adds an etag header to images. We don't handle sending 304 responses because this is handled for us with an ODB
Fixes #372


lindsaylevine (approved):
rad!! only comment is to maybe rebase or squash and merge since there seem to be 2 identical feat commits

PR #385, opened 2021-06-04
fix: basePath is undefined when a site has no config file

lindsaylevine:
Fixes https://github.com/netlify/netlify-plugin-nextjs/issues/382

i had previously thought basePath defaulted to '' from loadConfig no matter what. thanks to some users who reported very quickly, that is not the case! these updated checks should fix the issue


ascorbic (approved):

antho1404 (approved):
Tested with my project where I had the issue by using this branch as described here https://github.com/netlify/netlify-plugin-nextjs/issues/382#issuecomment-854472237, and everything works as expected now.

PR #408, opened 2021-06-11
fix: force serverless target

ascorbic:
This is another attempt at #343, forcing the target to "serverless". The previous PR had to be reverted, because it didn't work when there was no next.config.js file. The target stayed as "server", meaning no serverless files were generated. This is because for some reason Next.js didn't change the target unless the config file at least exists. We work around this by creating the file if it doesn't exist, but without any config in it.


helpers/verifyBuildTarget.js
```diff
@@ -0,0 +1,52 @@
...
+
+  delete require.cache[require.resolve('next/dist/telemetry/ci-info')]
+  delete require.cache[require.resolve('next/dist/next-server/server/config')]
+
+  // Clear memoized cache
+  getNextConfig.clear()
+
+  // Creating a config file, because otherwise Next won't reload the config and pick up the new target
+
+  if (!(await findUp('next.config.js'))) {
```
lindsaylevine, line 41: fyi this is part of the monorepo problem. just an fyi we can address later

test/index.js
```diff
@@ -89,14 +85,15 @@ describe('preBuild()', () => {
   })
 
   test('run plugin if the app has next export in an unused script', async () => {
+    process.env.hi = 'ok'
```
lindsaylevine, line 88: classic

test/index.js
```diff
@@ -89,14 +85,15 @@ describe('preBuild()', () => {
...
     await plugin.onPreBuild({
       netlifyConfig,
       packageJson: { ...DUMMY_PACKAGE_JSON, scripts: { export: 'next export' } },
       utils,
       constants: {},
     })
-
-    expect(await pathExists('next.config.js')).toBeTruthy()
+    expect(process.env.NEXT_PRIVATE_TARGET).toBe('serverless')
+    // expect(await pathExists('next.config.js')).toBeTruthy()
```
lindsaylevine, line 96: 🪓

test/index.js
```diff
@@ -185,6 +182,7 @@ describe('preBuild()', () => {
 })
 
 describe('onBuild()', () => {
+  // eslint-disable-next-line max-lines
```
lindsaylevine, line 185: 🧐

helpers/verifyBuildTarget.js
```diff
@@ -0,0 +1,52 @@
...
+    await writeFile(path.resolve('next.config.js'), `module.exports = {}`)
```
lindsaylevine, line 42: at this should we maybe just add the target prop in here as a backup? not necessary ofc

lindsaylevine (approved):

test/index.js
ascorbic: oops. all your review issues came back when I cherry-picked!

helpers/verifyBuildTarget.js
ascorbic: yeah, I'm unsure what's best.

PR #417, opened 2021-06-15
chore: refactor and extend cypress tests

ascorbic:
This PR refactors the next-on-netlify Cypress tests, adding some suites and extending them to check the whole plugin.

Currently the tests run "next-on-netlify" directly. This PR changes it to use the build plugin. This takes lonegr because it needs to bundle functions, but catches errors that running NoN doesn't. The build is done using `@netlify/build` rather than via the CLI. We also add a couple of extra tests: one where there is no next config, which tests the automatic serverless target. It also adds a test using Webpack 4.


lindsaylevine (approved):

lindsaylevine:
the timeouts 💀  this is so dope tho

ascorbic:
@lindsaylevine That is basically all bundling. esbuild will fix that, when @eduardoboucas 's webpack loader stuff lands.

tiffanosaurus:
Gonna go ahead and close this

PR #420, opened 2021-06-15
chore: update to Next 11

ascorbic:
Updates test sites to Next 11. Because the defaul now is to fail on linting errors, we've had to fix a few things and add eslint configs in various places.


lindsaylevine (approved):
!

PR #438, opened 2021-06-24
docs: add links from readme to docs pages

ascorbic:
This PR adds links from the README directly to the docs pages, rather than a single link to the dirctory which is not easy to use


rstavchansky (approved):
Nice tweak!
🚢

PR #441, opened 2021-06-25
feat: don't upscale images

ascorbic:
This PR prevents us from resizing images to larger than the source. Instead we can let next/image do the resizing in the browser. Rather than returning the max size image, we instead redirect to the image that is the correct size. This is to avoid having multiple URLs that all refer to the same image, which the browser will potentially download multiple times.

[Demo page](https://deploy-preview-441--netlify-plugin-nextjs-demo.netlify.app/image). The image is displayed at twice the source size. Open the image in a new tab and you'll see it has redirected to the smaller size


lindsaylevine (approved):
whoa

lindsaylevine:
npm ERR! @netlify/routing-local-proxy-darwin-x64 not accessible from netlify-cli:@netlify/routing-local-proxy

pain

PR #457, opened 2021-06-29
fix: handle empty publish directory

ascorbic:
Fixes a bug where if the publish dir was set to an empty string, then the plugin would think that the root of the Next site was a sibling of that, i.e. it was the directory above the repo root. This PR changes a few things. First, it adds specific handling for empty `PUBLISH_DIR` and `netlifyConfig.build.publish` values when running locally, as well as handling for when the publish dir is the site root. Ideally we'd fail if somebody sets the site root as the publish dir, like we do in the Gatsby plugin, but that would be a breaking change, so instead we work around it.

This fixes the issue that broke some sites after 3.5.0 was released.

### DIFF
--- a/helpers/checkNxConfig.js
+++ b/helpers/checkNxConfig.js
@@ -2,7 +2,7 @@ const { existsSync } = require('fs')
 const { EOL } = require('os')
 const path = require('path')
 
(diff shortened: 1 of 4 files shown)


lindsaylevine (approved):
amazing!!!!

PR #471, opened 2021-07-01
fix: skip plugin if build command is empty

ascorbic:
If the build command is empty then we won't be able to generate anything, so skip running the plugin

### DIFF
--- a/helpers/doesNotNeedPlugin.js
+++ b/helpers/doesNotNeedPlugin.js
@@ -1,10 +1,19 @@
 // Checks all the cases for which the plugin should do nothing
+const { redBright } = require('chalk')
+
 const doesSiteUseNextOnNetlify = require('./doesSiteUseNextOnNetlify')
 const isStaticExportProject = require('./isStaticExportProject')
 
 const doesNotNeedPlugin = ({ netlifyConfig, packageJson }) => {
   const { build } = netlifyConfig
   const { scripts = {} } = packageJson
+
+  if (!build.command) {
+    console.log(
+      redBright`⚠️  Warning: No build command specified in the site's Netlify config, so plugin will not run. This deploy will fail unless you have already exported the site.  ⚠️`,
+    )
+    return true
+  }
 
   return isStaticExportProject({ build, scripts }) || doesSiteUseNextOnNetlify({ packageJson })
 }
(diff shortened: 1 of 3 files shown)


helpers/doesNotNeedPlugin.js
```diff
@@ -6,6 +6,11 @@ const doesNotNeedPlugin = ({ netlifyConfig, packageJson }) => {
   const { build } = netlifyConfig
   const { scripts = {} } = packageJson
 
+  if (!build.command) {
+    console.log("No build command specified in the site's Netlify config. Skipping plugin")
```
lindsaylevine, line 10: should we maybe make this more noticeable somehow? i feel like it might get lost or like it might rly be worthy of a build failure. if we were going to fail the build tho, this check would have to come after/outside of `doesNotNeedPlugin`. therefore, we'd establish, ok, yes, so this site does need the plugin. it needs a build command for the plugin to work. _then_ we could do the check, and _then_ we should probably fail the build if it doesn't exist.

ascorbic: The reason I didn't make it fail the build is because of the auto-install. It might be a site that somebody is building elsewhere, doing an export, then deploying. "`doesNotNeedPlugin`" wouldn't catch that scenario, because the command could be run elsewhere so wouldn;t show up in the config

ascorbic: Maybe just making it bright red or something

lindsaylevine: if doesNotNeedPlugin doesn't catch that scenario, wouldnt that just fail blindly anyways?

ascorbic: It would fail because of this check, so if we moved it then it wouldn't fail.

helpers/doesNotNeedPlugin.js
```diff
@@ -1,11 +1,20 @@
...
 const isStaticExportProject = require('./isStaticExportProject')
 
 const doesNotNeedPlugin = ({ netlifyConfig, packageJson }) => {
   const { build } = netlifyConfig
   const { scripts = {} } = packageJson
 
+  if (!build.command) {
+    console.log(
+      redBright`⚠️  Warning: No build command specified in the site's Netlify config, so plugin will not run. This deploy will fail unless you have already exported the site.  ⚠️`,
+    )
```
ascorbic, line 14: @lindsaylevine How about this? It also shows up in the build status now.

lindsaylevine (approved):

PR #490, opened 2021-07-06
feat!: switch function bundling to esbuild

ascorbic:
This PR switches generated functions to use esbuild. This gives significant improvements in bundling time. This requires the latest version of the CLI, so it is a breaking change and we add a check for this.

In order to force the functions to use esbuild we switch them to be in TypeScript. This is because config mutation for bundling isn't fully supported in the CLI, and TypeScript always triggers the use of esbuild.

At the same time, I've simplified the wrapper, moving the helpers into a single file, and generating the entry point dynamically so that it includes the full import to the page.


lindsaylevine (approved):
5 stars

PR #496, opened 2021-07-07
fix: handle Storybook and other sites that don't build

### LINKED ISSUE #495: Detect storybook
In auto-installed sites we may have detected that the site is using Next.js, but the site may actually be building the Storybook. This currently causes the build to fail. This is a common enough scenario to deserve special handling, where we check the build command and if it's calling `build-storybook` then skip the plugin.

ascorbic:
Currently if we detect a site as using Next.js and auto-install the plugin then builds will fail with an error message that isn't very intuitive. In a lot of cases these are actual Next.js sites, but the deployed site is building the STorybook rather than the actual site. This is a common enough pattern for it to be worth detecting specifically. This PR detects the `build-storybook` command and skips the plugin. It also adds an explicit check for the Next build directory after the build, and if it's missing it fails the build with links to docs on configuration or removing the plugin.
Fixes #495

### DIFF
--- a/helpers/doesNotNeedPlugin.js
+++ b/helpers/doesNotNeedPlugin.js
@@ -1,8 +1,9 @@
 // Checks all the cases for which the plugin should do nothing
-const { redBright } = require('chalk')
+const { redBright, yellowBright } = require('chalk')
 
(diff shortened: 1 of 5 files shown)


lindsaylevine (approved):
nnice!!!!!!!!!

PR #505, opened 2021-07-09
fix: more robust static site detection

### LINKED ISSUE #503: isStaticExport logic fails if the build command doesn't contain "next export"
This is borderlining between a bug and feature request :)

**Describe the bug**
We had the plugin start failing recently. What ended up happening is that the [isStaticExport ](https://github.com/netlify/netlify-plugin-nextjs/blob/7b5cd439f94c0e0f3c04ce405b940697249192c0/helpers/isStaticExportProject.js#L6)wasn't detected anymore.

```
const isStaticExport = usesBuildCommand({ build, scripts, command: "next export" })
```

**To Reproduce**
1. Add `"build: next build"` and `"export: next export"` to your package.json.
2. Specify this as your Netlify build command as `"netlify:build": "npm run build && npm run export"` 
3. Run build and notice that the static export wasn't detected.

This is due to "next export" not found in the build command string.

**Expected behavior**

This could be a fragile way of detecting whether the plugin should run or not. Since there doesn't seem to be a way to opt out of the plugin install, I think it's worth reviewing the approach here.

Option 1: Provide more flexibility to detect the static build
Option 2: Provide a way to opt out of having the plugin to auto-install at all (I think many will prefer having this option in an env var or something)

**Versions**
- Next.js: 11.0.1
- plugin (if installed at fixed version): 3.6.3

ascorbic:
We skip running the plugin if we detect that `next export` is being called. Previously we only checked if it was either the build command, or in the npm script that was called by the build command. This meant it missed build commands that were called indirectly, e.g. when the `build` script calls `npm run export`.

To fix this, I have used a package called `parse-npm-script`, which is designed for exactly this. However it misses a lot of edge cases, so [I have vendored it](https://github.com/netlify/parse-npm-script) and added support for parsing `npm-run-all` and `cross-env` calls. I've not done an upstream PR, because I've also removed features as I want it to be sync.

~~Finally, I've added support for an env var `NEXT_IS_STATIC_EXPORT` which can be set and will enforce this too.~~

Fixes #503


lindsaylevine (commented):
this is so awesome!! my only question is -- i was thinking the env var would be more generic. like NETLIFY_OPT_OUT_NEXT_PLUGIN or something. any reason it's so specific right now vs being an overall opt out?

ascorbic:
Good point. I've removed the env var, as I think it needs more discussing separately from this

lindsaylevine (approved):

PR #518, opened 2021-07-12
feat: add support for using image config from next.config.js

### LINKED ISSUE #516: Read image domains from next config
Our current workaround of using an env var to set image domains is not ideal, and turns out to be particularly problematic because it doesn't work for env vars set in the netlify.toml, but only for ones set in the UI.

Related: #506

ascorbic:
Currently we require users to set an env var to specify allowed domains for images. This is because the next config is not available inside the function at runtime. This is not ideal from a DX perspective, but is particularly bad because of the fact that funcitons can only use env vars that have been set in the UI, not in the toml.

This PR instead writes a snapshot of the image config as a JSON file in the image function dir, while can then `require()` it. It still supports the env var for compatibility reasons, by merging any domains specified there. However it also emits a warning telling the user to switch to using the `next.config.js` instead.

Fixes #516


lindsaylevine (approved):
!

PR #520, opened 2021-07-12
fix: use POSIX paths for Windows require()s

ascorbic:
Currently we are using paths generated by path.join to generate paths for require(), which causes issues on Windows. This PR wraps the calls in slash(), which switches everything to use / posix-style paths, which is what require() expects.


lindsaylevine (approved):
:O

ascorbic:
Checked with user and confirmed fix.

PR #530, opened 2021-07-15
docs: clarify env var usage

### LINKED ISSUE #515: Document env var limitations
Limitations in the runtime mean that only env vars defined in the Netlify UI are accessible inside functions. We need to document this, particularly because of needing it for image domains. 

Related: #506

ascorbic:
Mention in the docs that env vars can't be set in toml. Fixes #515


docs/image-handling.md
```diff
@@ -2,7 +2,7 @@
...
+By default, images are returned in the same format as the source image if they are in JPEG, PNG, WebP or AVIF format. If you are only targeting modern browsers and want to live life on the edge, you can [set the environment variable](https://docs.netlify.com/configure-builds/environment-variables/) `FORCE_WEBP_OUTPUT` to `"true"`, and it will return all images in WebP format. This must be set in the UI or via the `netlify env:set` command: env vars set in `netlify.toml` are not available to functions. Setting this will often lead to significant improvements in file size. However you should not use this if you need to support older browsers, as `next/image` does not support picture tag source fallback and images will appear broken. Check [browser support](https://caniuse.com/webp) to see if you are happy to do this.
```
lindsaylevine, line 5: >This must be set in the UI or via the `netlify env:set` command: env vars set in `netlify.toml` are not available to functions.

this seems to be a typo, i think? 🤔  should there be a period after command? like:

>`netlify env:set` command. Env vars set in netlify.toml are not available to functions.

?

ascorbic: No, it was just a colon to separate clauses. I can change to a period if that's clearer.

lindsaylevine: a period would be great!

ascorbic:
```suggestion
By default, images are returned in the same format as the source image if they are in JPEG, PNG, WebP or AVIF format. If you are only targeting modern browsers and want to live life on the edge, you can [set the environment variable](https://docs.netlify.com/configure-builds/environment-variables/) `FORCE_WEBP_OUTPUT` to `"true"`, and it will return all images in WebP format. This must be set in the UI or via the `netlify env:set` command. Any env vars set in `netlify.toml` are not available to functions. Setting this will often lead to significant improvements in file size. However you should not use this if you need to support older browsers, as `next/image` does not support picture tag source fallback and images will appear broken. Check [browser support](https://caniuse.com/webp) to see if you are happy to do this.
```

lindsaylevine (approved):

PR #547, opened 2021-07-21
feat: add NEXT_PLUGIN_FORCE_RUN to skip auto-detection

### LINKED ISSUE #546: Allow static site export to be set via env var
We currently use several heuristics to detect if a site is actually a static export, so we can disable the plugin. However it's not 100% effective, and we have some false positives and negatives. We should add an env var such as `NEXT_STATIC_EXPORT` which overrides the detection and sets it to either true or false.

ascorbic:
This PR adds support for an env var `NEXT_PLUGIN_FORCE_RUN`, which skips autodetection and forces the plugin to either always run (if set to `"true"` or `"1"`) or never run (if set to `"false"` or `"0"`)

Fixes #546

### DIFF
--- a/helpers/usesBuildCommand.js
+++ b/helpers/usesBuildCommand.js
@@ -29,7 +29,7 @@ The build command is "${build.command}" and the available npm scripts are: ${Obj
         .map((script) => `"${script}"`)
         .join(', ')}
 
-If the site does use static export then you can set the env var NEXT_PLUGIN_FORCE_RUN to "false"`,
+If the site does use static export then you can set the env var NEXT_PLUGIN_FORCE_RUN to "false" or uninstall the plugin. See https://ntl.fyi/remove-plugin for instructions.`,
     )
   }
 }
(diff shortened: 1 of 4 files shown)


helpers/usesBuildCommand.js
```diff
@@ -24,11 +24,12 @@ const usesBuildCommand = ({ build, scripts, command }) => {
...
-      'There was an error parsing your build command:',
-      error.message,
-      `\n\nThe build command is "${build.command}" and the available npm scripts are: ${Object.keys(scripts)
+      `Static export detection disabled because we could not parse your build command: ${error.message}
+The build command is "${build.command}" and the available npm scripts are: ${Object.keys(scripts)
         .map((script) => `"${script}"`)
-        .join(', ')}`,
+        .join(', ')}
+
+If the site does use static export then you can set the env var NEXT_PLUGIN_FORCE_RUN to "false"`,
```
lindsaylevine, line 32: should we also point to docs to disable/remove/uninstall the plugin as well

lindsaylevine (approved):
rad !

helpers/usesBuildCommand.js
ascorbic: Done

PR #583, opened 2021-08-03
feat: switch to ipx for images

ascorbic:
This PR switches our custom sharp-based image handler function to a simpler wrapper around [ipx](https://github.com/unjs/ipx/). ipx is the image server used by nuxt/image, and is also sharp-based. Rather than maintaining our own image server code, we can delegate that to an existing, similar project.


lindsaylevine (approved):
woooooo

PR #606, opened 2021-08-17
feat: try to generate stable build ids

### LINKED ISSUE #604: Generate stable build IDs
The default build id is a random UUID, which means that builds are not reproducible. This means that functions need to be re-uploaded every time, meaning deploys take longer than needed. We should by default use a deterministic ID. We can do this in the generated Next config, and could probably detect and log a warning too.

ascorbic:
By default Next.js generates random build IDs. This means that every build will generate changed files, which means they need to be deployed every time. This PR changes the generated `next.config.js` to use a static build ID, and adds an `onPostBuild` check that logs a warning if it's using random IDs.

Fixes  #604

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -149,6 +149,9 @@ Temporary Items
 .apdisk
 
 # End of https://www.toptal.com/developers/gitignore/api/osx,node
+
 demo/package-lock.json
+# We generate this on build in the demo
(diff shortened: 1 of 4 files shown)


ingride (approved):
looks good to me 

🚉

PR #618, opened 2021-08-25
chore: update node version in tests

ascorbic:
Because Next.js has dropped support for Node 10, we can drop it from tests too. Instead, I've added tests on Node 12, which is the current minimum version


lindsaylevine (approved):
re-run for mac?

PR #681, opened 2021-10-08
fix: remove swc binaries

ascorbic:
Next 11.1.0 added a dependency on swc that meant a >30MB binary was being bundled inside every function. The latest version of zip-it-and-ship-it allows you to exclude node modules from the bundle. This fix approximately halves the size of the bundle on the demo site.


ricbartm:
folks, can I get notified when this is live and users get it automatically in their builds? I'd like to keep an extra eye as soon as we can see its effects

PR #761, opened 2021-11-02
fix: don't force rewrite in preview mode

### LINKED ISSUE #753: [Bug]: Images in public files don't load while in preview mode
### Steps to reproduce

I've created a project with create-next-app that simply includes an image on the page as well as the /api/ preview files taken from the demo project of this repo.

To see what's happening:

1. https://naughty-borg-b1bb33.netlify.app to see the index page which includes an img tag that loads an image from the public directory
2. https://naughty-borg-b1bb33.netlify.app/api/enterPreview to enter preview mode 
3. Go back here https://naughty-borg-b1bb33.netlify.app to see that the image no longer loads

I'm not quite sure why this is happening yet, but maybe the route handling for preview mode isn't handling images in the public folder? 

This only seems to affect preview mode as far as I can tell. I've tested in another repository and it seems to affect any file located in the public directory.

https://github.com/Christopotamus/next-netlify-preview-images

- [ ] I am building using file-based configuration (netlify.toml)

### What OS are you using?

_No response_

### Your `netlify.toml` file

_No response_

### Relevant log output (or link to your logs)

_No response_

ascorbic:
### Summary

Currently we force a root level rewrite when in preview mode, so that we don't use ODBs or static HTML. This causes a bug where other static files are not found (e.g. images, css etc). This PR removes the `force` attribute. Because this would mean that statically-generated page files would then be served if using the ﻿`EXPERIMENTAL_MOVE_STATIC_PAGES` option, this also adds forced redirects for these.

### Relevant links (GitHub issues, Notion docs, etc.) or a picture of cute animal (small picture)
image

Fixes #753


tiffanosaurus (approved):
[sand] Another test to check that force is true would be good
[sand] Most likely this needs docs

Other than ^ LGTM! Could you include test steps? I've been publishing alpha versions and using them, but I'm guessing there's a better way.

ascorbic:
@tiffanosaurus Normally we can test most PRs using the plugin's demo site. Otherwise, pointing another site at the repo using `package = "../path/to/netlify-plugin-nextjs"` should do the trick

ascorbic:
Re the docs, I'll try and cover this all in the rewrites/redirects doc that I'm trying to get done

tiffanosaurus (approved):

PR #793, opened 2021-11-12
fix: use forced catchall for preview mode

### LINKED ISSUE #791: Replace individual rewrites for preview mode with a catch-all
With moving static files enabled, we currently generate a forced redirect for every file in preview mode, so that the static file isn;t served. For large pre-rendered sites this generates a massive number of rewrites that causes builds to time out. We should be able to replace these with a catch-all, but would need to do some careful handling of other rewrites so that they don;t get override (particularly the `/_next/static` one).

ascorbic:
### Summary
For preview mode, instead of generating a forced redirect for every pre-rendered page, instead make the catch-all a forced redirect. This prevents the need to generate very large numbers of redirects.

### Test plan

1. In the demo site, load `/previewTest`. Check the message to see preview mode is disabled
2. Go to `/api/enterPreview`
3. Go back to `/previewTest`, and check that preview mode is enabled.
4. Go to `/api/exitPreview`

Fixes #791


tiffanosaurus (approved):
LGTM

PR #796, opened 2021-11-14
chore: update issue and pr templates

ascorbic:
Updates the PR and issue templates

image

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,6 +1,3 @@
-# Markdown
-*.md
-
 # Logs
 logs
 *.log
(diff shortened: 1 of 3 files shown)


.prettierignore
```diff
@@ -19,6 +16,5 @@ node_modules
 .parcel-cache
 
 # Test
-demo
```
tiffanosaurus, line 22: [dust] I don't really see a reason why we should ignore demo since it should automatically lint anyway 🤷‍♀️

tiffanosaurus (approved):

PR #804, opened 2021-11-16
chore: convert entrypoint to TypeScript

ascorbic:
### Summary

Updates the entrypoint to use TypeScript, allowing the demo to run without intermediate compile step. Converts one of the helper files too: we can gradually port the rest.


tiffanosaurus:
I'm guessing this is better to restart fresh from the last couple ts changes and close this

PR #812, opened 2021-11-18
feat: don't move files to CDN if they match middleware

### LINKED ISSUE #802: Disable static page serving for routes covered by middleware
When in static page serving mode (i.e. if `EXPERIMENTAL_MOVE_STATIC_PAGES` is enabled), any middleware is not run for pages that match it, because it's served from the CDN before reaching the origin. Because of this, we should detect if pages are covered by middleware, and if so do not move them, and instead log a warning saying that they won't be served from the CDN.

ascorbic:
### Summary
When static file moving is enabled, pre-rendered pages are moved to the CDN rather than being served from the origin with next/server. This causes a problem if the site uses middleware, as it does not run on pages served from the CDN. This PR checks any moved files to see if they match any middleware. If they do, it does not move them to the CDN and instead serves them from the origin. It also logs a warning about this.

### Test plan

- Deploy the site with `EXPERIMENTAL_MOVE_STATIC_PAGES`.
- Load `/middle`
- Confirm that the message says middleware is enabled

Fixes #802

### Standard checks:

---


src/helpers/files.js
```diff
@@ -11,12 +13,41 @@ const TEST_ROUTE = /(|\/)\[[^/]+?](\/|\.html|$)/
...
+const stripLocale = (rawPath, locales = []) => {
+  const [locale, ...segments] = rawPath.split('/')
+  if (locales.includes(locale)) {
+    return segments.join('/')
+  }
+  return rawPath
+}
+
+const matchMiddleware = (middleware, filePath) =>
+  middleware.includes('') ||
```
eduardoboucas, line 25: I don't have enough context to understand this check. Can you share an example of the shape `middleware` can take?

src/helpers/files.js
```diff
@@ -29,18 +60,61 @@ exports.moveStaticPages = async ({ netlifyConfig, target, i18n, failBuild }) =>
...
     return limit(moveFile, filePath)
   })
   await Promise.all(promises)
   console.log(`Moved ${files.length} files`)
 
+  if (matchedPages.size !== 0) {
+    console.log(
+      yellowBright(outdent`Skipped moving ${matchedPages.size} ${
+        matchedPages.size === 1 ? 'file because it matches' : 'files because they match'
+      } middleware, so cannot be deployed to the CDN and will be served from the origin instead. This is fine, but we're letting you know because it may not be what you expect.
```
eduardoboucas, line 91: 👌🏻

src/helpers/files.js
```diff
@@ -29,18 +60,61 @@ exports.moveStaticPages = async ({ netlifyConfig, target, i18n, failBuild }) =>
...
+    console.log(
+      outdent`
+      
+    The following middleware matched statically-rendered pages:
+
+    ${yellowBright([...matchingMiddleware].map((mid) => `- /${mid}/_middleware`).join('\n'))}
+    
+    `,
+    )
+    // There could potentially be thousands of matching pages, so we don't want to spam the console with this
```
eduardoboucas, line 104: Is it fine not to show any pages at all when this happens? Is it worth showing something like "The following files and X others matched middlware (...)", or is that overkill?

src/helpers/files.js
eduardoboucas: Ah, never mind. I think I figured it out by looking at the tests.

eduardoboucas (approved):
LGTM! Left minor comments, none blocking.

src/helpers/files.js
ascorbic: It's an array of paths, relative to the site root, that indicate directories which have middleware. Next.js applies middleware to any path below that. The leading slash has been stripped. There are some examples in the unit tests. An empty string indicates root-level middleware thata applies to everything

src/helpers/files.js
ascorbic: Hmm, possibly, though I'm not sure if that's useful information.

eduardoboucas (approved):

PR #832, opened 2021-11-23
fix: don't move files to the CDN if they match redirect/rewrite rules

### LINKED ISSUE #819: Avoid moving static pages that are the source of redirects
If a static page has been deployed to the CDN then it is returned before any Next.js rewrites or redirects can take effect. For this reason we should serve these from the origin instead.

ascorbic:
### Summary

Ensures that if a file would match a redirect or `beforeFiles` rewrite, it isn;t served from the CDN as that would prevent the rule from being applied. We could potentially be smarter and check if we can represent the rule using Netlify redirects, but that would be quite complicated to achieve and wouldn't work for lots of scenarios

### Test plan

1. Visit these pages and ensure they redirect to the locale root: [en](https://deploy-preview-832--netlify-plugin-nextjs-demo.netlify.app/redirectme) [fr](https://deploy-preview-832--netlify-plugin-nextjs-demo.netlify.app/fr/redirectme)
2. Visit these pages and ensure that they show the image page, rather than an error message: [en](https://deploy-preview-832--netlify-plugin-nextjs-demo.netlify.app/old/image/) [fr](https://deploy-preview-832--netlify-plugin-nextjs-demo.netlify.app/fr/old/image/)

Fixes #819

image

### DIFF
--- a/src/helpers/files.js
+++ b/src/helpers/files.js
@@ -42,6 +42,9 @@ const matchesRedirect = (file, redirects) => {
 }
 
 const matchesRewrite = (file, rewrites) => {
+  if (Array.isArray(rewrites)) {
+    return matchesRedirect(file, rewrites)
+  }
   if (!Array.isArray(rewrites?.beforeFiles)) {
     return false
   }
--- a/src/helpers/files.js
+++ b/src/helpers/files.js
@@ -1,4 +1,4 @@
-//  @ts-check
+/* eslint-disable max-lines */
 const { cpus } = require('os')
 
 const { yellowBright } = require('chalk')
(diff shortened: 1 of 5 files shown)


src/helpers/files.js
```diff
@@ -1,4 +1,4 @@
-//  @ts-check
+/* eslint-disable max-lines */
```
tiffanosaurus, line 1: [sand] We should probably avoid disabling eslint and separate this file into smaller ones moving forward

src/helpers/files.js
```diff
@@ -28,10 +28,36 @@ const matchMiddleware = (middleware, filePath) =>
...
+  return matchesRedirect(file, rewrites.beforeFiles)
+}
+
+exports.matchesRedirect = matchesRedirect
+exports.matchesRewrite = matchesRewrite
 exports.matchMiddleware = matchMiddleware
 exports.stripLocale = stripLocale
 exports.isDynamicRoute = isDynamicRoute
 
+// eslint-disable-next-line max-lines-per-function
```
tiffanosaurus, line 60: [sand] Same

tiffanosaurus (approved):
LGTM!

PR #833, opened 2021-11-23
feat: add experimental support for TTL

ascorbic:
### Summary

Adds experimental support for ttl in pages with revalidate set. Parses the cache-control header returned by Next and extracts the `s-maxage` value, which it then uses in the ttl response. This isnlt enabled by default, but requires an env var flag.

As part of the build process we need to patch the `next-server.js` file, because we need to disable caching for ISR pages and currently this is only supported for `minimalMode`. We do this by adding another env var to the line where it checks for all the conditions for disabling the SSG cache.

### Test plan

1. Visit [a page with revalidate set](https://deploy-preview-833--netlify-plugin-nextjs-demo.netlify.app/getStaticProps/withRevalidate/1/). The rendered-at time will probably be old. If the rendered-at time is the current wall time then it's a fresh deploy so wait a minute for it to be stale then load again.
2. Wait for 1 minute for the revalidated content to propagate to your edge node
3. Load the page again. The rendered-at time should match the time at step 1 (so about a minute in the past)

image


src/templates/handlerUtils.ts
```diff
@@ -31,3 +31,33 @@ export const downloadFile = async (url, destination) => {
...
+  return 0
+}
+
+export const getMultiValueHeaders = (
+  headers: Record<string, string | Array<string>>,
+): Record<string, Array<string>> => {
+  const multiValueHeaders: Record<string, Array<string>> = {}
+  for (const key of Object.keys(headers)) {
+    if (Array.isArray(headers[key])) {
+      multiValueHeaders[key] = headers[key] as Array<string>
```
tiffanosaurus, line 57: You could do the following instead so you don't need to coerce

```ts
    const header = headers[key];

    if (Array.isArray(header)) {
      multiValueHeaders[key] = header
    } else {
      multiValueHeaders[key] = [header]
    }
```

src/helpers/files.js
```diff
@@ -219,6 +219,57 @@ exports.moveStaticPages = async ({ netlifyConfig, target, i18n }) => {
   }
 }
 
+const patchFile = async ({ file, from, to }) => {
```
tiffanosaurus, line 222: ooo neat

tiffanosaurus (approved):
Sorry for the wait, I thought I clicked the thing. Looks awesome 😮!

src/helpers/files.js
ascorbic: I hate it!

PR #837, opened 2021-11-24
fix: handle missing i18n object

ascorbic:
### Summary

Handle situation where the i18n object is missing in the normalised config


tiffanosaurus (approved):
F

PR #856, opened 2021-11-30
fix: copy public directory output instead of input when using Nx

lourd:
### Summary

When copying the `public` directory, the build plugin should use the one that Next.js generates as part of the build, not the input one, as there may be files added to the directory as part of the build process. For example, using the `InjectManifest` plugin from `workbox-webpack-plugin` like this to produce the specifically named and located file `service-worker.js`:

```js
// next.config.js
module.exports = {
  webpack: (config, options) => {
    if (!options.dev) {
      const workboxManifestPlugin = new InjectManifest({
        swSrc: "./src/service-worker.ts",
        swDest: "../public/service-worker.js",
      })
      config.plugins.push(workboxManifestPlugin)
    }
  }
}
```

### Test plan

Added a test to the Nx cypress setup. Steps to run:

```sh
cd demos/nx-next-monorepo-demo
npm install
npm run build
npx nx run demo-monorepo:serve:production --port=3000
```

Then in a separate terminal, from the repo root

```
npm run cy:open
```

Then click the `general.spec.ts.` under the `nx` folder to run it.

IMG_2478

### Standard checks:

- [ ] Add docs when necessary

---


src/index.js
```diff
@@ -47,14 +47,14 @@ module.exports = {
 
     checkNextSiteHasBuilt({ publish, failBuild })
 
-    const { appDir, basePath, i18n, images, target, ignore } = await getNextConfig({ publish, failBuild })
+    const { appDir, basePath, i18n, images, target, ignore, outdir } = await getNextConfig({ publish, failBuild })
```
ascorbic, line 50: Can you explain where you're getting this config name? I'm not aware of `outdir` being a standard config value.

ascorbic:
The `public` dir is used as-is, not generated by Next. It's just meant to be used for static files. If you want to generate other files you should be creating them in `.next`. Are you referring to somethign else?

lourd:
Ah, looks like this is a difference when using Next with Nx. Nx copies the public directory over to the specified output directory, where it also puts the `.next` directory. I just tried a project started with create-next-app and it doesn't have `config.outdir` in required-server-files.json, and it doesn't create a new `public` directory, even when specifying an alternative `distDir`.

The issue with putting generated files in the generated `static` directory is that they're then served from `/static` or `/_next/static`. Some files need to served from root — for example, the `/.well-known/apple-app-site-association` that makes deep linking work on iOS. And some of those files get modified or created at build time. What's a good way to go about doing that that works with Netlify?

Would you support updating the PR to conditionally use/join `outdir` to `appDir` if it exists, to support this Nx-specific case?

lourd:
I went ahead and made it conditional with a comment explaining the case. I also made a Nx workspace with a Next.js project to demonstrate -- https://github.com/lourd/nx-next-public-example. To see what I'm talking about, clone it and run `yarn build`. It will generate a directory at `dist/apps/next-public-test` which as the `.next` directory, a `public` directory, and `next.config.js`. If you look in the `required-server-files.json` you'll see the `outdir` property with a value of `"../../dist/apps/next-public-test"`. And the `public` directory has the files that were in `public` to start with (`star.svg`, `nx-log-white.svg`) as well as the `service-worker.js` file that was specified added by the `InjectManifest` plugin as specified in the webpack setup in `next.config.js`.

If you want, you can also run the build to see the service worker getting registered and loaded from that public location. The command for that is `yarn nx run next-public-test:serve:production`. You'll only see the service worker logs the first time the page loads — to see it on subsequent loads you need to unregister it through the browser's dev tools.

lourd:
I also just added a test to the Cypress suite for `nx-next-monorepo-demo`. Updated the test plan in the PR description with steps to run it. Please let me know if there's anything else I can do!

src/index.js
lourd: It's added when you're using Next.js with Nx. You can see the [source code adding it here](https://github.com/nrwl/nx/blob/67ff4acea20300de7ae316a957d310cd4fb40e5e/packages/next/src/utils/config.ts#L158-L168)

cypress/integration/nx/general.spec.ts
```diff
@@ -9,4 +9,8 @@ describe('Default site', () => {
     cy.visit('//')
     cy.url().should('eq', `${Cypress.config().baseUrl}/`)
   })
-})
\ No newline at end of file
+
+  it('serves generated public files', () => {
+    cy.request(`service-worker.js`)
```
tiffanosaurus, line 14: You can be more explicit with the assetion here by making sure the status returned is a 200. Something like this:
```suggestion
    cy.request('service-worker.js').then((res) => {
      expect(res.status).to.eq(200)
    })
```

tiffanosaurus (approved):
Thank you for the test steps, comments and taking the time to explain the nx use case, super appreciated! I tested using your test steps with and without the outdir as well as using `netlify deploy --build` and things seem to be behaving as expected 🎉 .

cypress/integration/nx/general.spec.ts
lourd: 👍  went ahead and added that and a check for the content-type (in case it's serving some fallback html instead of js!)

lourd:
😃 For sure! Please let me know if there's anything else this needs to be merged.

tiffanosaurus (approved):

PR #860, opened 2021-11-30
chore: 🚙 type some helper files and constants

tiffanosaurus:
### Summary

Our Netlify Next Plugin is written in mostly JS, but we'd like it to be using TypeScript for readability and prevent some bugs which we've had that could have been avoided with TS. This PR changes a couple files to use TypeScript, as I go through the code and get more familiar with it.

### Point of entries

`src/helpers/index.js` is one of the bigger files, and uses many of the other helpers. Thus I decided to go through the rest of the helpers to minimize impact.

The files touched by this conversion:
- `src/constants.js`
- `src/helpers/cache.js`
- `src/helpers/config.js`

### Test plan

1. Visit the Deploy Preview, see that everything works as expected

### Standard checks:

- [ ] Add docs when necessary

---


src/helpers/config.ts
```diff
@@ -123,6 +131,8 @@ exports.generateRedirects = async ({ netlifyConfig, basePath, i18n }) => {
       from: `${basePath}/*`,
       to: HANDLER_FUNCTION_PATH,
       status: 200,
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
tiffanosaurus, line 135: I left that in as I did the first pass to TS. I figured we'd want to get this in incrementally and worry too much about problematic missing types for now.

src/helpers/config.ts
```diff
@@ -158,14 +168,16 @@ exports.generateRedirects = async ({ netlifyConfig, basePath, i18n }) => {
...
 }
 
-exports.getNextConfig = async function getNextConfig({ publish, failBuild = defaultFailBuild }) {
+export const getNextConfig = async function getNextConfig({ publish, failBuild = defaultFailBuild }) {
   try {
-    const { config, appDir, ignore } = await readJSON(join(publish, 'required-server-files.json'))
+    const { config, appDir, ignore }: RequiredServerFiles = await readJSON(join(publish, 'required-server-files.json'))
     if (!config) {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
tiffanosaurus, line 176: Same here, expects a second object and I just ignored it as I passed through :P

src/helpers/config.ts
```diff
@@ -158,14 +168,16 @@ exports.generateRedirects = async ({ netlifyConfig, basePath, i18n }) => {
...
       return failBuild('Error loading your Next config')
     }
     return { ...config, appDir, ignore }
-  } catch (error) {
+  } catch (error: unknown) {
```
tiffanosaurus, line 180: It is best practice to have errors as `unknown` since there's no good way to predict the type of an error

src/helpers/config.ts
```diff
@@ -1,25 +1,29 @@
...
 
-const { yellowBright } = require('chalk')
-const { readJSON, existsSync } = require('fs-extra')
-const { outdent } = require('outdent')
-const { join, dirname, relative } = require('pathe')
-const slash = require('slash')
+import { NetlifyConfig } from '@netlify/build'
+import { yellowBright } from 'chalk'
+import { readJSON } from 'fs-extra'
+import { PrerenderManifest } from 'next/dist/build'
```
ascorbic, line 6: Can this be `import type`?

src/helpers/config.ts
```diff
@@ -54,8 +58,12 @@ const getNetlifyRoutes = (nextRoute) => {
   return netlifyRoutes
 }
 
-exports.generateRedirects = async ({ netlifyConfig, basePath, i18n }) => {
-  const { dynamicRoutes, routes: staticRoutes } = await readJSON(
+export const generateRedirects = async ({ netlifyConfig, basePath, i18n }: { 
+  netlifyConfig: NetlifyConfig, 
+  basePath: string, 
+  i18n
```
ascorbic, line 64: You can get this type by importing the `NextConfig` type from `next`

src/helpers/config.ts
ascorbic: Is this type missing in `@netlify/build` then?

src/helpers/config.ts
```diff
@@ -180,7 +192,7 @@ const resolveModuleRoot = (moduleName) => {
 
 const DEFAULT_EXCLUDED_MODULES = ['sharp', 'electron']
 
-exports.configureHandlerFunctions = ({ netlifyConfig, publish, ignore = [] }) => {
+export const configureHandlerFunctions = ({ netlifyConfig, publish, ignore = [] }) => {
```
ascorbic, line 195: Can these args be typed?

src/helpers/requiredServerFilesType.ts
```diff
@@ -0,0 +1,129 @@
+/* eslint-disable eslint-comments/disable-enable-pair, @typescript-eslint/no-empty-interface, no-use-before-define */
```
ascorbic, line 1: Ah, nice. Maybe we should contribute this upstream!

src/helpers/cache.ts
```diff
@@ -1,23 +1,22 @@
-const {
-  posix: { join },
-} = require('path')
+import { posix } from 'path'
+
+export const restoreCache = async ({ cache, publish }) => {
```
ascorbic, line 3: I htink it's a good idea if we ensure that all exported functions have typed args and return values, even if they're just defined inline.

ascorbic (approved):
Fantastic! Thanks so much for doign this.

test/__snapshots__/index.js.snap
```diff
@@ -70,50 +70,7 @@ exports.resolvePages = () => {
...
-  "fr/getStaticProps/env.json",
-  "fr/getStaticProps/static.html",
-  "fr/getStaticProps/static.json",
-  "fr/image.html",
-  "fr/previewTest.html",
-  "fr/previewTest.json",
-  "fr/static.html",
-]
-`;
+exports[`onBuild() generates static files manifest 1`] = `Array []`;
```
ascorbic, line 73: You updated the snapshot after running `ntl build`. This is why I added the updateSnapshot npm script.

src/helpers/config.ts
tiffanosaurus: I can't do `import { PrerenderManifest } from 'next/types'` since they don't have it in their types, is that what you mean?

src/helpers/config.ts
tiffanosaurus: oo thanks! Wasn't sure where it was coming from

src/helpers/config.ts
tiffanosaurus: Nah just one that would require a bit of refactoring of the function this ts-ignore refers to

ascorbic (approved):
snapidoo!

src/helpers/requiredServerFilesType.ts
tiffanosaurus: Good call! Although I'm sure the definition file is missing some types since it was just auto-generated based on the demo's `required-server-files.json`

src/helpers/requiredServerFilesType.ts
```diff
@@ -0,0 +1,129 @@
...
+    ignore?:  string[];
+}
+
+export interface Env {
+}
+
+export interface Config {
+    env?:                         Env;
+    webpack?:                     null;
+    webpackDevMiddleware?:        null;
```
tiffanosaurus, line 22: ^ these will have to be manually updated, since it's clearly wrong

src/helpers/config.ts
ascorbic: No, I just mean use the type import syntax to make it clear that you're only importing the type:

`import type { PrerenderManifest } from 'next/dist/build'`

tiffanosaurus: Ohhh TIL

PR #861, opened 2021-12-01
fix: move locale detection to netlify redirects

### LINKED ISSUE #859: feat: handle locale autodetection in Netlify redirects
We should implement Next.js locale handling using Netlify redirects. If there are multiple locales and auto-detection is enabled, we should add the following redirects:

- If browser language does not match the default, but _does_ match one of the other locales, then redirect all requests to `/` to the appropriate lcoale prefix.
- If `NEXT_LOCALE` cookie is set, then rewrite to `/.netlify/functions/___netlify-handler` instead (to skip the locale detection)

ascorbic:
### Summary
Currently, locale detection requires the request to reach the handler function, which means it breaks for static or ODB routes. This PR moves the detection into Netlify redirects, so it happens before other redirects are processed.

This PR also refactors the redirects into a different file, because it was getting unwieldy keeping it inside the config helper

### Test plan

1. TODO: create a test site with a static or fallback page at the root
2. Visit the page with a browser locale of "es"
3. Observe a redirect to `/es/`

image

Fixes #859

### Standard checks:

- [ ] Add docs when necessary

---


src/helpers/redirects.ts
```diff
@@ -0,0 +1,164 @@
...
+    if (locale === i18n.defaultLocale) {
+      return
+    }
+    redirects.push({
+      from: `${basePath}/`,
+      to: `${basePath}/${locale}/`,
+      status: 301,
+      conditions: {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore Incorrect type
```
tiffanosaurus, line 41: So many ignores

worried kermit gif

ascorbic: I'm going to fix the `@netlify/build` types

ascorbic: https://github.com/netlify/build/pull/3908

tiffanosaurus (approved):

tiffanosaurus:
Using the test site, I'm not seeing the redirect

ascorbic:
I'm going to merge this, because even though the libredirect issue means the autodetection based on browser header isn;t working, it's better than the current behaviour

andykenward:
@ascorbic Is the next.js built in autodetection still not supported?

PR #864, opened 2021-12-02
fix: correct handling of data route JSON files

ascorbic:
### Summary

Currently, the JSON files for statically-rendered routes were not being moved into the correct location for loading from the CDN. This PR moves them into `_next/data/build-id/`, rather than alongside the HTML. 

It also fixes a bug in the fielsystem patching funcitons that load missing files from the CDN when required by Next. They were failing because even though they downloaded properly, Next would run `stat()` which would fail.

This adds lots of links to the test site index page, so that we can load all the different sorts of test page.

### Test plan

1. Visit [a rewrite route that points to a static page](https://deploy-preview-864--netlify-plugin-nextjs-demo.netlify.app/rewriteToStatic/) and observe that it doesn't 404
2. Load [a static data file](https://deploy-preview-864--netlify-plugin-nextjs-demo.netlify.app/_next/data/build-id/en/getStaticProps/1.json) and check the response headers to see that there is no `x-render-mode`, meaning that it was served from the CDN.

image

### Standard checks:

- [ ] Add docs when necessary

---


tiffanosaurus (approved):
🔥

PR #878, opened 2021-12-03
chore: convert more files to ts

ascorbic:
### Summary

Converts `functions` and the handler template files to TypeScript. Moves some of the handler code into the helper file.

image

### Standard checks:

- [ ] Add docs when necessary

---


src/templates/getHandler.ts
```diff
@@ -0,0 +1,125 @@
...
+const path = require('path')
+// eslint-disable-next-line node/prefer-global/url, node/prefer-global/url-search-params
+const { URLSearchParams, URL } = require('url')
+
+const { Bridge } = require('@vercel/node/dist/bridge')
+
+const { augmentFsModule, getMaxAge, getMultiValueHeaders, getNextServer } = require('./handlerUtils')
+
+const makeHandler =
+  () =>
```
tiffanosaurus, line 15: [sand] You can add a return type here, but you'll probably have to function type since it's a long one

src/templates/getHandler.ts
```diff
@@ -0,0 +1,125 @@
...
+  // We return a function and then call `toString()` on it to serialise it as the launcher function
+  // eslint-disable-next-line max-params
+  (conf: NextConfig, app, pageRoot, staticManifest: Array<[string, string]> = [], mode = 'ssr') => {
```
tiffanosaurus, line 18: [dust] missing some return types here which makes sense and can be addressed later

src/templates/getPageResolver.ts
```diff
@@ -0,0 +1,32 @@
...
+import { outdent } from 'outdent'
+import slash from 'slash'
+import glob from 'tiny-glob'
+
+import { HANDLER_FUNCTION_NAME } from '../constants'
+
+// Generate a file full of require.resolve() calls for all the pages in the
+// build. This is used by the nft bundler to find all the pages.
+
+export const getPageResolver = async ({ netlifyConfig, target }) => {
```
tiffanosaurus, line 12: [sand] Same, can be typed here as a follow up.

Good indicator that we should enable the eslint rule for typing functions, I'll add that to the board

src/templates/handlerUtils.ts
```diff
@@ -63,3 +67,110 @@ export const getMultiValueHeaders = (
...
   return multiValueHeaders
 }
+
+export const augmentFsModule = ({
+  promises,
+  staticManifest,
+  pageRoot,
+  getBase,
+}: {
+  promises: typeof fs.promises
```
tiffanosaurus, line 77: [dust] You can also import `promises` from `fs and use `typeof promises`

Huh I would have thought there was a type you could refer to directly but doesnt look like it

tiffanosaurus (approved):

src/templates/getPageResolver.ts
ascorbic: Yeah, good point. I'd say add the one to require it at module boundaries at least

src/templates/handlerUtils.ts
ascorbic: ikr. weird

ascorbic: I didn't do that, because I'm using `promises` there and don't like shadowing imports

Issue #839, opened 2021-11-24
[Chore]: Add extensive test cases (e2e)


tiffanosaurus:
Parent issue to keep track of all test cases/sites we want to have covered via Cypress. We will want to test most features of Next to ensure we don't break anything. Feel free to add more test cases.

## Default site
- [X] Set up cypress
- [x] Default site
- [x] `basePath`
- [x] Preview mode
- [x] i18n
- [x] Headers
- [x] Rewrites/redirects
- [x] `basePath` + i18n
- [x] Trailing slash
- [x] Catch-all/optional catch-all, getxxprops
- [x] Images
- [x] Dynamic imports
- [x] Custom 404/500
- [x] Static assets in `public`
- [x] cookie header locale
- [ ] middleware
- [ ] advanced rewrite rules

## Monorepo site
- [ ] Custom dist
- [ ] Monorepo: `next` run in root with site root in subdirectory

## Nx site
- [ ] nx create-app

nickytonline:
@tiffanosaurus, I'm going to go ahead and remove you from this as you've switched teams. Whoever picks this up, there's still some E2E tests to complete.

nickytonline:
Hi @netlify/frameworks! We've been working on the Next.js plugin quite a bit lately. Just wondering if any of that work involved E2E tests for the middleware and advanced rewrite rules?

orinokai:
Hey @nickytonline, no afaik we haven't added tests yet for those features. I'm hoping to chat to Erica later this week about a few testing things so I'll roll this into it.

MarcL:
@orinokai - Where are we at with this work? I was wondering if it's worth splitting this up into more manageable tickets if there's lots left to do.

orinokai:
I had planned to give the tests a good review after I finished the fallback work because the tests would need to be updated, but both aspects of that work are currently blocked and they should probably be a separate issue once the work is complete away. I'll make a start on this ticket this week.

PR #899, opened 2021-12-06
chore: add to cypress dashboard for all demo sites

tiffanosaurus:
### Summary

Fail status check on cypress failure for all three demo sites and record to cypress dashboard. 

Before this PR, cypress tests would run `onSuccess`, and the only way to tell if something failed was to check the deploy logs. Now there are a couple new ways, including comments and the cypress dashboard (in the deploy logs, it also shows a link of the recorded run). You are also able to access video recording of runs which is helpful to debug failed runs.

Keep in mind we are in the free tier here, and we'll have to see if we need to upgrade
CleanShot 2021-12-09 at 12 22 02@2x

### Test plan
See cypress comments below and deploy logs

#839 

### Standard checks:

- [ ] Add docs when necessary

---


ascorbic:
LOVE THIS!!

ascorbic (approved):

PR #955, opened 2021-12-10
chore: remove "environment" from release-please workflow

ascorbic:
Turns out I didn't properly understand how this works, so let's switch back to the "normal" way to do this then ask for advice from one of our GH experts


tiffanosaurus (approved):
oki doki

PR #958, opened 2021-12-10
chore: fix accept-encoding test

ascorbic:
### Summary

Currently every build fails on the Cyporess Accept-Encoding test, but that's because of a libredirect bug not anything in the plugin. This PR changes that test so that ti does pass now, though we should change it back once the redirects work properly so that it's more realsitic.

### Test plan

Check the static site cypress tests pass

### DIFF
--- a/cypress/integration/static-root/i18n.spec.ts
+++ b/cypress/integration/static-root/i18n.spec.ts
@@ -27,7 +27,9 @@ describe('Localization', () => {
   it('should use Accept-Language to choose a locale', () => {
     cy.visit('/', {
       headers: {
-        'Accept-Language': 'fr-FR,fr;q=0.5',
+        // FIXME: switch back once libredirect bug is fixed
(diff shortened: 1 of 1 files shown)


cypress/integration/static-root/i18n.spec.ts
```diff
@@ -27,7 +27,9 @@ describe('Localization', () => {
   it('should use Accept-Language to choose a locale', () => {
```
tiffanosaurus, line 27:
```suggestion
  it.skip('should use Accept-Language to choose a locale', () => {
```

`it.skip` should work

ascorbic: I thought rather than skipping entirely, I'd change it to at least test _something_, so it shows that the plugin redirects themselves are working, even if the libredirect part is broken

tiffanosaurus (approved):
👍 thanks for clarifying

PR #1032, opened 2021-12-18
Update large-functions.md

chrishannaby:
Removing this line from the docs since I don't think it applies now that MOVE_STATIC_FILES is the default behaviour

> - **Large numbers of pre-rendered pages** If you have a very large number of pre-rendered pages, these can take up a lot
  of space in the function. There are two approaches to fixing this. One is to consider deferring the building of the
  pages. If you return `fallback: "blocking"` from `getStaticPaths` the rendering will be deferred until the first user
  requests the page. This is a good choice for low-traffic pages. It reduces build and deploy time, and can make your
  bundle a lot smaller.

### Test plan

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/docs/large-functions.md
+++ b/docs/large-functions.md
@@ -18,9 +18,3 @@ solution. The list of largest files shown in the build logs will help you see wh
 
   If you do need large modules at runtime (e.g. if you are running Puppeteer in a Next API route), consider changing to
   a Netlify function which will have less overhead than the equivalent Next.js function.
-
-- **Large numbers of pre-rendered pages** If you have a very large number of pre-rendered pages, these can take up a lot
-  of space in the function. There are two approaches to fixing this. One is to consider deferring the building of the
-  pages. If you return `fallback: "blocking"` from `getStaticPaths` the rendering will be deferred until the first user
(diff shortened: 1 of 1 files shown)


sjBao:
Hello I'm tryin to dig around to find out why my generated zip file is so large. I tried setting fallback: "blocking" for every instance of `getStaticPaths` but still running into this issue: 
```bash
Next.js cache saved.
The function zip .netlify/functions/___netlify-odb-handler.zip size is 120 MB, which is larger than the maximum supported size of 52.4 MB.
There are a few reasons this could happen. You may have accidentally bundled a large dependency, or you might have a
large number of pre-rendered pages included.
Contains 4315 files
```

ascorbic:
Hi @sjBao 
What does it show as the largest files in the log?

ascorbic:
@chrishannaby I'm holding off merging this, because there are a few subtleties that I need to work out and document at the same time. Not every static file is moved to the CDN, so we do need to mention the details of that.

sjBao:
@ascorbic sorry for the late reply, I think updating to the latest next plugin and netlify-cli fixed this issue for me

chrishannaby:
As @ascorbic said earlier

Issue #1008, opened 2021-12-16
[Bug]: `_redirects` entry not included in `.next`


itsnotrisky:
### Steps to reproduce

1. Change plugin version to v4 from "Plugin" menu and selecting Options dropdown.
2. Change "Publish directory" to `.next` from "Site settings".
3. Retry deploy from Builds menu.

### Plugin version

4.0.0

### More information about your build

- [ ] I am building using the CLI
- [ ] I am building using file-based configuration (`netlify.toml`)

### Your `netlify.toml` file

<details>
  <summary>`netlify.toml`</summary>
  
```toml
# Paste content of your `netlify.toml` file here
```

</details>

### Relevant log output (or link to your logs)

<details>
  <summary>Outdated plugin notification from first build</summary>
  
```
1:14:58 PM: ❯ Outdated plugins
1:14:58 PM:    - @netlify/plugin-nextjs@3.9.2: latest version is 4.0.0
1:14:58 PM:      Migration guide: https://ntl.fyi/next-plugin-migration
1:14:58 PM:      To upgrade this plugin, please uninstall and re-install it from the Netlify plugins directory (https://app.netlify.com/plugins)
```

</details>

<details>
  <summary>Build logs with v4</summary>
  
```
1:25:22 PM: ────────────────────────────────────────────────────────────────
1:25:22 PM:   3. @netlify/plugin-nextjs (onBuild event)                     
1:25:22 PM: ────────────────────────────────────────────────────────────────
1:25:22 PM: ​
1:25:22 PM: Moving static page files to serve from CDN...
1:25:22 PM: Moved 7 files
1:25:22 PM: Netlify configuration property "redirects" value changed to [
1:25:22 PM:   {
1:25:22 PM:     from: '/_next/image*',
1:25:22 PM:     query: { url: ':url', w: ':width', q: ':quality' },
1:25:22 PM:     to: '/_ipx/w_:width,q_:quality/:url',
1:25:22 PM:     status: 301
1:25:22 PM:   },
1:25:22 PM:   { from: '/_ipx/*', to: '/.netlify/builders/_ipx', status: 200 },
1:25:22 PM:   { from: '/cache/*', to: '/404.html', status: 404, force: true },
1:25:22 PM:   { from: '/server/*', to: '/404.html', status: 404, force: true },
1:25:22 PM:   { from: '/serverless/*', to: '/404.html', status: 404, force: true },
[...]

PR #1078, opened 2021-12-23
chore: warn for incorrect _redirects file

ascorbic:
### Summary

Adds a warning for `_redirects` files in the site root. We used to silently add these to the user's redirects, but we now follow the documented behaviour, which is that the files should go in the publish dir not site root. #1130 adds this to the docs, and the CLI warns about it too, but this adds Next.js-specific logs, with instructions on where it needs to go.

### Test plan

1. Add a `_redirects` file to the root of a demo site
2. Run a build

Issue that shows the problem: #1008 
capybara

### Standard checks:

- [ ] Add docs when necessary

---


src/helpers/verification.ts
```diff
@@ -187,4 +187,14 @@ export const warnForProblematicUserRewrites = ({
     `),
   )
 }
+
+export const warnForRootRedirects = ({ appDir }: { appDir: string }) => {
+  if (existsSync(join(appDir, '_redirects'))) {
+    console.log(
+      yellowBright(
+        `You have a "_redirects" file in your root directory, which is not deployed and will be ignored. If you want it to be used, please move it into "public".`,
```
naelkhann, line 195: 🙌

naelkhann (dismissed):
👍  definitely a small helpful change to the docs

src/helpers/verification.ts
```diff
@@ -187,4 +187,14 @@ export const warnForProblematicUserRewrites = ({
     `),
   )
 }
+
+export const warnForRootRedirects = ({ appDir }: { appDir: string }) => {
```
tiffanosaurus, line 191:
```suggestion
export const warnForRootRedirects = ({ appDir }: { appDir: string }): void => {
```

tiffanosaurus (approved):

PR #1131, opened 2022-01-17
chore: Add tailwind prose formatting for easier use of demos

tiffanosaurus:
### Summary
Going through the demos is kind of a chore because there's no formatting, making it hard to tell what are links and where things are. This PR adds a top level `prose` class from [Tailwind's typography plugin](https://tailwindcss.com/docs/typography-plugin) to automatically format everything in our demo

### Test plan

I mainly did this because writing all the use case tests was kind of annoying

### Standard checks:

- [ ] Add docs when necessary

---


ascorbic:
Do you reckon it's ok to add something external like this to the default demo, as we're trying to test something as vanilla as possible? I guess it shouldn't affect the Next.js code, right?

tiffanosaurus:
Yeah I don't see a problem with it, we already have other demo sites that dont use this too. There's some issue happening with postcss so I'm just going to use a minified css version of the tailwind prose plugin instead too. Closing this in favor of it

PR #1133, opened 2022-01-17
chore: add prose css

tiffanosaurus:
### Summary

Add minified global css for the default demo (using `npx tailwindcss -o demos/default/style/global.css --minify`) for ease of reading and debugging.

See #1131. We can use full on tailwind, but since we're not going to be concentrating on the looks of the demo, the css is not meant to change, thus leading to less of a need for full on tailwind. In order to facilitate readability for other users, we don't want to add too many classnames or components either. 

Before:
CleanShot 2022-01-17 at 17 23 28@2x

After:
CleanShot 2022-01-17 at 17 22 42@2x

### Test plan

1. Visit the [Deploy Preview](https://deploy-preview-1133--netlify-plugin-nextjs-demo.netlify.app/) and see it's nicer to navigate now

### Standard checks:

- [ ] Add docs when necessary

---


ascorbic (dismissed):
Nice! Looks much better

ascorbic (approved):

PR #1172, opened 2022-01-27
chore: add e2e tests and demos

tiffanosaurus:
### Summary
Add demos and e2e tests to the default site for a bunch of next features.

Namely: 
- Environment tokens
- Font handling
- Routing
- CSS handling
- Layouts
- TypeScript
- Scripts
- Images
- Middleware

### Test plan

1. Visit the Default Deploy Preview (`/` should work even though it's typescript)
2. Go to `/css`
3. Go to `/script`
4. Go to `/layouts`
5. Go to `/env`
6. Go to `/font`
7. Go to `/broken-image` (expected broken page)
8. Go to `/middle`
8. See the default demo cypress test run

Example new demo page:
CleanShot 2022-01-27 at 20 08 38@2x

### Standard checks:

- [ ] Add docs when necessary

---


.vscode/feat-test.code-snippets
```diff
@@ -0,0 +1,30 @@
+{
```
tiffanosaurus, line 1: This makes it so in. vscode, you can use `ft` and it will create a snippet for a feature test. I was thinking of adding some auto-generation tooling but decided to create tests with this instead.

ascorbic: Nice!

ascorbic (approved):
Looking good!

ascorbic:
The errors should be fixable by regenrating the package lock

ascorbic:
So close! Needs `npm run test:update`

tiffanosaurus:
Arghh yeah thanks for the reminder

ascorbic:
*cries*

tiffanosaurus:
these snapshots are gonna be the end of me

ascorbic:
So the order of the redirects in the snapshots is based on the order in the prerender manifest. This seems to imply that the ordering there is indeterminate. That's a bit of a nightmare!

ascorbic (dismissed):
OMG

.github/workflows/test.yml
```diff
@@ -12,12 +12,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        node-version: [12, '*']
+        node-version: [16, '*']
```
ascorbic, line 15: This is causing it to not pass. Why is it needed? Doesn't `*` do version 16?

tiffanosaurus, line 15:
```suggestion
        node-version: [12, '*']
```

.github/workflows/test.yml
```diff
@@ -12,12 +12,12 @@ jobs:
...
         exclude:
           - os: macOS-latest
-            node-version: 12
+            node-version: 12.0.10
```
tiffanosaurus, line 18:
```suggestion
            node-version: 12
```

.github/workflows/test.yml
```diff
@@ -12,12 +12,12 @@ jobs:
...
           - os: windows-latest
-            node-version: 12
+            node-version: 12.0.10
```
tiffanosaurus, line 20:
```suggestion
            node-version: 12
```

.github/workflows/test.yml
tiffanosaurus: this was me hitting all the buttons it is not needed

ascorbic (approved):

PR #1236, opened 2022-02-22
added yarn command

sureshdsk:
### Summary
updated readme to add yarn command to install @netlify/plugin-nextjs

(Created as part Hashnode Netlify Hackathon)

### Standard checks:
- [x] Add docs when necessary


ascorbic:
Hi @sureshdsk ! Thanks for your contribution, but I'm afraid we won;t be merging this. We don't as a rule add yarn-specific commands, because the assumption is that anybody who uses yarn is able to convert from the equivalent npm command.

PR #1265, opened 2022-03-16
fix: change working directory to site root in SSR

ascorbic:
### Summary

Next.js expects the cwd to be the site root. In some cases with monorepos that isn't the case, which can cause problems with packages that expect to be able to resolve things such as config files relative to the root. This PR changes the working directory to the site root as part of the handler bootstrap.

### Test plan

1. Visit `/api/hello` in the default deploy preview, and check that `cwd` is the site root.

capybara and babies

### Standard checks:

- [ ] Add docs when necessary

---


krider2010 (approved):
Nice one!

PR #1304, opened 2022-04-11
refactor: move into npm workspace

ascorbic:
### Summary
This PR moves the plugin and one demo site into an npm workspace, to make it easier to manage packages, and to better handle dependency caching.

Before it can be merged  the release please scripts and config need changing to reflect the fact that the plugin isn't in the root.

### Test plan

This will involve ensuring all the demo sites build properly

caviforms

### Standard checks:

- [ ] Add docs when necessary

---


.nvmrc
```diff
@@ -0,0 +1 @@
+17.9.0
```
ericapisani, line 1: Should this perhaps be set to v16 so that this is on an LTS version of node? Or is there something about v17 that we want for these changes? 

Depending on when it's looking like these changes will be merged and released we might be able to even go up to the newest LTS version (v18) which will be released on the April 19th.

ericapisani (commented):
Just one question so far 😄

.nvmrc
ascorbic: Yeah, I did this because I thought it needed to be 17 for workspaces to work properly, but I think 16 should be ok. Good point re 18 though. We should probably migrate later.

I still need to work out the best way to run the node 12 tests though, because right now they're failing as they don't support workspaces at all.

.github/workflows/test.yml
```diff
@@ -23,13 +23,20 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: Use Node.js ${{ matrix.node-version }}
+      - name: Installing with latest Node.js
```
ascorbic, line 26: This is so we can use workspaces, which require node 16, but still run tests in node 12

demos/default/pages/env.tsx
```diff
@@ -1,19 +1,37 @@
...
+        ❌ Private Environment token should not be found: <code>{process.env.TEST_ENV_VAR ?? 'Everything worked'}</code>
+      </p>
+      <p>Rendered at {time}</p>
     </div>
   )
 }
 
-export default EnvTest
\ No newline at end of file
+export const getServerSideProps = async ({ params }) => {
```
ascorbic, line 29: When this was statically rendered it was causing errors, because private env vars _are_ available on the server, so the HTML was mismatched. By SSRing it, we ensure that it's not included.

ericapisani (dismissed):
This LGTM 🚀 Appreciate the workaround to support workspaces with a modern version of node while still running tests in npm v12. Fortunately we won't have to support v12 for too much longer since it will have reached end of life at the end of the month 🥳

.github/workflows/release-please.yml
```diff
@@ -16,25 +16,27 @@ jobs:
         id: release
         with:
           token: ${{ steps.get-token.outputs.token }}
-          release-type: node
-          package-name: '@netlify/plugin-nextjs'
+          command: manifest
```
ascorbic, line 19: We're changing to manifest releases because the package is in a subdirectory

krider2010 (approved):
Nice! Let's 🚢

PR #1309, opened 2022-04-18
feat: ✨ add out-of-the-box support for NextAuth.js  

### LINKED ISSUE #1242: Support NextAuth out of the box
Ensure that NextAuth works without extra config. At a minimum this will involve setting `NEXTAUTH_URL`

ericapisani:
### Summary

Adds out of the box support for [NextAuth.js](https://next-auth.js.org/) by setting the [`NEXTAUTH_URL`](https://next-auth.js.org/configuration/options#nextauth_url) to the URL representing the main address of the Netlify site when the `next-auth` package is detected.

A user will still be required to set the `NEXTAUTH_SECRET` in production in order to encrypt the NextAuth.js JWT and hash email verification tokens. [Related docs](https://next-auth.js.org/deployment)

### Test plan

1. Visit the Deploy Preview ([here](https://ep-next-auth-support.netlify.app/)) ...

Fixes #1242 

hedgehog

### Standard checks:

- [ ] Add docs when necessary

---


plugin/src/helpers/utils.ts
```diff
@@ -174,3 +174,14 @@ export const findModuleFromBase = ({ paths, candidates }): string | null => {
   }
   return null
 }
+
+export const isNextAuthInstalled = (): boolean => {
+  try {
+    // eslint-disable-next-line import/no-unassigned-import, node/no-extraneous-require, import/no-extraneous-dependencies
+    require('next-auth')
```
ericapisani, line 181: This approach follows what's suggested in the npm documentation around [optional dependencies](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#optionaldependencies)

demos/nextAuth-demo/.gitignore
```diff
@@ -0,0 +1 @@
+.env.local
```
nickytonline, line 1: [boulder] There will be other build artifacts and potentially other files that we may want to omit. Running `npx gitignore node` will generate a thorough .gitignore for Node/frontend projects.

demos/nextAuth-demo/LICENSE
```diff
@@ -0,0 +1,15 @@
+ISC License
```
nickytonline, line 1: Just wondering if this is the correct license or not? MIT would make sense to me, but I believe we need to have the same license as projects we reference.

demos/nextAuth-demo/netlify.toml
```diff
@@ -0,0 +1,18 @@
+[build]
+command = "next build"
+publish = ".next"
+ignore = "git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF . ../../plugin"
+
+[build.environment]
+TERM = "xterm"
+NODE_VERSION = "17"
```
nickytonline, line 8: [pebble] Node.js 17 is not an LTS version. Consider an LTS version, e.g. 16. As well, if this is set here, a corresponding [engines version](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines) should be set in package.json.

plugin/src/helpers/utils.ts
```diff
@@ -174,3 +174,14 @@ export const findModuleFromBase = ({ paths, candidates }): string | null => {
...
+  } catch {
+    // Ignore the MODULE_NOT_FOUND error
+    return false
+  }
+  return true
```
nickytonline, line 186: [dust] Consider returning `true` right after `require('next-auth')` instead of after the catch block.

demos/nextAuth-demo/netlify.toml
ericapisani: Node 17 began to be specified within the demo projects as part of the changes to support [npm workspaces](https://github.com/netlify/netlify-plugin-nextjs/pull/1304/files#r847648971) within the project, and the use of an LTS version was brought up as part of the review. 

At the time we deferred the specification of an LTS version until v18 was released (which it now has been) because we weren't going to be waiting that much longer from the time of those changes, but it's likely worth doing all in one go as part of one set of changes rather than making all of those changes within the context of this pull request.

Will make the change to the `engines` version in the meantime though 👍🏻

ericapisani: Mentioned offline but have created an issue tracking the upgrade to Node 18 for all the demo sites within this project so will consider this comment resolved for now

ascorbic:
This looks great! Can we get the demo site setup to build for PRs? I'd say this is a boulder unless there's a good reason not to

ascorbic (approved):
🎉

PR #1310, opened 2022-04-19
feat: add support for Edge Functions

ascorbic:
### Summary
Adds beta support for Netlify Edge Functions. This includes support for deploying middleware to the edge, but also running the full site using edge runtime. To enable this during the beta, set `NEXT_USE_NETLIFY_EDGE` in the UI or cli.

### Test plan

1. Visit the deploy previews for the middleware and server-components demos (coming soon)

loads of capybaras in the sun

### Standard checks:

- [ ] Add docs when necessary

---


plugin/src/index.ts
```diff
@@ -100,6 +97,15 @@ const plugin: NetlifyPlugin = {
       nextConfig: { basePath, i18n, trailingSlash, appDir },
       buildId,
     })
+
+    if (process.env.NEXT_USE_NETLIFY_EDGE) {
+      console.log(outdent`
+        ✨ Deploying to ${greenBright`Netlify Edge Functions`} ✨
+        This feature is in beta. Please share your feedback here: https://ntl.fyi/next-netlify-edge
```
orinokai, line 104: This link seems to just point to the Netlify homepage

plugin/src/templates/edge/ipx.ts
```diff
@@ -0,0 +1,32 @@
...
+import type { Context } from 'netlify:edge'
+
+/**
+ * Implement content negotiation for images
+ */
+
+const handler = async (req: Request, context: Context) => {
+  const { searchParams } = new URL(req.url)
+  const accept = new Accepts(req.headers)
+  const type = accept.types(['avif', 'webp'])
```
orinokai, line 11: Just AVIF and WEBP?

orinokai (dismissed):
Looks awesome to me! Just that minor docs issue and image format question.

plugin/src/index.ts
ascorbic: We'll create this when we launch. There's a bullet in the launch checklist

plugin/src/templates/edge/ipx.ts
ascorbic: Yeah. If they don't support either of those then we don't emforce a type, and let ipx serve the original format

plugin/src/templates/edge/ipx.ts
```diff
@@ -0,0 +1,32 @@
...
+    return new Response('Invalid request', {
+      status: 400,
+    })
+  }
+
+  const modifiers = [`w_${width}`, `q_${quality}`]
+
+  if (type) {
+    modifiers.push(`f_${type}`)
+  }
```
ascorbic, line 27: ...here is where we add the modifier if they support a modern format

orinokai: Got it, thanks 👍🏼

orinokai (approved):
🚢

PR #1325, opened 2022-04-26
fix: handle absolute rewrite URLs

ascorbic:
### Summary

When rewriting a URL in middleware, it attaches an `x-middleware-rewrite` header with the target URL. This is used both for server-side rewrites, but also for client-side preflights. The returned value can be either a relative or absolute URL. Previously we passed-through the header unchanged. This caused problems for two reasons. First, if it is an absolute URL it is treated as external by the router on the client, even if it is not, which caused Next to get into an infinite reload loop as it tried to redirect to itself. For actual external URLs it was failing because Netlify Edge Functions don't support rewriting to external URLs.

This fixes these problems in two parts:

- If the absolute URL is actually local then it makes it into a relative URL using [the same algorithm as Next.js](https://github.com/vercel/next.js/blob/canary/packages/next/shared/lib/router/utils/relativize-url.ts#L6)
- If the URL is external, then instead of using `context.rewrite()` it instead calls `fetch()`, adds the appropriate headers to the response, and returns the response. 

### Test plan

1. Visit [the middleware deploy preview](https://deploy-preview-1325--next-plugin-edge-middleware.netlify.app/)
2. View the absolute URL demo. Check that the page loads correctly and that the `x-middleware-rewrite` headers is relative
3. Visit the external URL demo. Check that it returns the proxied content of "example.com". Check that the `x-middleware-rewrite` header is "http://example.com".

Fixes issue mentioned in [this comment](https://github.com/netlify/netlify-plugin-nextjs/issues/1321#issuecomment-1109184031)

black capybara

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/plugin/src/templates/edge/utils.ts
+++ b/plugin/src/templates/edge/utils.ts
@@ -5,6 +5,19 @@ export interface FetchEventResult {
   waitUntil: Promise<any>
 }
 
+/**
+ * This is how Next handles rewritten URLs.
+ */
+ export function relativizeURL(url: string | string, base: string | URL) {
+  const baseURL = typeof base === 'string' ? new URL(base) : base
+  const relative = new URL(url, base)
+  const origin = `${baseURL.protocol}//${baseURL.host}`
+  return `${relative.protocol}//${relative.host}` === origin
+    ? relative.toString().replace(origin, '')
+    : relative.toString()
+}
+
+
 export const addMiddlewareHeaders = async (
   originResponse: Promise<Response> | Response,
   middlewareResponse: Response,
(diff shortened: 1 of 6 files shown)


plugin/src/templates/edge/utils.ts
```diff
@@ -5,6 +5,19 @@ export interface FetchEventResult {
   waitUntil: Promise<any>
 }
 
+/**
+ * This is how Next handles rewritten URLs.
```
ericapisani, line 9: [dust] Might be handy to include the same link that was provided in the pull request summary here so it makes it easier to folks to find the relevant code that's being referred to here

ericapisani (approved):
Looking good to me 🚀 

Also as a [pebble] comment - would be good to add test coverage here as well. Based on existing coverage it looks like Cypress might be the way to go but if there's a different approach that we want to take in this case that's fine too.

PR #1381, opened 2022-06-10
ci: build demos whenever plugin changes or cache cleared

ascorbic:
### Summary

Updates the demos' `ignore` commands so that the sites will all build if the plugin changes. It is different from before because it also will now build if the two commit hashes are the same. This is the case when the cache is cleared. Previously this would return a value of unchanged, meaning it wasn't possibel to manually build the demo sites.

### Test plan

1. Check that the sites build

capybara

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/demos/base-path/netlify.toml
+++ b/demos/base-path/netlify.toml
@@ -1,7 +1,7 @@
 [build]
 command = "next build"
 publish = ".next"
-ignore = "git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF ../../"
+ignore = "if [ $CACHED_COMMIT_REF == $COMMIT_REF ]; then (exit 1); else git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF . ../../plugin; fi;"
 
 [build.environment]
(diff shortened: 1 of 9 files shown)


demos/base-path/netlify.toml
```diff
@@ -1,7 +1,7 @@
 [build]
 command = "next build"
 publish = ".next"
-ignore = "git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF ../../"
+ignore = "if [ $CACHED_COMMIT_REF == $COMMIT_REF ]; then (exit 1); else git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF . ../../plugin; fi;"
```
nickytonline, line 4: Just putting a note for folks if they're not sure why we're doing an `exit 1`. See https://docs.netlify.com/configure-builds/ignore-builds/#custom-ignore-command

nickytonline (approved):

sarahetter (approved):

PR #1392, opened 2022-06-14
fix: check for existence of experimental images

sarahetter:
### Summary

Fixes error if next does not add an `images` object to `experimental`.

### Test plan

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/plugin/src/index.ts
+++ b/plugin/src/index.ts
@@ -72,22 +72,17 @@ const plugin: NetlifyPlugin = {
 
     checkNextSiteHasBuilt({ publish, failBuild })
 
-    const {
-      appDir,
-      basePath,
-      i18n,
-      images,
-      target,
-      ignore,
-      trailingSlash,
-      outdir,
-      experimental: {
-        images: { remotePatterns },
+    let experimentalRemotePatterns = []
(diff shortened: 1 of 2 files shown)


plugin/src/index.ts
```diff
@@ -72,22 +72,17 @@ const plugin: NetlifyPlugin = {
...
+        failBuild,
       },
-    } = await getNextConfig({
-      publish,
-      failBuild,
-    })
+    )
+
+    if (experimental.images) {
+      experimentalRemotePatterns = experimental.images.remotePatterns
```
ericapisani, line 84: [potential boulder] Is it possible that `remotePatterns` might not be defined even if `experimental.images` exists? 

If there's a dependency on `experimentalRemotePatterns` needing to be an array, then it might be worth doing the following as an added measure of safety

```suggestion
      experimentalRemotePatterns = experimental.images.remotePatterns ?? []
```

ericapisani (commented):
Other than the one question the change makes sense. I'd like to see a test case for this behaviour, but given folks are currently blocked on this change if they're installing the latest version of the plugin I'm ok with seeing that in a follow up

ericapisani (approved):
🚀

PR #1405, opened 2022-06-21
chore: added cypress/screenshots to git ignore

nickytonline:
### Summary

Cypress snapshots were showing up in my git changes, so adding the folder to the .gitignore.

### Test plan

1. This doesn't changed code, so all existing previews should deploy and tests pass.

### Standard checks:

- [ ] Add docs when necessary

---

Someone sweeping up a cat jokingly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -147,3 +147,6 @@ Temporary Items
 demos/default/.next
 .parcel-cache
 plugin/lib
+
+# Cypress
+cypress/screenshots
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,7 @@
         "@types/jest": "^27.0.2",
         "@types/mocha": "^9.0.0",
         "@types/node": "^17.0.10",
-        "@types/react": "^17.0.38",
+        "@types/react": "^17.0.43",
         "babel-jest": "^27.2.5",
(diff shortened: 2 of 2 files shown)


sarahetter (approved):
LGTM, but as sand: is there any way to turn off these screenshots?

nickytonline:
> LGTM, but as sand: is there any way to turn off these screenshots?

You can. It's a setting in the Cypress configuration file. It's on to capture errors as screenshots. You can have a different configuration for local development if you want to get rid of them for local dev. We can do that, but it'd mean creating separate config files for each demo site locally and when you call them, load a config for the development environment.

Here's an example of how to run it based on the environment variable. https://github.com/forem/forem/blob/main/cypress/plugins/index.js

ascorbic:
I think the screenshots are helpful in general. The problem is ones like this, where they're failing for some weird reason. I've seen this before, but don't know why it happens. Why is it showing app.netlify.com, instead of the demo site?

sarahetter:
> Why is it showing app.netlify.com, instead of the demo site?
This is because of the cancelled build previews (due to the netlify.toml `ignore`, as we were just talking about on slack)
If the build fails, then the action gets passed app.netlify.com instead of the actual build preview url.

PR #1435, opened 2022-06-29
fix: check for Middleware manifest & chain

sarahetter:
### Summary

Adds a check for sortedMiddleware to fix error here: https://app.bugsnag.com/netlify/netlify-build/errors/62bc82c51f3d1700080caab6?event_id=62bc82c50095183a54e70000&i=sk&m=nw

### Test plan

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/plugin/src/index.ts
+++ b/plugin/src/index.ts
@@ -148,15 +148,11 @@ const plugin: NetlifyPlugin = {
 
     const middlewareManifest = await loadMiddlewareManifest(netlifyConfig)
 
-    if (
-      !process.env.NEXT_USE_NETLIFY_EDGE &&
-      middlewareManifest.sortedMiddleware &&
-      middlewareManifest.sortedMiddleware.length !== 0
-    ) {
+    if (!process.env.NEXT_USE_NETLIFY_EDGE && middlewareManifest?.sortedMiddleware?.length) {
       console.log(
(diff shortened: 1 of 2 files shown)


plugin/src/index.ts
```diff
@@ -148,7 +148,7 @@ const plugin: NetlifyPlugin = {
 
     const middlewareManifest = await loadMiddlewareManifest(netlifyConfig)
 
-    if (!process.env.NEXT_USE_NETLIFY_EDGE && middlewareManifest.sortedMiddleware.length !== 0) {
+    if (!process.env.NEXT_USE_NETLIFY_EDGE && middlewareManifest?.sortedMiddleware?.length) {
```
nickytonline, line 151: [dust] Consider being explicit about the length instead of coercing to boolean.

```suggestion
    if (!process.env.NEXT_USE_NETLIFY_EDGE && middlewareManifest?.sortedMiddleware?.length > 0) {
```

nickytonline (approved):
Left a suggestion, but otherwise good to go.

plugin/src/index.ts
sarahetter: That won't work - if middlewareManifest does not exist, then it resolves to undefined. `undefined > 0` will error.

PR #1438, opened 2022-06-30
feat: use edge functions for content negotiation by default

### LINKED ISSUE #1374: Support nextConfig.images.formats
The next/image edge function returns AVIF images if the browser supports it. However it can be slower to process than other formats, so may time out (e.g. https://github.com/netlify/netlify-plugin-nextjs/issues/687#issuecomment-1113524399 ). In sites that need to use large images, a workaround could be to add support for [specifying supported image formats](https://nextjs.org/docs/api-reference/next/image#acceptable-formats), and not rewrite to AVIF format if it is disabled. The value could be read from the config and included in the edge function.

### LINKED ISSUE #1437: Enable image content negotiation by default
Currently next/image content negotiation is only enabled when a site opts-in to edge functions. This would be a good way of ramping up traffic to edge functions as we approach GA. It is a big user benefit, and the failure mode is the current behaviour, so it's low impact.

ascorbic:
### Summary

This PR enables content negotiation for images using edge functions to be opt-out, rather than opt-in. The PR ensures that the old redirect is also written, but it will not match if the edge function runs.

This PR also ensures that the edge function respects the Next image formats config option, which means that by default webp is the only modern format offered, but that AVIF is an option.

The PR adds some docs on the behaviour, as well as some general information about image handling. I rearranged some of the docs to adjust priorities.

@KyleBlankRollins This adds a new env var: `NEXT_DISABLE_EDGE_IMAGES`, which disables the edge fucntion used for content negotiation and instead always serves the fallback format (jpg or png)

### Test plan

1. Load the default demo deploy preview and check that the image page serves a webp image
2. Load the canary demo and check that an avif image is served

Fixes #1374. Fixes #1437 
capybara among flowers

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,42 +37,30 @@ If you build on Netlify, this plugin will work with no additional configuration.
 deploying locally using the Netlify CLI, you must deploy using `netlify deploy --build`. Running the build and deploy
 commands separately will not work, because the plugin will not generate the required configuration.
 
-## Migrating from an older version of the plugin
+## Using `next/image`
 
-You can manually upgrade from the previous version of the plugin by running the following command:
+If you use [`next/image`](https://nextjs.org/docs/basic-features/image-optimization), your images will be automatically
+optimized at runtime, ensuring that they are served at the best size and format. The image will be processed on the
+first request which means it may take longer to load, but the generated image is then cached at the edge and served as a
+static file to future visitors. By default, Next will deliver WebP images if the browser supports it. WebP is a new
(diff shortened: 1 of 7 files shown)


sarahetter (approved):
LGTM. On first load, the Canary demo showed a broken image, but on refresh it did load. Not sure if this is related to this PR, or just a fluke.

ascorbic:
Was probably AVIF timing-out. Let's see how it gets on in the wild. We may need to disable it, if it fails too much.

PR #1447, opened 2022-07-04
feat: dx updates to default demo site

orinokai:
### Summary

While testing the various rendering modes I made a few changes to the demo pages so that they were easier to test and it was more obvious how each page should be rendered. I thought it might be good to push these changes, if everyone agrees they are useful?

* Added descriptive text at the top of each page with a dynamic path that explains which paths are prerendered and what fallback options are enabled for other slugs
* Made the prerendered paths more consistent (always IDs 1 and 2)
* Added some package scripts for testing with `next build && next start`
* Added a simple VS Code launch.json for attaching the debugging to running inspector process

chill cat

### DIFF
--- a/demos/default/package.json
+++ b/demos/default/package.json
@@ -4,7 +4,9 @@
   "private": true,
   "description": "",
   "scripts": {
-    "build": "next build"
+    "build": "next build",
+    "dev": "next dev",
+    "serve": "next build && next start"
   },
   "repository": {
     "type": "git",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,15 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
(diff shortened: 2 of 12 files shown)


ascorbic (dismissed):
Much more coherent!

demos/default/package.json
```diff
@@ -4,7 +4,9 @@
   "private": true,
   "description": "",
   "scripts": {
-    "build": "next build"
+    "build": "next build",
+    "dev": "next dev",
+    "serve": "next build && next start"
```
ericapisani, line 9: [dust] would be great if we had a 'clean' command as well that removes all of the build artifacts as there's been times where some weird problems have resolved themselves after having those build artifacts removed prior to a build

ericapisani (approved):
These are great changes, thanks for doing this!

demos/default/package.json
orinokai: good point, though afaik there is no Next command for that? i have added a draft issue for it

MarcL (approved):
LGTM! 🎉

PR #1457, opened 2022-07-12
fix: explicitly type image config in ipx edge function

ascorbic:
### Summary

Explicitly type the imageconfig.json in the ipx content negotiation edge function. This fixes a bundling error in the case where the Next config is missing the formats key, meaning the inferred type was invalid.

baby capybara


MarcL (approved):
:shipit:

PR #1467, opened 2022-07-13
fix: add better error messaging for builds

### LINKED ISSUE #1362: Warn if build missing and directory is not `.next`
If a built site is not detected and user has set the publish directory to something other than `.next`, then we should check for a build in `.next` and if found then use a more specific error message, saying the user should change `publish` to `.next`. See  #1361 for an example where a message like that would have helped the user

ericapisani:
### Summary

Adds better error messaging for when:
* a user has a built site that's not detected
* the user has set the publish directory to something other than `.next`
* the netlify plugin detects a production build in the `.next` folder

### Test plan

* Go to demos/default test project and build it such that there's a production build in the `.next`
* Change the `publish` directory in `demos/default/netlify.toml` to something other than `.next`
* Attempt to build the project using `ntl build`

**Expected behaviour:** You should see an updated error message prompting you to set the 'publish' directory to `.next` due to detecting the presence of a production build in that directory

Fixed #1362

### DIFF
--- a/test/helpers/verification.spec.ts
+++ b/test/helpers/verification.spec.ts
@@ -12,7 +12,7 @@ jest.mock('fs', () => {
 
 describe('checkNextSiteHasBuilt', () => {
   let failBuildMock
-  let { existsSync } = require('fs')
+  const { existsSync } = require('fs')
   
   beforeEach(() => {
     failBuildMock = (jest.fn() as unknown) as FailBuild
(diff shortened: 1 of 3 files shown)


test/helpers/verification.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  
+  afterEach(() => {
+    jest.clearAllMocks()
+    jest.resetAllMocks()
+  })
+
+  it('returns error message about incorrectly publishing the ".next" directory when "next export" was run', () => {
+    existsSync.mockReturnValue(true)
+
+    const expectedFailureMessage = outdent`
```
sarahetter, line 29: Can we make this a const to avoid wording change errors?

sarahetter (approved):
Just one dust comment, LGTM

ascorbic (approved):
Nice! Much better

PR #1479, opened 2022-07-22
feat: add enhanced middleware support

ascorbic:
#### Let's give Next.js users some Edge Functions superpowers! 🚀 
This demo shows a way of adding some of the powers of Netlify Edge Functions to Next.js Middleware. Normal Next.js middleware doesn't provide actual access to the Response but rather allows the user to return a `NextResponse` object which is just used by the handler to set headers etc on the real response when it is returned. Calling `NextResponse.next()` doesn't actually send a request to the origin, but is just a placeholder for setting response headers which are applied later. It also does not allow changes to be made to the Request, but instead allows a `rewrite()` to be returned, which the wrapper uses to change the request. This library gives proper access to Request and Response in much the same way that Netlify Edge Functions do, but with some additional Next-specific helpers.

It currently adds these features:

- **Page data transforms**. Next passes `getXProps` data to a page component, and transfers it either as a JSON data file (for internal navigation with `next/link`) or embedded in the page HTML in a script tag for server-rendered pages. This feature allows users to modify those props on the fly. It works with both the data in HTML pages and data requests.
- **HTML rewrites** Allows rewriting of SSR HTML using the [HTMLRewriter](https://github.com/worker-tools/html-rewriter) stream transformer. Supports the full `HTMLRewriter` syntax.
- **Request headers** Next middleware allows response headers to be modified. This extends that to the request headers that are sent to the origin. Particularly helpful for rewrites to external sites, as for example it allows authentication headers to be added to the proxied request.
- **Access to response body** while the other features make it easier to modify reponses, you can also get full read/write access to the response body as a stream

### API
This is how I've implemented the API, but I'd welcome feedback. It's designed to be similar to the rest of the Next middleware API.

#### `MiddlewareRequest`
While `NextResponse` uses static factory methods to generate its responses, we use methods of a `MiddlewareRequest` object. This is because we need access to the edge function Request and Context objects, which we do by instantiating the class with the `NextRequest` object.
The following async methods return a `MiddlewareResponse` object.
- **`next()`**  passes the request on to the origin, and returns the response
- **`rewrite()`** rewrites the request URL and passes it on to the origin. Like `NextResponse` and unlike Netlify Edge Fucntions it supports external URLs, for which it uses `fetch()` directly.

#### `MiddlewareResponse`
This wraps the Netlify edge function `OriginResponse`. It is a subclass of `NextResponse`, so can be returned directly from the middleware. Unlike a `NextResponse`, once returned this includes the actual response from the origin. 

- **`originResponse`** the `OriginResponse` object
- **`setPageProp(key: string, value: unknown)`** sets the value of a single page prop
```typescript
    const request = new MiddlewareRequest(req)
    const response = await request.next()
    const message = `This was static but has been transformed in ${request.geo.city}`
    response.setPageProp('message', message)
```
- **`replaceText(selector: string, valueOrReplacer: string | ((input: string) => string))`** replaces the text content of an element. It can either take a string or a function that is passed the current value and returns the new value.
```typescript
    const request = new MiddlewareRequest(req)
    const response = await request.next()
    const message = `This was static but has been transformed in ${request.geo.city}`
    response.replaceText('#message', message)
```
- **`transformData(transform: (props: any) => any)`** add a transform function to modify the returned page data
```typescript
    const request = new MiddlewareRequest(req)
    const response = await request.next()
    const message = `This was static but has been transformed in ${request.geo.city}`

    // Transform the response page data
    response.transformData((data) => {
      data.pageProps.message = message
      data.pageProps.showAd = true
      return data
    })
```
- **`rewriteHTML(selector: string, handlers: ElementHandlers)`** allows the returned HTML to be written using [HTMLRewriter](https://github.com/worker-tools/html-rewriter)

```typescript
    const request = new MiddlewareRequest(req)
    const res = await request.next()
    const message = `This was static but has been transformed in ${request.geo.city}`
    // Transform the response HTML
    res.rewriteHTML('p[id=message]', {
      text(textChunk) {
        if (textChunk.lastInTextNode) {
          textChunk.replace(message)
        } else {
          textChunk.remove()
        }
      },
    })
```

**Caveats**
Modifying the reponse can cause React hydration errors if the values don't match. There are two ways around this. The best is if you are able to change both the text content and the prop. This means the value will still match. If this is not possible (if, for example, the displayed HTML is too complex to generate), then you can use two-pass rendering for the element. This is where you don't render it in SSR, but instead update the state in `useEffect` and then display it at that point. The logic for this is quite simple, and can be encapsulated in a custom hook to make it even simpler.
```typescript
import * as React from 'react'

const useHydrated = () => {
  const [hydrated, setHydrated] = React.useState(false)
  React.useEffect(() => {
    setHydrated(true)
  }, [])
  return hydrated
}

const Page = ({ message, showAd }) => {
  const hydrated = useHydrated()
  return (
    <div>
      {hydrated && showAd ? (
        <div>
          <p>This is an ad that isn't shown by default</p>
          <img src="http://placekitten.com/400/300" />
        </div>
      ) : (
        <p>No ads for me</p>
      )}
    </div>
  )
}

export async function getStaticProps() {
  return {
    props: {
      showAd: false,
    },
  }
}

export default Page

```

#### Modifying request headers

Headers on the `MiddlewareRequest` object can be added or modified, and the changed headers will then be passed along to the origin if you call `request.next()` or `request.rewrite()`. You can also modify headers on the original `NextRequest` object and they will be be passed along in the same way. This can be used for example to add authentication headers to an externally proxied request, or attach a bucket name header for A/B testing.

```typescript
const request = new MiddlewareRequest(req)
  if (pathname.startsWith('/api/hello')) {
    // Add a header to the request
    request.headers.set('x-hello', 'world')
    return request.next()
  }

  if (pathname.startsWith('/headers')) {
    // Add a header to the rewritten request
    request.headers.set('x-hello', 'world')
    return request.rewrite('/api/hello')
  }
```

### Test plan

Visit [the middleware deploy preview](https://deploy-preview-1479--next-plugin-edge-middleware.netlify.app/). For each example check both navigating from the homepage via next/link, and then reloading the page to check SSR
1. For "Adds `x-hello` request header" and "Adds `x-hello` request header to rewrite" ensure that the JSON includes the `x-hello` header
2. For "Rewrite static page content", ensure that the message shows a geolocated city, and that the "ad" kitten is displayed.

### Standard checks:

- [ ] Add docs when necessary

---


plugin/src/templates/edge/ipx.ts
```diff
@@ -1,14 +1,12 @@
-import { Accepts } from "https://deno.land/x/accepts@2.1.1/mod.ts";
-import type { Context } from "netlify:edge";
+import { Accepts } from 'https://deno.land/x/accepts@2.1.1/mod.ts'
```
ascorbic, line 1: These changes are because we've enabled prettier for Deno files because I can't work out how to make `deno fmt` run on a subsection of project files.

plugin/src/helpers/files.ts
```diff
@@ -331,13 +331,21 @@ const baseServerReplacements: Array<[string, string]> = [
```
ascorbic, line 331: This is fixing a bug that I found while building this. The change is included in a separate PR: #1480

plugin/src/templates/edge/utils.ts
```diff
@@ -34,6 +36,21 @@ export const addMiddlewareHeaders = async (
   })
   return response
 }
+
```
ascorbic, line 39: These are re-declared here, because the other declaration is Node code which depends on the `next` package. We just need the types here, and it's fine for our purposes to pretend that `NextResponse`s are actually just `Response`s.

plugin/src/templates/edge/utils.ts
```diff
@@ -43,13 +60,63 @@ export const buildResponse = async ({
   request: Request
   context: Context
 }) => {
+  // They've returned the NetlifyResponse directly, so we'll call `next()` for them.
+  if ('originalRequest' in result.response) {
+    result.response = await (result.response as unknown as NetlifyResponse).next()
+  }
```
ascorbic, line 66: Enforcement vs affordance 🤷🏻‍♂️

plugin/src/templates/edge/utils.ts
```diff
@@ -43,13 +60,63 @@ export const buildResponse = async ({
...
+  // This means it's a Netlify Next response.
+  if ('dataTransforms' in result.response) {
```
ascorbic, line 68: 🦆

plugin/src/templates/edge/utils.ts
ericapisani: [dust] Could we leave a comment about why it's fine for us to pretend that `NextResponse`s are just `Response`s in the code itself? 

I worry that there might be some confusion at a later point in time over why `Response` isn't `NextResponse`. Based on what you're saying I'm assuming it's because we don't need access to the other properties and methods that `NextResponse` has available at this time.

plugin/src/templates/edge/utils.ts
```diff
@@ -43,13 +60,63 @@ export const buildResponse = async ({
   request: Request
   context: Context
 }) => {
+  // They've returned the NetlifyMiddleware directly, so we'll call `next()` for them.
+  if ('originalRequest' in result.response) {
+    result.response = await (result.response as unknown as NetlifyMiddleware).next()
+  }
+  // This means it's a Netlify Next response.
+  if ('dataTransforms' in result.response) {
+    const response = result.response as NetlifyNextResponse
```
ericapisani, line 69: [sand] I think we can use a type guard here to replace the need for the `dataTransforms in result.response` check and casting of the `result.response` to NetlifyNextResponse in the following line with a single line. Something like

```
function isNetlifyNextResponse(response: NextResponse): response is NetlifyNextResponse {
  return 'dataTransforms' in response
}
```

and using it in the `if` should do the trick

ericapisani (commented):
Loving these changes so far 😍

plugin/src/templates/edge/ipx.ts
nickytonline: It looks like you can do the following in deno.json. See https://deno.land/manual/getting_started/configuration_file#configuration-file

```
{
  "fmt": {
    "files": {
      "include": [
        "src/templates/edge"
      ]
    }
  }
}
```

plugin/src/templates/edge/runtime.ts
```diff
@@ -38,13 +38,28 @@ const handler = async (req: Request, context: Context) => {
...
+
+  Object.defineProperty(geo, '__nf_context', {
+    value: context,
+    enumerable: false,
+  })
+
+  Object.defineProperty(geo, '__nf_request', {
+    value: req,
+    enumerable: false,
+  })
```
nickytonline, line 58: Smuggling it in on `geo` for the proof of concept works, but if we ship it, could we not add '__nf_context' as a header that can be deserialized when consumed or somewhere else?

plugin/src/templates/edge/utils.ts
nickytonline: 🦆

demos/middleware/next.config.js
```diff
@@ -6,6 +6,7 @@ const nextConfig = {
     // your project has ESLint errors.
     ignoreDuringBuilds: true,
   },
+  generateBuildId: () => 'build-id',
```
nickytonline, line 9: [dust] Consider leaving. a comment linking to https://nextjs.org/docs/api-reference/next.config.js/configuring-the-build-id

plugin/src/middleware/html-rewriter.ts
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable max-classes-per-file */
+
+// These types are inlined from the HTMLRewriter package, because we don't use the actual package here
```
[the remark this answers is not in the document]
nickytonline: Can we not import the types, e.g.

```typescript
import type { Element } from 'https://deno.land/x/html_rewriter@v0.1.0-pre.17'
```

Also, I'm assuming https://deno.land/x/html_rewriter@v0.1.0-pre.17 is the HTMLRewriter package you're talking about?

jlengstorf:
Since this isn't necessarily adding _Netlify_ features but instead enhancing the _core functionality of Next's middleware_, does it make sense for it be called `NetlifyMiddleware`?

This ostensibly could be implemented on any edge provider, right? I wonder if a more appropriate name would be something like `EnhancedNextMiddleware` or `NextMiddlewareComplete` or something about extending the feature set.

Not a blocker by any means. Just a thought.

biilmann:
Agree with @jlengstorf on the general naming idea. Ideally we would make this feel like a native next API that feels good enough that if other platforms want to support the same features they would use it as well.

plugin/src/templates/edge/runtime.ts
ascorbic: It's certainly a hack and is undoubtedly fragile. Using a header would be ideal, but it's not serialisable as a string. Context is a class and we need access to the instance methods. `geo` is the only object that is passed-through from outside the EF. The alternative I've been looking at is adding it to a global `Map` (or `WeakMap`), and then passing the key in as a header.

plugin/src/templates/edge/ipx.ts
ascorbic: Thanks! I'm not sure if that will work with VS Code format-on-save, but I'll give it a go.

plugin/src/middleware/html-rewriter.ts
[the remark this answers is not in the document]
ascorbic: This is Node code, so sadly not. It's compiled as part of the Next build.

ascorbic:
Good idea about making the name more "standard". I've gone with `EnhancedMiddleware` for now, though I'm open to other suggestions.

ascorbic:
OK, I've had a rethink about naming and changed it to `MiddlewareRequest` and `MiddlewareResponse`, which is generic and easier to understand. I think that `middlewareRequest.rewrite()` (or `redirect()` or `next()`) returning a `MiddlewareResponse` seems more intuitive.

plugin/src/middleware/request.ts
```diff
@@ -21,15 +21,16 @@ type AugmentedGeo = NextRequest['geo'] & {
...
   originalRequest: Request
 
-  constructor(public request: NextRequest) {
+  constructor(private nextRequest: NextRequest) {
+    super(nextRequest)
     if (!('Deno' in globalThis)) {
       throw new Error('NetlifyMiddleware only works in a Netlify Edge Function environment')
     }
-    const geo = request.geo as AugmentedGeo
+    const geo = nextRequest.geo as AugmentedGeo
```
nickytonline, line 33: [dust/sand] Consider extending `NextRequest` with `geo` typed as `AugmentedGeo` and have that be called something like `NetlifyRequest`, or `EdgeRequest`

plugin/src/middleware/request.ts
```diff
@@ -52,17 +53,35 @@ export class MiddlewareRequest {
...
+  get ip() {
+    return this.nextRequest.ip
+  }
+
+  get nextUrl() {
+    return this.nextRequest.url
+  }
+
+  get url() {
+    return this.nextRequest.url.toString()
```
nickytonline, line 83: [dust] Would there be a situation where we'd want an `URL` instead of a `string` for `url`? An `URL` would be more flexible than a `string` depending on what the consumer of the property would be doing. Or maybe there's a reason it's only exposed as a string?

ascorbic: This is a subclass of `Request`, so `url` will be a string.

plugin/src/templates/edge/runtime.ts
ascorbic: I've switched it to a global map, keyed by request ID. This is less likely to be broken by future changes in Next.js

plugin/src/templates/edge/ipx.ts
ascorbic: This didn't seem to work, because there's no way to tell VS Code to try both

ascorbic:
OK, I'm marking this ready for review. I suggest we put this out so people can try it. It's opt-in, so shouldn't affect anything. We may decide to publish it in a separate package, but right now it can be imported from here.

ericapisani (commented):
A couple of small things within the code itself, but I'm noticing there's no automated test coverage for the new functionality.

Is the plan to add some tests in a follow up?

plugin/src/middleware/request.ts
```diff
@@ -0,0 +1,80 @@
+import { NextURL } from 'next/dist/server/web/next-url'
+import { NextResponse } from 'next/server'
+import type { NextRequest } from 'next/server'
+
+import { MiddlewareResponse } from './response'
+
+// TODO: add Context type
```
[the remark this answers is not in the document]
ascorbic: No, I mean the real context type, if it gets published at some point

ascorbic:
So it turns out we don't have middleware e2e tests at all, so I've added some, including tests for the new stuff.

plugin/src/middleware/response.ts
```diff
@@ -0,0 +1,95 @@
+import { NextResponse } from 'next/server'
+
+import type { ElementHandlers } from './html-rewriter'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type NextDataTransform = <T extends { pageProps?: Record<string, any> }>(props: T) => T
```
nickytonline, line 6: [dust] Prefer `unknown`

```suggestion
export type NextDataTransform = <T extends { pageProps?: Record<string, unknown> }>(props: T) => T
```

plugin/src/middleware/response.ts
```diff
@@ -0,0 +1,95 @@
...
+            textChunk.remove()
+          }
+        },
+      })
+    }
+  }
+
+  get headers(): Headers {
+    // If we have the origin response, we should use its headers
+    return this.originResponse?.headers || super.headers
```
nickytonline, line 93: [dust] Prefer nullish coalescing.

```suggestion
    return this.originResponse?.headers ?? super.headers
```

nickytonline (approved):
This looks great @ascorbic! This is definitely in a great place where folks can start playing with this. 🚀

plugin/src/middleware/response.ts
ascorbic: I love nullish coalescing, but IMO it's only worth using if there's a legitimate reason for the value to be falsy. Otherwise it just introduces a (tiny) bit of extra transpiled code and execution time for no reason. This will either be a `Headers` object or `undefined`, so a simple `||` does the job.

demos/middleware/middleware.ts
```diff
@@ -1,11 +1,40 @@
...
 
-export function middleware(request: NextRequest, ev: NextFetchEvent) {
+import { MiddlewareRequest } from '@netlify/plugin-nextjs/middleware'
+
+export async function middleware(req: NextRequest) {
   let response
   const {
     nextUrl: { pathname },
-  } = request
+  } = req
```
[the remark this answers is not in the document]
ascorbic: `request` and `response` are already used for the `MiddlewareRequest` and `MiddlewareResponse`. I considered changing them to `nextRequest` and `nextResponse` but thought it made the code examples a bit verbose. Hard to say which is better tbh

plugin/src/middleware/response.ts
ascorbic: The reason I didn't do this is because it would mean that users would need to create a type and explicitly cast the props value whenever they used `transformData`, which could be quite cumbersome. I'll add a comment.

demos/middleware/middleware.ts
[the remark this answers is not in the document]
ascorbic: I realise I've mixed the different styles, so yeah this should be tidied up in a follow-up.

PR #1525, opened 2022-08-12
feat(graph): populate graph token for ISR

dwwoelfel:
### Summary

Allows the `getSecretsForBuild` and `getNetlifyGraphTokenForBuild` functionality to work with ISR in next.js.

Populates an environment variable that graph helpers exported from `@netlify/functions` will be able to read.

### Test plan

Here is an example of it in action: https://bucolic-torte-c9a71a.netlify.app/posts/15
Corresponding code: https://github.com/dwwoelfel/nextjs-blog-theme-9/blob/main/pages/posts/%5Bslug%5D.js#L17

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/packages/runtime/src/templates/getHandler.ts
+++ b/packages/runtime/src/templates/getHandler.ts
@@ -132,15 +132,11 @@ const makeHandler = (conf: NextConfig, app, pageRoot, staticManifest: Array<[str
     )
 
     const graphToken = event.netlifyGraphToken
-    if (graphToken) {
-      if (requestMode === 'ssr') {
-        multiValueHeaders['X-Netlify-Graph-Token'] = [graphToken]
-      } else {
-        // Prefix with underscore to help us determine the origin of the token
-        // allows us to write better error messages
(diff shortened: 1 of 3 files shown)


packages/runtime/src/templates/getHandler.ts
```diff
@@ -130,6 +130,15 @@ const makeHandler = (conf: NextConfig, app, pageRoot, staticManifest: Array<[str
     console.log(
       `[${event.httpMethod}] ${event.path} (${requestMode?.toUpperCase()}${result.ttl > 0 ? ` ${result.ttl}s` : ''})`,
     )
+
+    const graphToken = event.netlifyGraphToken
```
ascorbic, line 134: At this point the request will already have been processed, so this will only set it for the _next_ request

dwwoelfel: Thanks!

ascorbic (approved):
:shipit:

PR #1536, opened 2022-08-15
feat: skip run if @netlify/plugin-nextjs installed

ericapisani:
Disable in order to avoid conflicts should the old package be installed in the project.

### Summary

Disables the runtime if it detects that the old plugin is installed in the project. Show a message prompting user to remove the package from their project so they can use the runtime instead.

### Test plan
- [x] Confirm message shows in project that has `@netlify/plugin-js` installed (test done in demos/default)
- [x] Confirm site without `@netlify/plugin-js` works as expected

[Issue](https://github.com/netlify/pod-ecosystem-frameworks/issues/185)

baby-tiger

---

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,4 +1,4 @@
 {
-  "packages/runtime": "4.15.0",
+  "packages/runtime": "4.16.0",
   "packages/next": "0.0.1"
 }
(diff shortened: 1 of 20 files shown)


sarahetter (approved):
LGTM!

PR #1540, opened 2022-08-16
feat: ✨  change over to next-runtime

ericapisani:
### Summary

Change over to `next-runtime`.

There's still some references to `plugin-nextjs` in some of the package locks for the demo sties, I think those will need to be fixed once the initial package of `next-runtime` is released (since the error right now for those relate to not being able to find the package in the npm registry)

[Github issue](https://github.com/netlify/pod-ecosystem-frameworks/issues/187)
let's do this


ascorbic:
If you run npm install in the monorepo, it _should_ update the lockfiles without it needing to be published. It can see the package is in the workspace.

ericapisani:
@ascorbic I had done that but for some odd reason some are working and others aren't. As an example - the default demo installs correctly while the `next-auth` one fails (even after removing the `package-lock.json` file within `next-auth` as well) 🤔 Perhaps I just need to dig a little more into this

Screen Shot 2022-08-16 at 12 41 18 PM

ericapisani:
@ascorbic I found the issue - we forgot to add the `next-auth` project to the `workspaces` list when we transitioned to it a while back 😅

ascorbic:
Well spotted!

sarahetter (approved):
LGTM

PR #1541, opened 2022-08-16
fix: support non-prerendered dynamic routes with fallback false

### LINKED ISSUE #1179: [Bug]: Fallback false is ignored
### Summary

Netlify triggers a new ODB function instead of returning the expected 404 when using `fallback: false`.

See: https://github.com/netlify/netlify-plugin-nextjs/blob/6fd7bcc99aacf447559de46f60de6d8cb33e7a59/src/helpers/utils.ts#L80-L88

**Context**:
We have a top level dynamic route, so that we can create new landing pages directly from our CMS. Because `fallback: false` was ignored, every request to `myWebsite.com/:uid` resulted in a new ODB function being triggered. However, these requests were for example also triggered by crawlers requesting `robots.txt` (which was missing at that time). Another edge case leading to a huge number of function invocations was the usage of the nextjs `Link` component for a route with a configured redirect (in _redirects). Because Nextjs prefetches routes, it tried to prefetch the JSON of the route (which didn't exist because it's a redirect). This would not have been a problem (only a failed request in the network tab of the browser, and it can be fixed by disabling the prefetching)...However, with Netlify ignoring `fallback: false`, a new ODB function was spawned for every prefetch request (which could be many per user session). 

All in all we accumulated 170k+ function invocations in one weekend. I know that the above described scenario is very specific (and could have been avoided) but I wanted to emphasize, that the i18n fix can result in some unpleasent surprises for Netlify customers that use `fallback: false` and expect it to work as described in the nextjs docs.

### Steps to reproduce

.

_No response_

### Plugin version

4.2.1

### What OS are you using?

_No response_

### Your netlify.toml file

<details>
  
```toml
```

</details>

<details>
  
```toml
```

</details>

<details>
  
```toml
```

</details>

<details>
  <summary>Build logs</summary>
  
```
# Paste logs here
```

</details>

### Function logs

<details>
  
```
# Paste logs here
```

</details>

### .next JSON files

<details>
  
```
```

</details>

orinokai:
### Summary

Next.js uses an internal cache to persist page creation and updates during incremental static regeneration (ISR).

In order to preserve atomic deploys, Netlify has a different model that uses our ODB functions for creating and updating pages after deployment.

To support this, we currently patch the Next.js server to disable the internal cache, which ensures our ODB functions always receive fresh content and can handle the caching and revalidation directly.

However, our current solution has an incidental bug that bypasses the `fallback: false` and `fallback: true` functionality and always serves the page like `fallback: blocking`. In addition, it creates an unnecessary function invocation and runs `getStaticProps` unexpectedly.

Fixing this bug enables support for all the fallback options, but doesn’t work in one specific case: pre-rendered ISR pages with `fallback: false` and a revalidation TTL. These pages get served as a 404 because there is no fallback and the server thinks there is no statically pre-rendered page. This does not happen for ISR pages without revalidation because they are served statically, bypassing the origin entirely.

This PR implements a different strategy: we continue to disable cache writes, but enable cache reads and force manual revalidation mode so fresh content is always served.

Further work is currently ongoing in order to support bypassing the ODB handler for 404 pages.

- [x] enable cache reads to support `fallback: false` and `fallback: true`, but force manual revalidation for fresh content
- [x] add tests

### Test plan

See test coverage

Fennec fox stack

Fixes #1179 

---


hu0p:
> adding documentation to warn i18n users who don't prerender all their locales

Is the motivation for this due to [Matt's comment](https://github.com/netlify/next-runtime/issues/1179#issuecomment-1042804719) or is this a nice to have? I'm not sure this is necessary (unless I'm missing something) as it appears it's the intended behavior according to [the docs](https://nextjs.org/docs/advanced-features/i18n-routing#dynamic-routes-and-getstaticprops-pages) (excerpt below).

> For pages using getStaticProps with [Dynamic Routes](https://nextjs.org/docs/routing/dynamic-routes), all locale variants of the page desired to be prerendered need to be returned from [getStaticPaths](https://nextjs.org/docs/basic-features/data-fetching/get-static-paths). Along with the params object returned for paths, you can also return a locale field specifying which locale you want to render. For example:

packages/runtime/src/helpers/redirects.ts
```diff
@@ -23,6 +24,14 @@ import {
...
   middleware.some((middlewarePath) => route.startsWith(middlewarePath))
 
+const generateHiddenPathRedirects = ({ basePath }: Pick<NextConfig, 'basePath'>): NetlifyConfig['redirects'] =>
+  HIDDEN_PATHS.map((path) => ({
+    from: `${basePath}${path}`,
+    to: '/404.html',
+    status: 404,
+    force: true,
+  }))
+
```
orinokai, line 34: minor refactor here wasn't strictly necessary, but I was testing some things and needed to move it out

packages/runtime/src/helpers/files.ts
```diff
@@ -333,8 +333,13 @@ const getServerFile = (root: string, includeBase = true) => {
...
 }
 
+// force manual revalidation so the ODB handler always receives fresh content
+// ensure fallback behaviour is bypassed for pre-rendered pages
+// ensure ISR 404 pages send the correct SWR cache headers
 const baseServerReplacements: Array<[string, string]> = [
-  [`let ssgCacheKey = `, `let ssgCacheKey = process.env._BYPASS_SSG || `],
+  [`checkIsManualRevalidate(req, this.renderOpts.previewProps)`, `checkIsManualRevalidate({ headers: null }, null)`],
+  [`isManualRevalidate && (fallbackMode !== false || hadCache)`, `isManualRevalidate && hadCache`],
+  [`private: isPreviewMode || is404Page && cachedData`, `private: isPreviewMode && cachedData`],
```
orinokai, line 342: These are the new patches for forcing manual revalidation. It's also possible to force this by sending an `x-prerender-revalidate` request header in the handler, but that involves pulling in the `prerender-manifest.json` to get a token and I thought this might actually be more succinct

packages/runtime/src/helpers/redirects.ts
```diff
@@ -191,13 +248,14 @@ const generateDynamicRewrites = ({
...
-    if (isApiRoute(route.page)) {
+    if (isApiRoute(route.page) || is404Route(route.page, i18n)) {
       return
     }
     if (route.page in prerenderedDynamicRoutes) {
+      const { fallback } = prerenderedDynamicRoutes[route.page]
       if (matchesMiddleware(middleware, route.page)) {
         dynamicRoutesThatMatchMiddleware.push(route.page)
-      } else {
+      } else if (fallback !== false) {
```
orinokai, line 258: this is the important bit as it allows non-prerendered dynamic routes with `fallback: false` to fall through and hit the 404 redirects

packages/runtime/src/helpers/redirects.ts
```diff
@@ -231,14 +289,7 @@ export const generateRedirects = async ({
...
   )
 
-  netlifyConfig.redirects.push(
-    ...HIDDEN_PATHS.map((path) => ({
-      from: `${basePath}${path}`,
-      to: '/404.html',
-      status: 404,
-      force: true,
-    })),
-  )
```
orinokai, line 241: refactoring the hidden paths function as mentioned above

packages/runtime/src/helpers/redirects.ts
```diff
@@ -294,12 +345,7 @@ export const generateRedirects = async ({
   netlifyConfig.redirects.push(...dynamicRewrites)
   routesThatMatchMiddleware.push(...dynamicRoutesThatMatchMiddleware)
 
-  // Final fallback
-  netlifyConfig.redirects.push({
-    from: `${basePath}/*`,
-    to: HANDLER_FUNCTION_PATH,
-    status: 200,
-  })
```
orinokai, line 302: removing the catch all redirect in lieu of the new custom 404 redirects

ascorbic (commented):
I don't totally understand what's happening with the SSG cache. What happens when it tries to write to disk? Does it handle the failure? Is this being done so that it can return the prebuilt placeholder HTML?

hu0p:
> Thanks @hu0p, that last item on the checklist is just because the current behaviour will render all locales regardless of whether a user specifies them in `getStaticPaths`. The new behaviour will 404 for unspecified locales, which is the correct behaviour, as you pointed out, but probably needs to be called out in the release notes.

Got it. Thanks for the clarification! Watching this as much as I can. I appreciate both of your work on all of this. Thank you! I've been eying that highlighted portion of the redirects file off and on for months.

packages/runtime/src/helpers/files.ts
```diff
@@ -334,7 +334,13 @@ const getServerFile = (root: string, includeBase = true) => {
 }
 
 const baseServerReplacements: Array<[string, string]> = [
-  [`let ssgCacheKey = `, `let ssgCacheKey = process.env._BYPASS_SSG || `],
+  // force manual revalidate during cache fetches
+  [
+    `checkIsManualRevalidate(req, this.renderOpts.previewProps)`,
+    `checkIsManualRevalidate(process.env._REVALIDATE_SSG ? { headers: { 'x-prerender-revalidate': this.renderOpts.previewProps.previewModeId } } : req, this.renderOpts.previewProps)`,
```
ericapisani, line 340: Do we know if `process.env._REVALIDATE_SSG` could ever be 'false' (which would result in an unintentional truthy evaluation)?

I'm also not certain about if `previewProps` are required within `renderOpts`, but just in case - is there a chance that it could be `undefined` such that an error would be thrown when `previewModeId` is accessed?

test/index.js
```diff
@@ -1030,12 +1030,14 @@ describe('utility functions', () => {
...
     const patchedData = await readFileSync(serverFile, 'utf8')
-    expect(patchedData.includes('_BYPASS_SSG')).toBeTruthy()
+    expect(patchedData.includes('_REVALIDATE_SSG')).toBeTruthy()
+    expect(patchedData.includes('private: isPreviewMode && cachedData')).toBeTruthy()
 
     await unpatchNextFiles(process.cwd())
 
     const unPatchedData = await readFileSync(serverFile, 'utf8')
-    expect(unPatchedData.includes('_BYPASS_SSG')).toBeFalsy()
+    expect(unPatchedData.includes('_REVALIDATE_SSG')).toBeFalsy()
```
ericapisani, line 1039: Related to my comment above - while I know `_REVALIDATE_SSG` is being explicitly set to `'true'`, the fact that this statement is passing makes me a little wary about unexpected behaviour if  `_REVALIDATE_SSG` is set to `'false'` since `toBeFalsy` checks if the value is any of false, 0, '', null, undefined, and NaN, and everything else is true (from the [Jest docs](https://jestjs.io/docs/expect#tobefalsy))

packages/runtime/src/helpers/files.ts
orinokai: Nope, `_REVALIDATE_SSG` will never be false when deployed on Netlify. It will be undefined if the code runs outside of our handler functions, e.g. locally with `next dev` (unless a user explicitly set the env var, but it's not publicly documented).

Thanks for the heads-up, but I've double-checked the Next server code and `renderOpts` will always contain `previewProps`, which will always contain `previewModeId`.

ericapisani (approved):

PR #1546, opened 2022-08-17
feat: add edge middleware support to `ntl dev`

ascorbic:
### Summary
Adds support for running middleware using edge functions with `ntl dev`. 

When `next dev` is run, it doesn't emit any of the compiled files that we need to run middleware via our normal edge function. This means that currently we need to run `ntl build` before `ntl dev`, and changes would require a recompiled. This is not good DX: we want to be able to run `ntl dev` without building or linking the site, and have it update middleware when it is changed.

In this PR this is implemented with a special dev-only edge function which runs the middleware directly, rather than via the Next server. While importing the middleware.ts file directly works in simple cases, it breaks if the middleware imports anything else, because the imports won't be Deno-compatible. We handle this by running `esbuild --watch` on the middleware, and importing the compiled middleware. It still works if the middleware hasn't been created yet because esbuild will compile it when it is.

The wrapper edge function checks at request time whether the compiled middleware exists. If not, it just silently returns as there is no middleware. If it does exist then it dynamically imports it, then executes it. This means that it will always have the latest version without needing to reload the edge function.

The edge function is generated in a new `onDev` hook, which also spawns `esbuild`. A regular build will delete the dev edge function to ensure it's never deployed, and the dev hook deletes the production edge functions to ensure they don't conflict.

This needs e2e tests, which will follow later: #1553

### Test plan

1. Check out repo
2. Run npm i in root
3. Install latest version of CLI once [11.0.0](https://github.com/netlify/cli/pull/4970) is released
4. Edit `demos/middleware/netlify.toml` to switch plugin package names
5. Run `npm i` and `ntl dev` in middleware demo
6. Load site and check the "rewrite static" demo. You should see a cat and your city.

capy lifeguard

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/demos/middleware/middleware.ts
+++ b/demos/middleware/middleware.ts
@@ -5,16 +5,13 @@ import { MiddlewareRequest } from '@netlify/next'
 
 export async function middleware(req: NextRequest) {
   let response
-  const {
-    nextUrl: { pathname },
-  } = req
+  const { pathname } = req.nextUrl
 
   const request = new MiddlewareRequest(req)
-
   if (pathname.startsWith('/static')) {
     // Unlike NextResponse.next(), this actually sends the request to the origin
     const res = await request.next()
-    const message = `This was static but has been transformed in ${req.geo.city}`
+    const message = `This was static but has been transformed in ${req.geo?.city}`
 
     // Transform the response HTML and props
     res.replaceText('p[id=message]', message)
@@ -56,6 +53,16 @@ export async function middleware(req: NextRequest) {
       url.pathname = '/shows/100'
       response = NextResponse.rewrite(url)
       response.headers.set('x-modified-in-rewrite', 'true')
+    }
+
+    if (pathname.startsWith('/shows/redirectme')) {
+      const url = req.nextUrl.clone()
+      url.pathname = '/shows/100'
+      response = NextResponse.redirect(url)
+    }
+
+    if (pathname.startsWith('/shows/redirectexternal')) {
+      response = NextResponse.redirect('http://example.com/')
     }
 
     if (!response) {
--- a/demos/middleware/netlify.toml
+++ b/demos/middleware/netlify.toml
@@ -4,7 +4,9 @@ publish = ".next"
 ignore = "if [ $CACHED_COMMIT_REF == $COMMIT_REF ]; then (exit 1); else git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF ../..; fi;"
 
 [[plugins]]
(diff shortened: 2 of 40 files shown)


demos/middleware/middleware.ts
```diff
@@ -58,6 +55,16 @@ export async function middleware(req: NextRequest) {
       response.headers.set('x-modified-in-rewrite', 'true')
     }
 
+    if (pathname.startsWith('/shows/redirectme')) {
```
ascorbic, line 58: Redirects weren't working, so this is a test for the fix

ascorbic, line 58: Redirects were broken, so these are the added tests

demos/middleware/netlify.toml
```diff
@@ -4,7 +4,7 @@ publish = ".next"
 ignore = "if [ $CACHED_COMMIT_REF == $COMMIT_REF ]; then (exit 1); else git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF ../..; fi;"
 
 [[plugins]]
-package = "../plugin-wrapper/"
+package = "@netlify/plugin-nextjs"
```
ascorbic, line 7: `ntl dev` won't de-dupe if the wrapper is used

packages/runtime/src/index.ts
```diff
@@ -218,5 +219,29 @@ const plugin: NetlifyPlugin = {
     warnForRootRedirects({ appDir })
   },
 }
-module.exports = plugin
+// The types haven't been updated yet
+const nextRuntime = (
+  _inputs,
+  meta: { events?: Set<string> } = {},
+): NetlifyPlugin & { onPreDev?: NetlifyPlugin['onPreBuild'] } => {
+  if (!meta?.events?.has('onPreDev')) {
```
eduardoboucas, line 227: Great touch!

packages/runtime/src/helpers/dev.ts
```diff
@@ -0,0 +1,44 @@
+import { resolve } from 'path'
+
+import { OnPreBuild } from '@netlify/build'
+import execa from 'execa'
+import { unlink, existsSync } from 'fs-extra'
+
+import { writeDevEdgeFunction } from './edge'
+import { patchNextFiles } from './files'
+
+// The types haven't been updated yet
```
eduardoboucas, line 10: Ah, good point.

eduardoboucas (approved):
Didn't go deep into the Next.js parts, but LGTM!

PR #1548, opened 2022-08-17
chore: revert skipping runtime, revert package name to plugin-nextjs

sarahetter:
### Summary

Reverts https://github.com/netlify/next-runtime/pull/1536, and changes the npm package name to `@netlify/plugin-nextjs` in the package's package.json


ascorbic (approved):
Farewell next-runtime. We hardly knew ye

ascorbic:
I think the auth demo is failing because of duplicate versions of React. The error is the classic "invalid hook call" that usually means that.

PR #1551, opened 2022-08-18
chore: fix broken next-auth demo

ascorbic:
### Summary

Update demo from React 17, which broke builds. Also changes tests to run with Node LTS (16) because Node 18 was causing engines issues.

### Test plan

1. Check the next-auth demo preview builds

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/demos/next-auth/pages/_app.tsx
+++ b/demos/next-auth/pages/_app.tsx
@@ -1,10 +1,9 @@
 import { SessionProvider } from "next-auth/react"
-import type { AppProps } from "next/app"
 import "./styles.css"
 
 // Use of the <SessionProvider> is mandatory to allow components that call
 // `useSession()` anywhere in your application to access the `session` object.
-export default function App({ Component, pageProps }: AppProps) {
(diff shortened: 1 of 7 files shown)


demos/next-auth/pages/_app.tsx
```diff
@@ -1,10 +1,9 @@
 import { SessionProvider } from "next-auth/react"
-import type { AppProps } from "next/app"
 import "./styles.css"
 
 // Use of the <SessionProvider> is mandatory to allow components that call
 // `useSession()` anywhere in your application to access the `session` object.
-export default function App({ Component, pageProps }: AppProps) {
```
estephinson, line 7: Big sads the props are gone, any reason?

estephinson (approved):
LGTM

demos/next-auth/pages/_app.tsx
ascorbic: Something weird with Next types that are not important to fix today!

PR #1558, opened 2022-08-18
fix: ensure newly-created middleware works

ascorbic:
### Summary

Currently when running `ntl dev`, if there isn't any middleware when you start the server and then create one, even though the file is correctly compiled, the middleware isn't imported until a restart because Deno has cached the "fact" that it's missing. It correctly updates any changes if the middleware already exists at the time the server starts.

This PR adds a cache-busting hash to the import URL, which ensures it try to load dynamically on every request.

Published as `@netlify/plugin-nextjs@4.17.1-runtime.4`

### Test plan

1. Install `@netlify/plugin-nextjs@4.17.1-runtime.4` in a new Next.js project
2. start `ntl dev`
3. Add a `middleware.ts`, such as the one below
4. Reload the page and check it works

```typescript
import { NextResponse } from 'next/server'

export async function middleware(req) {
    const response = NextResponse.next()
    response.headers.set('x-hello-deno', 'Deno' in globalThis ? 'true' : 'false')
    return response
}
```

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/packages/runtime/src/helpers/dev.ts
+++ b/packages/runtime/src/helpers/dev.ts
@@ -31,6 +31,7 @@ export const onPreDev: OnPreBuild = async ({ constants, netlifyConfig }) => {
     `--bundle`,
     `--outdir=${resolve('.netlify')}`,
     `--format=esm`,
+    `--target=esnext`,
     '--watch',
     // Watch for both, because it can have either ts or js
     resolve(base, 'middleware.ts'),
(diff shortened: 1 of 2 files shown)


sarahetter:
On refresh, I get a page that says "middleware is not a function"

sarahetter (approved):
Yayy!

PR #1569, opened 2022-08-24
chore: allow overriding the check for the bundle size

minivan:
### Summary

Skipping the checks for the bundle size are helpful when folks know what kind of bundles they are uploading to Netlify

image

---

### DIFF
--- a/packages/runtime/src/helpers/verification.ts
+++ b/packages/runtime/src/helpers/verification.ts
@@ -105,6 +105,8 @@ export const checkForRootPublish = ({
 }
 
 export const checkZipSize = async (file: string, maxSize: number = LAMBDA_MAX_SIZE): Promise<void> => {
+  // Experimental: We can skip checking for the bundle size on the plugin side.
+  // Please reach out to the Netlify support team if you'd like to enable this feature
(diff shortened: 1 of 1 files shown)


packages/runtime/src/helpers/verification.ts
```diff
@@ -105,6 +105,13 @@ export const checkForRootPublish = ({
 }
 
 export const checkZipSize = async (file: string, maxSize: number = LAMBDA_MAX_SIZE): Promise<void> => {
+  // Experimental: We can skip checking for the bundle size on the plugin side.
+  // Please reach out to the Netlify support team if you'd like to enable this feature
+  if (!process.env.NETLIFY_CHECK_BUNDLE_SIZE) {
```
eduardoboucas, line 110: I'd thought this would be the other way around, where you'd set an environment variable if you want to bypass the size check. I don't see anyone going out of their way to enable this variable, so this would effectively amount to disabling the check?

orinokai:
thanks for the PR @minivan, much appreciated, but i think this has been superseded by https://github.com/netlify/next-runtime/pull/1570

PR #1575, opened 2022-08-26
feat: improve messaging on middleware detection

ericapisani:
### Summary

Improve middleware messaging when middleware is not detected in a site. At the moment it appears like something is wrong if a site does not have middleware due to the middleware path not being able to be resolved in the `execa` calls.

### Test plan

Tested running `ntl dev` when the middleware file is not there, is in JS, or is in TS.

bluejay

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,4 +1,4 @@
 {
-  "packages/runtime": "4.18.1",
+  "packages/runtime": "4.19.0",
   "packages/next": "1.1.0"
 }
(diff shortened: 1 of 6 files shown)


orinokai (commented):
Unless I'm mistaken, if middleware isn't detected then we don't need to run esbuild at all?

ericapisani:
@orinokai You're correct, good catch. I've moved the `execa` calls to within the `else` block so we also can get rid of the flag I initially introduced

orinokai (approved):
Woo! LGTM 🚀

PR #1584, opened 2022-08-31
fix: remove Next.js as a peer dependency

ericapisani:
### Summary

Removes Next.js as a peer dependency within the runtime package as at times it was leading to multiple version of Next.js being installed, and incompatible versions being bundled with the functions when the project was using an older version.

### Test plan

Deploy preview: https://ep-next-test.netlify.app/

The https://ep-next-test.netlify.app/api/hello route is an SSR route

Test case provided in issue: https://ep-hbo-clone.netlify.app/movie (this should render correctly, no error)

Fixes https://github.com/netlify/pod-ecosystem-frameworks/issues/220

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5185,9 +5185,9 @@
       "devOptional": true
     },
     "node_modules/@types/react": {
-      "version": "17.0.48",
-      "resolved": "https://registry.npmjs.org/@types/react/-/react-17.0.48.tgz",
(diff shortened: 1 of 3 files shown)


nickytonline (approved):
This makes sense based on the issue @danez reported and folks will always have Next installed since the runtime is for Next. 😎

ascorbic:
> folks will always have Next installed since the runtime is for Next. 😎

This is why I thought it would be safe to have it as a peer dep. `:old man yells at npm 7 peer dependencies:` I've also had build errors importing it when it's not listed as a peer dep. Presumably this is working ok, but it is worth looking out for.

danez:
In that case adding it as dev dependency would work. As our ci only installs production dependencies.

PR #1586, opened 2022-08-31
feat: create edge function demo

ericapisani:
### Summary

Demo site that uses an edge function

### Test plan

Deploy preview: https://630fc8bcddda2d2489526966--ep-next-edge-demo-site.netlify.app/

Rather than render the 'Welcome to Next.js' page, this will return a JSON response from the edge function included in the site

PXL_20220831_170918383


MarcL (dismissed):
[pebble] We don't need most of the HTML in `index.js` and the icon and Vercel image, given we render the JSON instead, so we could simplify and remove if you want to. Depends on how simple we want to make it versus what Next installs by default.

sarahetter (dismissed):

sarahetter (approved):

PR #1588, opened 2022-09-02
fix: clean up edge functions during build

ascorbic:
### Summary

Ensure edge functions directory is emptied during a build. This fixes the problem where a user with no middleware runs `ntl dev` and then builds. Previously the `next-dev` edge function was only cleaned up if other edge fucntions were being generated, and if none were then it wasn't cleaned up. This meant that the dev edge function could be deployed to production. This PR ensures that it is always emptied, even if edge functions aren't being generated.

capybara family

### Standard checks:

- [ ] Add docs when necessary

---


.github/workflows/codeql-analysis.yml
```diff
@@ -9,14 +9,14 @@
 # the `language` matrix defined below to confirm you have the correct set of
```
ascorbic, line 9: This had to be included because it would fail linting otherwise

orinokai (dismissed):
Tested and all good :shipit:

orinokai (approved):

PR #1592, opened 2022-09-05
fix: watch for creation of middleware

ascorbic:
### Summary

We need to start esbuild watching for middleware files even if they don't exist, so that if the user creates them then they're immediately available. However this gives a scary warning about the file being missing (sometimes... see later). This PR adds a filter so that this particular message isn't shown.

HOWEVER. Right now, once the dev command has started, nothign output from the `onPreDev` command is actually shown. Until this is fixed, most of this output will disappear into the void.

### Test plan

1. Run `ntl dev` in the demo repo with no middleware (e.g. `demos/custom-routes`). 
2. WIth `ntl dev` still running, create a middleware.ts file and see that it immediately creates `.netlify/middleware.js`

capy drinking maté

### Standard checks:

- [ ] Add docs when necessary

---


.github/workflows/codeql-analysis.yml
```diff
@@ -9,14 +9,14 @@
 # the `language` matrix defined below to confirm you have the correct set of
```
ascorbic, line 9: Ignore this linting stuff

packages/runtime/src/helpers/dev.ts
```diff
@@ -20,21 +23,31 @@ export const onPreDev: OnPreBuild = async ({ constants, netlifyConfig }) => {
   })
   await writeDevEdgeFunction(constants)
 
-  if (!existsSync(resolve(base, 'middleware.ts')) && !existsSync(resolve(base, 'middleware.js'))) {
-    console.log(
-      "No middleware found. If you did intend to use middleware, create a 'middleware.ts' or 'middleware.js' file in your project root to add custom middleware.",
```
ascorbic, line 25: This was really just for my benefit when I was developing it, and I meant to remove it.

orinokai (approved):
Tested and all looks good 🚢

PR #1597, opened 2022-09-06
fix: handle stricter Next type

ascorbic:
### Summary
A small change to Next types broke our builds in canary. Changing from an interface to type fixes it.


sarahetter (approved):
Thanks!

Issue #1605, opened 2022-09-08
Prevent creation of IPX function


gualterandre:
Would it be possible to prevent the creation of the `ipx` function if one does not need it?

Currently, next-runtime is always creating this function but there could be some edge cases where this isn't necessary

<img width="600" alt="image" src="https://user-images.githubusercontent.com/35470944/189076772-e6d33c2f-3421-40bc-a67c-d557bf035177.png">

h3ku:
> More asking in order to help us triage this - is the presence of this in the cases where it isn't necessary leading to an undesirable outcome (functionally, monetarily, etc.) for some users?

Having it enabled by default means that if a customer has a domain where anyone can upload files configured in `next.config.js` (Common behaviour since this usually contain CDN domains, like cloudflare or contentfull), a malicious user could upload a js/html file to the CDN and then make the following request to the victim site.
`http://domain/_ipx/w_32,q_32/https://images.ctfassets.net/id` 

Which would lead to an xss on domain.com

shaungcheng-asana:
Hi @h3ku, I see you created a pull request to add an environment variable to disable `_ipx`: https://github.com/netlify/next-runtime/pull/1609 but the PR is closed and not merged. Do we still have plans to expose a env var perhaps as a workaround, etc?

h3ku:
> Hi @h3ku, I see you created a pull request to add an environment variable to disable `_ipx`: #1609 but the PR is closed and not merged. Do we still have plans to expose a env var perhaps as a workaround, etc?

Hey! Im not a maintainer of the package, I opened the pull request but it was not ideal since it would break some implementations.

In the end I end up using [patch-package](https://www.npmjs.com/package/patch-package) to disable the registration of the IPX function as a temporary solution while Netlify decides to give users the option to disable the function via an env var.
https://gist.github.com/h3ku/1a82585929fb33553bb29d430cecd771

PR #1609, opened 2022-09-08
Add the option to disable IPX function

h3ku:
### Summary

Add the option to disable IPX function using DISABLE_IPX env var.

To be honest I would disable IPX by default and give the option to enable it via an env var (Having this enabled by default put Netlify customer at risk of XSS), but this PR does the opposite since it would be a breaking change.


h3ku:
https://github.com/netlify/next-runtime/issues/1605

PR #1631, opened 2022-09-22
fix: correctly enable edge images

ascorbic:
### Summary

When we changed the logic so that EFs are used by default for middleware, we accidentally chnaged back the logic so that image content negotiation was no longer done with them by default. This PR fixes that, and EFs are used for image content negotiation unless disabled.

Fixes https://github.com/netlify/next-runtime/issues/1648

### Test plan

1. Visit the `/image` page in the default demo and ensure that the generated images are _not_ redirected to `/_ipx`

emo capy

### Standard checks:

- [ ] Add docs when necessary

---


ericapisani (commented):
These changes generally look good to me. Is it possible to get some of this under some sort of test coverage as well?

packages/runtime/src/helpers/edge.ts
```diff
@@ -200,7 +200,18 @@ export const writeEdgeFunctions = async (netlifyConfig: NetlifyConfig) => {
...
+  const nextConfigFile = await getRequiredServerFiles(publish)
+  const nextConfig = nextConfigFile.config
+  await copy(getEdgeTemplatePath('../edge-shared'), join(edgeFunctionRoot, 'edge-shared'))
+  await writeJSON(join(edgeFunctionRoot, 'edge-shared', 'nextConfig.json'), nextConfig)
+
+  if (
+    process.env.NEXT_DISABLE_EDGE_IMAGES !== '1' &&
+    process.env.NEXT_DISABLE_EDGE_IMAGES !== 'true' &&
+    process.env.NEXT_DISABLE_NETLIFY_EDGE !== 'true' &&
+    process.env.NEXT_DISABLE_NETLIFY_EDGE !== '1'
```
ericapisani, line 213: [sand] It's a shame I wasn't able to get my IPX work in before these changes because I'm adding the `isEnvSet` method that we also have in the [Gatsby plugin](https://github.com/netlify/next-runtime/blob/276b948c288e3b592e4db6a96c206dd61c3d9b51/packages/runtime/src/helpers/utils.ts#L232) to help make logic like this a little easier on the eyes.

You're free to include the method in your PR if you'd like to use here and I can resolve the conflicts in my branch, but no worries if you're a bit pressed for time and would rather leave this as is for the time being.

ericapisani (approved):

packages/runtime/src/index.ts
```diff
@@ -88,30 +83,30 @@ const plugin: NetlifyPlugin = {
...
+    ) {
+      failBuild(outdent`
+        You are using Next.js experimental edge runtime, but have set NEXT_DISABLE_NETLIFY_EDGE to true. This is not supported.
+        To use edge runtime, remove the env var ${bold`NEXT_DISABLE_NETLIFY_EDGE`}.
+      `)
     }
 
-    if (middlewareManifest?.middleware && Object.keys(middlewareManifest.middleware).length !== 0) {
-      usingEdge = true
-      if (process.env.NEXT_DISABLE_NETLIFY_EDGE === 'true' || process.env.NEXT_DISABLE_NETLIFY_EDGE === '1') {
```
ascorbic, line 105: Just realised I accidentally removed the logic in this

packages/runtime/src/helpers/edge.ts
ascorbic: @ericapisani I've updated it to use [destr](https://github.com/unjs/destr)

ericapisani (approved):

packages/runtime/src/helpers/edge.ts
ascorbic: Yeah, but bear in mind we can't use it inside the patched code

PR #1634, opened 2022-09-22
fix: add missing data to middleware request object

ericapisani:
### Summary

When the middleware runs on the edge, it's missing data that is hydrated by the NextServer when run in the context of the origin server.

Also fixes which property is accessed on nextURL in the context of the MiddlewareRequest.

### Test plan

[Test site](https://app.netlify.com/sites/ep-next-enhanced-middleware-test)

Can also do the following:
* pull this branch
* Update the `next.config.js` of the middleware demo to include the following:
  
```
  i18n: {
    locales: ["en-US"],
    defaultLocale: "en-US",
  }
```

* Add a console.log below [this line](https://github.com/netlify/next-runtime/blob/5cf8c52ead9a9013906201697a6a4a32fd8368f7/demos/middleware/middleware.ts#L8) referencing the `locale` and `defaultLocale`
* Confirm that the values are `en-US`

**Note:** It's a known issue that the middleware example isn't working entirely as expected and are not related to these changes. For instance, the path to set the `netlifyCookie` via the 'Cookies API' route isn't setting the cookie properly, although the user will be routed to the correct page. Issue for this has been created [here](https://github.com/netlify/next-runtime/issues/1639)

Fixes https://github.com/netlify/pod-ecosystem-frameworks/issues/190

### Standard checks:

- [x] Add docs when necessary

---


packages/runtime/src/templates/edge/runtime.ts
```diff
@@ -81,10 +82,11 @@ const handler = async (req: Request, context: Context) => {
   const request: RequestData = {
     headers: Object.fromEntries(req.headers.entries()),
     geo,
-    url: url.toString(),
+    url,
```
ericapisani, line 85: This changes addresses the missing `pathname` property on `req.nextURL`

packages/next/src/middleware/request.ts
```diff
@@ -83,7 +83,7 @@ export class MiddlewareRequest extends Request {
   }
 
   get nextUrl() {
-    return this.nextRequest.url
```
MarcL, line 86: Should we look to add some unit tests around the code here? I'm wondering if we'd have caught this bug earlier if we had some.

ericapisani: Yeah I'll take a look at doing so, but going to move this into 'ready for review' to start getting some 👀 on this

README.md
```diff
@@ -48,6 +48,12 @@ Next.js Middleware works out of the box on Netlify. By default, middleware runs
 support for running Middleware at the origin, set the environment variable `NEXT_DISABLE_NETLIFY_EDGE` to `true`. Be
 aware that this will result in slower performance, as all pages that match middleware must use SSR.
 
+### Limitations
+
+Due to how the site configuration is handled when it's run using Netlify Edge Functions, data such as `locale` and `defaultLocale` will be missing on the `req.nextUrl` object when running `netlify dev`.
```
ascorbic, line 55: When you say "such as", does that mean there are other things missing?

ericapisani: I haven't yet done an audit of all the properties that are affected by this so I'm not 100% certain that these are the only properties affected on the `nextUrl` object.

Something to add here as well but the pathname (as reported by the user who originally raised this issue with us) was also incorrect before these changes and should be mentioned.

jestSetup.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line n/no-unpublished-require
+require('jest-fetch-mock').enableMocks()
```
ericapisani, line 2: Without this package and setup, the `Request` and `Headers` objects aren't found in the context of the test file that was added (`request.spec.ts`)

ascorbic (approved):

package.json
```diff
@@ -18,7 +18,7 @@
     "postinstall": "run-s build install-husky",
     "install-husky": "if-env CI=1 || husky install node_modules/@netlify/eslint-config-node/.husky",
     "test": "run-s build:demo test:jest",
-    "test:jest": "jest",
+    "test:jest": "jest --testMatch=**/test/request.spec.ts",
```
nickytonline: Did this get updated by accident while debugging tests?

```suggestion
    "test:jest": "jest",
```

ericapisani: Yes it did, good catch!

PR #1647, opened 2022-09-30
fix: set undefined ODB 404 TTLs to 60 seconds

orinokai:
### Summary

This PR fixes a bug that causes 404 responses to be cached indefinitely by ODB functions, meaning that once an ODB route returns a 404 it can never be invalidated by a subsequent 200 response.

The bug occurred because Next Server sends standard cache control headers (SWR, max-age one year) for a 404 response, even if the URL is an ISR route which would otherwise (for a 200 response) send headers with a max age equal to the revalidate value. This means that our handler function was not adding a TTL to the builder function when an ISR page returned a 404.

Standard 404 pages on non-ISR routes will be routed to the standard non-ODB handler function and will therefore bypass the logic in this PR, which should apply to ISR routes that conditionally return a `notFound: true` in `getStaticProps`.

### Test plan

TODO: create an ISR test page that flip-flops between 200 and 404 based on minutes past the build timestamp

TODO: link to all the relevant issues!


ascorbic (approved):
Tried it and it works great!

PR #1653, opened 2022-10-03
feat: Add ability to disable ipx

### LINKED ISSUE #1605: Prevent creation of IPX function

ericapisani:
### Summary

Allows for the ability to disable `ipx` via the `DISABLE_IPX` environment variable.

If this is enabled, this will require users to a loader (such as Cloudinary), otherwise the images that would've previously been served by `ipx` will return a 404.

### Test plan

Deploy preview - https://ep-disable-ipx.netlify.app

* Visit 'Rewrite (should display image)' link
* Images should not render and return 404s within the network tab

Fixes #1605 

Related documentation PR: https://github.com/netlify/docs/pull/2414

PXL_20221001_220245959

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,9 @@ by targeting the `/_next/image/*` route:
 
 ## Disabling `ipx`
 
-If you wish to disable the use of the `ipx` package, set the `DISABLE_IPX` environment variable to `true`.
+If you wish to disable the use of the `ipx` package, set the `DISABLE_IPX` environment variable to `true`. Please note that some requests to `/_next/image/*` may fail unless an image loader, such as Cloudinary or Imgix, is specified as a replacement for `ipx`. 
+
+See the [Next.js documentation](https://nextjs.org/docs/api-reference/next/image#built-in-loaders) for options.
 
 ## Next.js Middleware on Netlify
 
--- a/README.md
+++ b/README.md
@@ -42,6 +42,10 @@ by targeting the `/_next/image/*` route:
     X-Test = 'foobar'
 ```
 
+## Disabling `ipx`
+
+If you wish to disable the use of the `ipx` package, set the `DISABLE_IPX` environment variable to `true`.
+
 ## Next.js Middleware on Netlify
 
 Next.js Middleware works out of the box on Netlify. By default, middleware runs using Netlify Edge Functions. For legacy
--- a/README.md
+++ b/README.md
@@ -42,11 +42,13 @@ by targeting the `/_next/image/*` route:
     X-Test = 'foobar'
 ```
 
-## Disabling `ipx`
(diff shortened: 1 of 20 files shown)


packages/runtime/src/helpers/utils.ts
```diff
@@ -228,4 +228,7 @@ export const getRemotePatterns = (experimental: ExperimentalConfigWithLegacy, im
   }
   return []
 }
+
+export const isEnvSet = (envVar: string) => process.env[envVar] === 'true' || process.env[envVar] === '1'
```
ascorbic, line 232: We really should be consistent about using this or destr. I'm not particularly bothered either way but it doesn't make sense to have both.

README.md
```diff
@@ -42,6 +42,12 @@ by targeting the `/_next/image/*` route:
     X-Test = 'foobar'
 ```
 
+## Disabling `ipx`
```
ascorbic, line 45: [pebble] I don't like the way this is worded. I don't think ipx is the thing to talk about here, because there's no reason anyone should care or know about what ipx is. The point is that it's disabling the image loader, and that they should only do it if they are either not using `next/image`, or are using a different loader.

ericapisani: I've updated the README, let me know if that's a bit better or if you would like to see further changes

packages/runtime/src/helpers/functions.ts
```diff
@@ -69,35 +70,50 @@ export const setupImageFunction = async ({
...
-    ...imageconfig,
-    basePath: [basePath, IMAGE_FUNCTION_NAME].join('/'),
-    remotePatterns,
-    responseHeaders,
-  })
-  await copyFile(join(__dirname, '..', '..', 'lib', 'templates', 'ipx.js'), join(functionDirectory, functionName))
+  if (destr(process.env.DISABLE_IPX)) {
+    // If no image loader is specified, need to redirect to a 404 page since there's no
+    // backing loader to serve local site images once deployed to Netlify
+    if (!IS_LOCAL && imageconfig.loader && imageconfig.loader === 'default') {
```
ascorbic, line 78: [sand] does this need the check for `imageconfig.loader`?

ascorbic (dismissed):

packages/runtime/src/helpers/functions.ts
ericapisani: No, I had written that out of habit when a property is optional, but this could be achieved with the `imageconfig.loader === 'default'` check

sarahetter (approved):

Issue #1475, opened 2022-07-20
Upgrade minimum Node version to 14 in netlify-plugin-next


ericapisani:
Currently the "engines" property in the plugin's `package.json` is set to node v12.

Given node 12 was EOLed recently, this should be upgraded to 14.

ascorbic:
For frameworks plugins we've been matching the minimum versions in the frameworks themselves. Last time I checked, Next.js still supports 12, but hopefully that will change soon. Gatsby moved to 14 last year, which is why we can require that.

charkour:
With Next v13, the minimum node version is v14: [Docs](https://nextjs.org/docs/upgrading#v13-summary)

taty2010:
Will need to hold off on updating since Next v12 still has node 12 as the minimum.

serhalp:
Superseded by v5.

PR #1659, opened 2022-10-04
chore: remove support for node 12

### LINKED ISSUE #1646: chore: Drop Node 12 support in @netlify/next-runtime
https://github.com/netlify/next-runtime/blob/main/packages/runtime/package.json#L67

ericapisani:
Fixes #1646

### DIFF
--- a/packages/next/package.json
+++ b/packages/next/package.json
@@ -32,6 +32,6 @@
   },
   "homepage": "https://github.com/netlify/next-runtime#readme",
   "engines": {
-    "node": ">=12.0.0"
+    "node": ">=14.0.0"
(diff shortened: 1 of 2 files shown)


ascorbic:
This should be marked as a breaking change, so the major version is incremented

ericapisani:
We're going to hold off on merging this change for now since this will require a major version bump of the runtime

LekoArts:
Closing for now as it's not actionable at the moment. We can re-open once it's decided in https://github.com/netlify/next-runtime/issues/1475 that Next.js has upgraded their Node.js requirements

PR #1676, opened 2022-10-11
feat: add support for WebAssembly in edge runtime

ascorbic:
### Summary

Adds support for wasm in the edge runtime. When Next.js builds a site that uses wasm in the edge runtime, it emits the wasm files into the `.next/server/edge-chunks` directory. The runtime expects these chunks to be available as arraybuffers as variables, named according to the chunk name. We do this by inlining them as base64-encoded strings, then wrapping those in `decode()` calls (using the Deno base64 lib).

Similarly, static assets such as fonts are replaced with `blob:` URLs, which use the chunk name for the path. We handle this in the same way as the Next edge runtime: by wrapping `fetch` and intercepting any calls that request a blob URI, returning the base64-encoded file in a Response instead.

This PR also adds the OG image lib to the default demo, and uses it for an e2e test.

### Test plan

1. Visit the [the OG demo](https://deploy-preview-1676--netlify-plugin-nextjs-demo.netlify.app/api/og?username=ascorbic) in the deploy preview

image

### Standard checks:

- [ ] Add docs when necessary

---


demos/default/next.config.js
```diff
@@ -8,6 +8,9 @@ module.exports = {
     defaultLocale: 'en',
     locales: ['en', 'es', 'fr'],
   },
+  eslint: {
+    ignoreDuringBuilds: true,
```
nickytonline, line 12: Can we fix the lint errors instead of ignoring them during build?

packages/runtime/src/helpers/edge.ts
```diff
@@ -98,6 +124,22 @@ const getMiddlewareBundle = async ({
 }): Promise<string> => {
   const { publish } = netlifyConfig.build
   const chunks: Array<string> = [preamble]
+
+  if ('wasm' in edgeFunctionDefinition) {
+    for (const { name, filePath } of edgeFunctionDefinition.wasm) {
+      const wasm = await fs.readFile(join(publish, filePath))
```
nickytonline, line 130: Can we use the file system in Edge functions now? I thought for some reason we didn't have access to `Deno.fs*`?

ascorbic: This is run during the build, so it uses Node

nickytonline: Doh! I could've sworn I saw a Deno.readFile. Late in the day PR reviews. 🫠

test/index.js
```diff
@@ -617,6 +617,14 @@ describe('onBuild()', () => {
   test('does not generate an ipx edge function if Netlify Edge is disabled', async () => {
     process.env.NEXT_DISABLE_NETLIFY_EDGE = '1'
     await moveNextDist()
+
+    // We need to pretend there's no edge API routes, because otherwise it'll fail
+    // when we try to disable edge runtime.
+    const manifest = path.join('.next', 'server', 'middleware-manifest.json')
+    const manifestContent = await readJson(manifest)
+    manifestContent.functions = {}
+    await writeJSON(manifest, manifestContent)
```
nickytonline, line 626: This doesn't block the PR, but could we not have a separate test for the scenario when `NEXT_DISABLE_NETLIFY_EDGE=1` and we have routes, that we expect it to throw? Or maybe I'm misunderstanding the error being explained here.

demos/default/pages/api/og.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const config = {
+  runtime: 'experimental-edge',
+}
+
+export default async function handler(req: NextRequest) {
+  const { searchParams } = req.nextUrl
+  const username = searchParams.get('username')
+
+  return new ImageResponse(
```
nickytonline, line 12: For context where this component is coming from, see the Vercel example, https://vercel.com/docs/concepts/functions/edge-functions/og-image-generation#getting-started

nickytonline (approved):
Tested and we got the OG image happening. WASM! (but make it somehow sound like Shazam!) 🚀 

CleanShot 2022-10-17 at 18 48 28

nickytonline (approved):
Thanks for this @ascorbic!

nickytonline (approved):
🚀

PR #1677, opened 2022-10-11
feat: cache redirects for 60 seconds if no ttl provided

orinokai:
### Summary

Next.js uses 307 and 308 codes for its redirects, but our ODB functions currently only invalidate cached content for 200, 301 and 404 codes. This means ISR pages that conditionally return a redirect are not updating as expected.

A fix to Netlify Proxy for the above issue is currently in progress and this PR is a corresponding fix for a specific Next.js case: ISR pages with redirects, but without a `revalidate` value, are returned to the ODB with a default `max-age` of 1 year, which causes the ODB to cache them indefinitely. This PR ensures that redirects without `revalidate` are cached ephemerally for 60s.

### Test plan

~Visit /getStaticProps/with-revalidate-redirect/ on the demo site deploy preview and observe that the page should redirect if the last digit of the current minute is 0-4 and should show a successful 200 response if the last digit is 5-9.~
Unable to create a test page in the default demo because redirect cannot be returned when pre-rendering. Tests to follow in a new demo site.

### Relevant links (GitHub issues, Notion docs, etc.)

netlify/pillar-runtime#443
netlify/proxy#950

### DIFF
--- a/packages/runtime/src/templates/getHandler.ts
+++ b/packages/runtime/src/templates/getHandler.ts
@@ -131,8 +131,9 @@ const makeHandler = (conf: NextConfig, app, pageRoot, staticManifest: Array<[str
           // ODBs currently have a minimum TTL of 60 seconds
           result.ttl = Math.max(ttl, 60)
         }
-        // Only cache 404s ephemerally
-        if (ttl === ONE_YEAR_IN_SECONDS && result.statusCode === 404) {
+        const ephemeralCodes = [301, 302, 307, 308, 404]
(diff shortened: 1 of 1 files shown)


nickytonline:
@orinokai, can we add an end to end test for this fix or is it not possible atm as we'd need to wait at least 60 seconds for it to invalidate?

orinokai:
@nickytonline the main problem with adding a test is that the site won’t build when there is a redirect in `getStaticProps` so we’d need a way to ensure it only returns a redirect after the initial build. I have a plan but it involves creating a new demo site with a timestamp as the build ID and then setting up some long running Cypress tests. Since that will take some time and this is an urgent fix, I’d planned to do that as a separate bit of work.

PR #1680, opened 2022-10-12
fix: use process.env.URL when deploying to production

ericapisani:
### Summary

* Upgrades the version to `next-auth` to the latest
* Updates the logic for creating the default `NEXTAUTH_URL`. 
  * In a production context, the unique hash that gets pre-pended to the URL leads to issues because `next-auth` uses that URL as the basis for updating it's provider information. For instance - the callback URL used for OAuth applications (since it expects a stable value). This change makes the URL stable by using the ['URL' rather than the `DEPLOY_PRIME_URL`](https://docs.netlify.com/configure-builds/environment-variables/#deploy-urls-and-metadata) when it detects that the [`CONTEXT`](https://docs.netlify.com/configure-builds/environment-variables/#build-metadata) is set to 'production'

### Test plan

1. Visit the Deploy Preview - https://deploy-preview-1680--netlify-plugin-nextjs-next-auth-demo.netlify.app/
2. Visit the '/admin' and '/me' pages.

* Visiting those pages should not result in a server error
* If you inspect the 'sign in' button shown as part of the default demo when visiting `/admin` or `/me`, the URL that it goes to should be the URL of the site (e.g: `https://deploy-preview-1680--netlify-plugin-nextjs-next-auth-demo.netlify.app/api/callback....`)

1. Visiting this test site https://upgrade-next-auth--ep-next-auth.netlify.app/
2. Visit the `/admin` and `/me` pages

* Visiting those pages should not result in a server error
* If you inspect the 'sign in button' on `/admin` or `/me`, the URL that it goes to should start with `https://ep-next-auth.netlify.app/` not `https://upgrade-next-auth--ep-next-auth.netlify.app`
* If you sign in (there's a test OAuth app backing that site), you should be able to visit the `/me` endpoint, but the `/admin` page should still be inaccessible since `userRole: admin` is not set on the returned token that makes that page accessible.


test/index.js
```diff
@@ -253,6 +255,53 @@ describe('onBuild()', () => {
...
+
+    await nextRuntime.onBuild(defaultArgs)
+
+    expect(onBuildHasRun(netlifyConfig)).toBe(true)
+    const config = await getRequiredServerFiles(netlifyConfig.build.publish)
+
+    expect(config.config.env.NEXTAUTH_URL).toEqual(mockUserDefinedSiteUrl)
+  })
+  
+  test('sets the NEXTAUTH_URL to the URL when CONTEXT env variable is \'production\'', async () => {
```
nickytonline, line 281: [dust] Be more specific which URL.

```suggestion
  test('sets the NEXTAUTH_URL to the user defined site URL when CONTEXT env variable is \'production\'', async () => {
```

packages/runtime/src/index.ts
```diff
@@ -136,7 +136,12 @@ const plugin: NetlifyPlugin = {
 
         await updateRequiredServerFiles(publish, config)
       } else {
-        const nextAuthUrl = `${process.env.DEPLOY_PRIME_URL}${basePath}`
+        // Using the deploy prime url in production leads to issues because the unique deploy ID is part of the generated URL
+        // and will not match the expected URL in the callback URL of an OAuth application.
+        const nextAuthUrl =
+          process.env.CONTEXT === 'production'
+            ? `${process.env.URL}${basePath}`
+            : `${process.env.DEPLOY_PRIME_URL}${basePath}`
```
nickytonline, line 144: [dust] Tiny refactor suggestion.

```suggestion
        const nextAuthUrl = `${
          process.env.CONTEXT === 'production' ? process.env.URL : process.env.DEPLOY_PRIME_URL
        }${basePath}`
```

nickytonline (approved):
Looks great @ericapisani! Left some dusty comments 🚀

nickytonline (approved):
Thanks for making the changes! 🚀

nickytonline (approved):
🚀

PR #1682, opened 2022-10-14
fix: prevent Next from defining duplicate global property in edge functions

ascorbic:
### Summary

Currently Next.js creates a global function in the edge runtime called `__import_unsupported` by defining it as an immutable property on `globalThis`. This means that if there is more than one of these definitions in a single edge function bundle then there is an error when they both try to define it. This PR patches the generated bundle (yeah, I know) to add a check before defining it. I suppose I _could_ try contributing upstream.

The only scenario where this happens now is if you have more than one Next.js edge function or middleware. This means that you need to define an API route to use `experimental-edge` runtime, and also have middleware.

### Test plan

1. Visit the middleware deploy preview
2. Load the edge API page `/api/edge`

image

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/packages/runtime/src/helpers/edge.ts
+++ b/packages/runtime/src/helpers/edge.ts
@@ -74,7 +74,7 @@ const sanitizeName = (name: string) => `next_${name.replace(/\W/g, '_')}`
 const preamble = /* js */ `
 
 globalThis.process = { env: {...Deno.env.toObject(), NEXT_RUNTIME: 'edge', 'NEXT_PRIVATE_MINIMAL_MODE': '1' } }
-const _ENTRIES = {}
+let _ENTRIES = {}
 // Deno defines "window", but naughty libraries think this means it's a browser
 delete globalThis.window
 // Next uses "self" as a function-scoped global-like object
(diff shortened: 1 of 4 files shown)


packages/runtime/src/helpers/edge.ts
```diff
@@ -90,17 +97,20 @@ const getMiddlewareBundle = async ({
...
+    // Next defines an immutable global variable, which is fine unless you have more than one in the bundle
+    // This adds a check to see if the global is already defined
+    data = IMPORT_UNSUPPORTED.reduce(
+      (acc, val) => acc.replace(val, `('__import_unsupported' in globalThis)||${val}`),
+      data,
+    )
     chunks.push('{', data, '}')
   }
 
-  const middleware = await fs.readFile(join(publish, `server`, `${edgeFunctionDefinition.name}.js`), 'utf8')
```
ascorbic, line 100: This isn't needed anymore, because the middleware file itself is included in `edgeFunctionDefinition.files`

packages/runtime/src/helpers/edge.ts
```diff
@@ -90,17 +97,20 @@ const getMiddlewareBundle = async ({
...
-
-  chunks.push(middleware)
-
```
ascorbic, line 103: This isn't needed anymore, because the middleware file itself is included in `edgeFunctionDefinition.files`

ericapisani (approved):

PR #1693, opened 2022-10-18
fix: correctly handle matchers with lookaheads and i18n

ascorbic:
### Summary

Currently we translate the Next.js middleware matcher regex into a regex used to target the edge function. We double-check when the edge funciton runs, because there are addiitonal conditions that can be set in matchers that can't be expressed in the path regex. However the Next.js and Netlify Edge Function regexes are not fully compatible, because for performance reasons the Go regexp library (used in the Netlify proxy) doesn't support lookaheads like JavaScript does, and the Next.js docs now suggest using these. 

This PR parses the matcher regexes, and strips any lookahead clauses before writing them to the Netlify manifest. Loosening them in this way doesn't prevent them from working, because it retains the lookaheads in the matchers used in the edge functions themselves, so the condition is still enfored at that point. 

Additionally, if a site uses i18n, the generated regex was not matching the default locale. This is because Next.js checks the path after it has been normalized and the default locale interpolated. However Netlify runs the match before it it normalised. This PR loosens that regex by making the locale slug optional, meaning it matches the default locale with no slug included. 

### Test plan

1. The middleware test site now includes i18n, ensuring the tests pass with it enabled. It also includes a matcher with a negative lookahead. Load [this page](https://deploy-preview-1693--next-plugin-edge-middleware.netlify.app/shows/99), which is excluded with a negative lookahead and ensure the `x-modified-edge` header is not set in the response. Load [this page](https://deploy-preview-1693--next-plugin-edge-middleware.netlify.app/shows/11) and see that it _is_ set.

### Standard checks:

- [ ] Add docs when necessary

---


test/matchers.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    expect(checkPath('/shows', stripped)).toBe(false)
+    expect(checkPath('/shows/11', stripped)).toBe(true)
+    // These will be true because the regex is not as strict as the original one
+    // The strict test will be done in the JS entrypoint
+    expect(checkPath('/shows/99', stripped)).toBe(true)
+    expect(checkPath('/shows/888', stripped)).toBe(true)
+  })
+})
+
+// "regexp": "^(?:\\/(_next\\/data\\/[^/]{1,}))?(?:\\/([^/.]{1,}))\\/shows(?:\\/((?!99|88).*))(.json)?[\\/#\\?]?$"
```
sarahetter, line 72: Was this comment meant to be left in?

ascorbic: Ah, no. Fixed.

sarahetter (approved):
LGTM!

PR #1714, opened 2022-10-25
feat: support Next 13

ascorbic:
### Summary

Updates the test sites and dependencies to Next 13 (actually 12.3.2-canary.43, but that is the same as next 13, which isn;t yet in npm)

### Test plan

1. Check the deploy previews and e2e tests!

image

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/demos/default/pages/deep/import.js
+++ b/demos/default/pages/deep/import.js
@@ -20,9 +20,7 @@ const Show = ({ show }) => (
 
     <hr />
 
-    <Link href="/">
-      <a>Go back home</a>
-    </Link>
+    <Link href="/">Go back home</Link>
   </div>
 )
 
--- a/packages/runtime/src/templates/edge/runtime.ts
+++ b/packages/runtime/src/templates/edge/runtime.ts
@@ -33,7 +33,7 @@ export interface RequestData {
     name?: string
     params?: { [key: string]: string }
   }
-  url: URL
+  url: string
   body?: ReadableStream<Uint8Array>
 }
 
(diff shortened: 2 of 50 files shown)


packages/runtime/src/templates/edge/runtime.ts
```diff
@@ -82,7 +82,7 @@ const handler = async (req: Request, context: Context) => {
   const request: RequestData = {
     headers: Object.fromEntries(req.headers.entries()),
     geo,
-    url,
+    url: req.url,
```
nickytonline, line 85: Was this broken before or were we doing `request.toString()` currently where this is being called?

ascorbic: This is passed to Next.js. It broke things here: https://github.com/netlify/next-runtime/actions/runs/3321501353/jobs/5489233564#step:10:348

ascorbic:
@ericapisani the next/image tests are failing at the moment. The images seem to display fine, but the tests fail. I'm going to switch them to use next/image/legacy and we can look into why the new component is breaking thing.

demos/default/pages/deep/import.js
```diff
@@ -20,9 +20,7 @@ const Show = ({ show }) => (
 
     <hr />
 
-    <Link href="/">
-      <a>Go back home</a>
-    </Link>
+    <Link href="/">Go back home</Link>
```
nickytonline, line 23: For context, you no longer need to do this.

> Note: Before Next.js 12.2, it was required that the Link component wrapped an `<a>` tag, but [this is not required in versions 12.2 and above](https://nextjs.org/blog/next-12-2).

See https://nextjs.org/learn/basics/navigate-between-pages/link-component

nickytonline (approved):
The changes look good. Assuming all checks pass, this is good to 🚀

ascorbic:
The i18next demo will fail to build for now (until Next 13 is published to npm), but aside from that it should all be good to go.

PR #1730, opened 2022-10-28
fix: add cookies route to middleware config

### LINKED ISSUE #1639: [Bug] Next runtime middleware demo not working correctly
While working on https://github.com/netlify/next-runtime/pull/1634 I noticed that the path for the 'Cookies API' in `demos/middleware/middleware.ts` wasn't being reached where the `netlifyCookie` is set, and this behaviour is happening on the `main` branch. 

This seems to happen for most of the routes listed within the middleware file (the first conditional - `/static` is working correctly however).

Determine why the behaviour within the middleware demo isn't working as expected, where the user is able to visit the routes (correctly) but some of the side effects of the middleware are not being run (like the setting of the cookie).

sarahetter:
### Summary

The middleware demo isn't setting a cookie on the `/cookies` route currently because we forgot to add it to the config. 

Fixes #1639

### Test plan

1. Visit the middleware Deploy Preview 
2. Click the "Cookies API" link, or visit `/cookies` 
3. See that the "netlifyCookie" has been set.

PXL_20210524_182610476 MP
### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/cypress/integration/middleware/standard.spec.ts
+++ b/cypress/integration/middleware/standard.spec.ts
@@ -19,6 +19,12 @@ describe('Standard middleware', () => {
       expect(response.headers).to.have.property('x-modified-edge', 'true')
     })
   })
+
+  it('adds cookies', () => {
+    cy.request('/cookies').then(() => {
+      cy.getCookie('netlifyCookie').should('have.property', 'value', 'true')
+    })
+  })
 })
 
 describe('Middleware matchers', () => {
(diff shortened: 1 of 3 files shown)


ascorbic:
What a handsome hound.

Do we have a test that would have caught this?

sarahetter:
@ascorbic Good point! Added a cypress test to check for the cookie.

ascorbic (approved):
🚢 🍪

PR #1745, opened 2022-11-07
fix: exclude swc by default, and don't exclude sharp if included

ascorbic:
### Summary

Currently we exclude several large node modules from being bundled with functions, to avoid going over the 50MB limit. Since Next 13, functions are trying to bundle swc binaries, which are massive. This PR excludes them. However, some users _do_ want to bundle sharp, for example if they're using it in API call. This PR means it no longer excludes sharp if the user has manually added it to `included_files`

### Test plan

1. Run a local build of the demo site.
2. Run `zipinfo .netlify/functions/___netlify-odb-handler.zip | sort -nr -k 4 | head`. Ensure that no files matching `@next/swc*` are included.

capy in severance

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -162,6 +162,7 @@ beforeEach(async () => {
   netlifyConfig.headers = []
   netlifyConfig.functions[HANDLER_FUNCTION_NAME] && (netlifyConfig.functions[HANDLER_FUNCTION_NAME].included_files = [])
   netlifyConfig.functions[ODB_FUNCTION_NAME] && (netlifyConfig.functions[ODB_FUNCTION_NAME].included_files = [])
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -517,6 +518,7 @@ describe('onBuild()', () => {
       '.next/BUILD_ID',
       '.next/static/chunks/webpack-middleware*.js',
       '!.next/server/**/*.js.nft.json',
+      '!**/node_modules/@next/swc*/**/*',
```
nickytonline, line 521: [dust] consider a separate PR for not bundling swc.

nickytonline (approved):
Thanks @ascorbic! Letf one suggestion. 🚀

PR #1756, opened 2022-11-10
fix: publish from subdirectory

ascorbic:
### Summary

Currently we expect Next.js to publish the `distDir` (i.e. `.next`). This is how we ascertain how to find the build files, particularly `required-server-files.json`. However most of the generated outputs are not created in the correct structure, so we need to copy or move them there as part of the build. This has worked fine so far, but has some important drawbacks. The main one is that it means the whole folder is uploaded as part of the deploy, even though most of it is not needed - and in fact needs 404 redirects to avoid making them publicly available. This hasn't been a major problem in the past (once the first deploy has been uploaded there aren't many extra files that need uploading). However Next 13 has started generating very large cache files that change on every build, which are now being uploaded every time.

Since this plugin was created it is now possible to change the publish dir from within a build plugin. This gives us another option that retains the best of these worlds. We can use `.next` as the default publish dir, but during the build change it to a hidden `.next/dist` dir, which then contains the actual files.

### Test plan

1. Ensure that the deploy previews work

image

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/packages/runtime/src/helpers/files.ts
+++ b/packages/runtime/src/helpers/files.ts
@@ -1,7 +1,6 @@
 /* eslint-disable max-lines */
 import { cpus } from 'os'
 
-import type { NetlifyConfig } from '@netlify/build'
 import { yellowBright } from 'chalk'
 import { existsSync, readJson, move, copy, writeJson, readFile, writeFile, ensureDir, readFileSync } from 'fs-extra'
 import globby from 'globby'
@@ -60,11 +59,11 @@ export const matchesRewrite = (file: string, rewrites: Rewrites): boolean => {
   return matchesRedirect(file, rewrites.beforeFiles)
 }
 
-export const getMiddleware = async (publish: string): Promise<Array<string>> => {
+export const getMiddleware = async (distDir: string): Promise<Array<string>> => {
   if (process.env.NEXT_DISABLE_NETLIFY_EDGE !== 'true' && process.env.NEXT_DISABLE_NETLIFY_EDGE !== '1') {
     return []
   }
-  const manifestPath = join(publish, 'server', 'middleware-manifest.json')
+  const manifestPath = join(distDir, 'server', 'middleware-manifest.json')
   if (existsSync(manifestPath)) {
     const manifest = await readJson(manifestPath, { throws: false })
     return manifest?.sortedMiddleware ?? []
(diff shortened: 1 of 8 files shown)


packages/runtime/src/helpers/redirects.ts
```diff
@@ -25,14 +25,6 @@ import {
 const matchesMiddleware = (middleware: Array<string>, route: string): boolean =>
   middleware.some((middlewarePath) => route.startsWith(middlewarePath))
 
-const generateHiddenPathRedirects = ({ basePath }: Pick<NextConfig, 'basePath'>): NetlifyConfig['redirects'] =>
-  HIDDEN_PATHS.map((path) => ({
```
ascorbic, line 29: Not needed anymore, as these aren't published

packages/runtime/src/helpers/redirects.ts
```diff
@@ -66,21 +58,6 @@ const generateLocaleRedirects = ({
   return redirects
 }
 
-export const generateStaticRedirects = ({
-  netlifyConfig,
```
ascorbic, line 70: No need to redirect, because the files are in the right place

packages/runtime/src/helpers/redirects.ts
```diff
@@ -66,21 +58,6 @@ const generateLocaleRedirects = ({
...
-}: {
-  netlifyConfig: NetlifyConfig
-  nextConfig: Pick<NextConfig, 'i18n' | 'basePath'>
-}) => {
-  // Static files are in `static`
-  netlifyConfig.redirects.push({ from: `${basePath}/_next/static/*`, to: `/static/:splat`, status: 200 })
-
-  if (i18n) {
-    netlifyConfig.redirects.push({ from: `${basePath}/:locale/_next/static/*`, to: `/static/:splat`, status: 200 })
-  }
```
ascorbic, line 81: Seems i18n still loads from `/_next` in the root, so it's not needed

packages/runtime/src/helpers/files.ts
```diff
@@ -74,32 +73,28 @@ export const getMiddleware = async (publish: string): Promise<Array<string>> =>
...
-  netlifyConfig: NetlifyConfig
-  target: 'server' | 'serverless' | 'experimental-serverless-trace'
+  distDir: string
   i18n: NextConfig['i18n']
   basePath?: string
+  publishDir
 }): Promise<void> => {
   console.log('Moving static page files to serve from CDN...')
-  const outputDir = join(netlifyConfig.build.publish, target === 'server' ? 'server' : 'serverless')
+  const outputDir = join(distDir, 'server')
```
sarahetter, line 87: Why have we removed serverless from here?

packages/runtime/src/helpers/redirects.ts
nickytonline: What are/were `HIDDEN_PATHS`?

ascorbic: They were the files in `.next` that we didn't want to be publicly-visible in the deploy, such as `cache`, `server` and various json files

packages/runtime/src/helpers/files.ts
ascorbic: We force builds to use `server` target anyway, and in Next 13 `serverless` breaks the build.

nickytonline (approved):
Looks good @ascorbic! 🚀

PR #1757, opened 2022-11-10
fix: correctly rewrite default locale ISR homepage to ODB handler

orinokai:
### Summary

ISR homepages with i18n were not correctly matched as static routes and were being served as SSR pages because no rewrite was created. This was because the matcher only worked for URLs with subpaths.

This PR updates the logic to match URLs prefixed with the default locale, with or without a subpath, and ensures that, when slicing off the locale, an index page at the root defaults to '/'.

### Test plan

Snapshots have been updated with the new redirects. Tests should pass as before.

Fixes netlify/pod-ecosystem-frameworks#233


ascorbic:
I think CodeQL's complaints could be fixed by casting the value with `Number()`

orinokai:
@ascorbic the new proxy validates that this is a number anyway, so I think it's fine to just ignore CodeQL in this case

orinokai:
Closing in favour of #1867 which is a more tidy version with the TV Maze API Proxy stuff split into a separate PR

PR #1761, opened 2022-11-11
feat: add next debug logging

orinokai:
### Summary

Occasionally we run into situations where it would be useful to know more about the origin response from the Next server to our handler functions.

This PR writes the origin headers and status code to the function logs when the request includes the header `x-next-debug-logging`.

### Test plan

1. Visit an SSR/ISR page that is routed to a handler function and include the above request header
2. Observe the origin response in the function logs

purple octopus


ascorbic (approved):
:shipit:

PR #1765, opened 2022-11-11
fix: ensure isr conditional 404 fix works for next 11

orinokai:
### Summary

[A previous PR](https://github.com/netlify/next-runtime/pull/1541) that fixed an issue with `fallback: false` also patched the Next server so that 404 pages could return a SWR header and so ISR pages that conditionally 404 will still have a TTL. However, it seems this patch was not compatible with Next 11. This PR addresses that.

### Test plan

Tests have been added as a new Next 11 demo site and should run and pass.

https://github.com/netlify/next-runtime/pull/1541


sarahetter (approved):
LGTM

packages/runtime/src/helpers/files.ts
```diff
@@ -366,7 +366,7 @@ const baseServerReplacements: Array<[string, string]> = [
     `checkIsManualRevalidate(process.env._REVALIDATE_SSG ? { headers: { 'x-prerender-revalidate': this.renderOpts.previewProps.previewModeId } } : req, this.renderOpts.previewProps)`,
   ],
   // ensure ISR 404 pages send the correct SWR cache headers
-  [`private: isPreviewMode || is404Page && cachedData`, `private: isPreviewMode && cachedData`],
+  [`private: isPreviewMode || is404Page`, `private: isPreviewMode || false`],
```
nickytonline, line 369: Are the `|| false`s necessary here and below? Would `isPreviewMode` be enough?

orinokai: I know it's weird, but they are necessary because of the way we patch/unpatch the files here: https://github.com/netlify/next-runtime/blob/main/packages/runtime/src/helpers/files.ts#L308

Without the `|| false` it will think the file has already been patched because `private: isPreviewMode` already exists in the original file.

nickytonline (approved):
🚀

.github/workflows/cypress-next-11.yml
```diff
@@ -0,0 +1,73 @@
...
+      - 'cypress/integration/next-11/**/*.{ts,js}'
+      - 'src/**/*.{ts,js}'
+jobs:
+  cypress:
+    name: Cypress
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        containers: [1, 2, 3, 4]
```
ericapisani, line 19: If there's not many tests that are included in this suite, it's likely not worth having so many containers dedicated to this. However, I'd leave it as a `matrix` so that as we add more tests it's really easy to add containers to keep the time to run the tests low.

```suggestion
        containers: [1]
```

demos/next-11/.eslintrc
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "next",
+  "root": true
```
ericapisani, line 3: In case anyone else is interested in what this property does (because I didn't know offhand either), from the ESLint docs site:

> By default, ESLint looks for configuration files in all parent folders up to the root directory. This can be useful if you want all of your projects to follow a certain convention, but can sometimes lead to unexpected results. To limit ESLint to a specific project, place "root": true inside the .eslintrc.* file or eslintConfig field of the package.json file or in the .eslintrc.* file at your project’s root level. ESLint stops looking in parent folders once it finds a configuration with "root": true.

@orinokai This is a bit of a nitpick, but it might be worth adding a small note here about ESLint not looking for configuration in parent directories past this point with this setting since this info is a little buried in the docs

demos/next-11/netlify.toml
```diff
@@ -0,0 +1,15 @@
+[build]
+command = "next build"
+publish = ".next"
+ignore = "if [ $CACHED_COMMIT_REF == $COMMIT_REF ]; then (exit 1); else git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF ../..; fi;"
+
+[[plugins]]
+package = "@netlify/plugin-nextjs"
```
ericapisani, line 7: I don't think this needs to be included since 

```
[[plugins]]
package = "@netlify/plugin-local-install-core"


[[plugins]]
package = "../plugin-wrapper"
```
 
is present in the file

ericapisani (commented):
Know this is still in a draft state, but noticed a few small things

nickytonline:
Going to go ahead and close this for now as we've done other work for 404s in relation to ISR.

PR #1775, opened 2022-11-15
chore(deps): upgrade everything to latest Next and React

ascorbic:
### Summary

Manually upgrades demo sites and packages to Next 13.0.3 and React 18.2.0. Fixes a unit test broken by this.

### Test plan

Check it all builds ok

### Standard checks:

- [ ] Add docs when necessary

---


ascorbic:
Something is up with the API route dependency resolution

demos/next-auth/netlify.toml
```diff
@@ -5,7 +5,6 @@ ignore = "if [ $CACHED_COMMIT_REF == $COMMIT_REF ]; then (exit 1); else git diff
 
 [build.environment]
 TERM = "xterm"
-NODE_VERSION = "17"
```
nickytonline, line 8: Is this safe to remove? I can't remember why we had this set.

demos/next-export/netlify.toml
```diff
@@ -5,7 +5,6 @@ ignore = "if [ $CACHED_COMMIT_REF == $COMMIT_REF ]; then (exit 1); else git diff
 
 [build.environment]
 NETLIFY_NEXT_PLUGIN_SKIP = "true"
-NODE_VERSION = "17"
```
nickytonline, line 8: Same as above

sarahetter (approved):

packages/next/test/request.spec.ts
```diff
@@ -1,6 +1,6 @@
 import Chance from 'chance'
 import { NextURL } from 'next/dist/server/web/next-url'
-import { NextCookies } from 'next/dist/server/web/spec-extension/cookies'
+import { RequestCookies } from 'next/dist/server/web/spec-extension/cookies'
```
nickytonline, line 3: Is this to fixe the `RespsonseCookies.getAll()`?

packages/runtime/src/helpers/files.ts
```diff
@@ -356,7 +356,7 @@ export const getDependenciesOfFile = async (file: string) => {
     return []
   }
   const dependencies = await readJson(nft, 'utf8')
-  return dependencies.files.map((dep) => resolve(file, dep))
+  return dependencies.files.map((dep) => resolve(dirname(file), dep))
```
nickytonline, line 359: This fixes the tests, i.e. snapshots below, but do we have something in place to test this aside from the snapshot?

demos/next-auth/netlify.toml
ascorbic: It was for Cypress. It caused breakages with this update, as there are packages with `engines` fields that don't like it. It didn't seem to be needed anymore, and the tests still pass.

packages/next/test/request.spec.ts
ascorbic: @nickytonline No, this is related to changes in Next 13, where they switched from a single `NextCookies` class into separate `RequestCookies` and `ResponseCookies` classes

packages/runtime/src/helpers/files.ts
ascorbic: @nickytonline I added tests in #1780, which is a more comprehensive fix

PR #1777, opened 2022-11-15
fix: add longitude, latitude, and timezone to RequestData.geo

nickytonline:
### Test plan

Closes https://github.com/netlify/pod-ecosystem-frameworks/issues/305
Relates to https://github.com/netlify/pod-compute/issues/270
Relates to https://github.com/netlify/cli/pull/5231

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,7 @@
         "@babel/preset-env": "^7.15.8",
         "@babel/preset-typescript": "^7.16.0",
         "@delucis/if-env": "^1.1.2",
-        "@netlify/build": "^28.2.0",
+        "@netlify/build": "^28.2.1",
         "@netlify/eslint-config-node": "^7.0.0",
         "@testing-library/cypress": "^8.0.1",
         "@types/fs-extra": "^9.0.13",
@@ -3318,15 +3318,15 @@
       "dev": true
     },
     "node_modules/@netlify/build": {
-      "version": "28.2.0",
-      "resolved": "https://registry.npmjs.org/@netlify/build/-/build-28.2.0.tgz",
-      "integrity": "sha512-K2OKtnnYPhGqZr2FVPWWXI3CUVO+ok2K/1rVG2O74CUdr2s3M7b6XZF1qwosjToztq5u/iz7eupRNfoogKMQGQ==",
+      "version": "28.2.1",
+      "resolved": "https://registry.npmjs.org/@netlify/build/-/build-28.2.1.tgz",
+      "integrity": "sha512-VwDVmYoDQ+36fpASefgh0kS/qF1QZ0Rd6Bvt4Bf9sr3s/OuLSVDnqFL7TOyChWHlpHOXpCwISeH44HgKwjNAPA==",
       "dev": true,
       "dependencies": {
         "@bugsnag/js": "^7.0.0",
         "@netlify/cache-utils": "^5.0.2",
         "@netlify/config": "^20.0.1",
-        "@netlify/edge-bundler": "4.2.0",
+        "@netlify/edge-bundler": "4.3.0",
         "@netlify/functions-utils": "^5.0.4",
         "@netlify/git-utils": "^5.0.2",
         "@netlify/plugins-list": "^6.54.0",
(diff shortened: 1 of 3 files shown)


packages/runtime/src/templates/edge/runtime.ts
```diff
@@ -18,8 +18,9 @@ export interface RequestData {
     city?: string
     country?: string
     region?: string
-    latitude?: string
-    longitude?: string
+    latitude?: number
```
nickytonline, line 21: Just updating this as our context object type has it marked as number for longitude and latitude.

demos/middleware/middleware.ts
```diff
@@ -29,6 +29,18 @@ export async function middleware(req: NextRequest) {
...
   }
 
+  if (pathname.startsWith('/api/geo')) {
+    // {"country":"CA","region":"QC","city":"Montreal","longitude":-73.5848,"latitude":45.4995,"timezone":"America/Toronto"}
+    req.headers.set('x-geo-country', req.geo.country)
+    req.headers.set('x-geo-region', req.geo.region)
+    req.headers.set('x-geo-city', req.geo.city)
+    req.headers.set('x-geo-longitude', req.geo.longitude)
+    req.headers.set('x-geo-latitude', req.geo.latitude)
+    req.headers.set('x-geo-timezone', (req.geo as any).timezone)
```
nickytonline, line 39: TODO: sort out the `any`

nickytonline (commented):
In Netlify context, `longitude` and `latitude` are of type `number`, but Next.js types them as `string | undefined`. See https://nextjs.org/docs/api-reference/next/server#nextrequest. I wonder if on our end we should convert them to string to avoid any issues on the Next.js side. I do find it odd that they're not treated as numbers on the Next side though.

packages/runtime/src/templates/edge/runtime.ts
```diff
@@ -63,10 +63,12 @@ const handler = async (req: Request, context: Context) => {
     return
   }
 
-  const geo = {
+  const geo: RequestData['geo'] = {
     country: context.geo.country?.code,
     region: context.geo.subdivision?.code,
     city: context.geo.city,
+    latitude: context.geo.latitude?.toString(),
```
nickytonline, line 70: Converting to string her as this is what Next.js is expecting.

ascorbic: How about adding this to `packages/next/src/middleware/request.ts`:

```typescript
export type GeoData = NextRequest['geo'] & {
  timezone?: string
}
```

... and then type the getter `get geo(): GeoData`

ascorbic: Maybe we should declare our own `NextRequest` type inside `@netlify/next` that extends the original one, and export it from there. We can also type the `geo()` getter in `MiddlewareRequest`

nickytonline:
I believe the failures for the checks are Next 11 related.

packages/next/src/middleware/request.ts
```diff
@@ -1,9 +1,22 @@
...
+        city?: string
+        country?: string
+        region?: string
+        latitude?: string
+        longitude?: string
+        timezone?: string
+      }
+    | undefined
+}
+
```
ascorbic, line 19: [dust] Rather than redefining the built-in geo type, how about extending it. Less verbose and easier to maintain. e.g. 

```suggestion
export type NextRequest = InternalNextRequest & {
  geo: InternalNextRequest['geo'] & {
    timezone?: string
  }
}

```

nickytonline: For some reason I thought I had to reimplement the whole thing as it was an actual getter. Didn't realize you could jus tag that on like that like a merge declaration. I'll go ahead and make the change.

packages/next/src/middleware/request.ts
```diff
@@ -5,16 +5,9 @@ import type { NextRequest as InternalNextRequest } from 'next/server'
...
-        city?: string
-        country?: string
-        region?: string
-        latitude?: string
-        longitude?: string
-        timezone?: string
-      }
-    | undefined
+  get geo(): {
+    timezone?: string
```
ascorbic, line 9: Oh, I didn't realise it would inherit by default!

nickytonline: Yeah that what I thought you meant initially and not an intersection, but it looks like it works. 😎

ascorbic (approved):

sarahetter (approved):
:tada:

nickytonline:
There's Next 11 demo sites failing, but just going to see about the other site failures.

nickytonline:
Looks like all things Cypress are passing, so I think all the failures are related to Next 11 tests.

ascorbic:
If you look at details you can see that several demo sites failed to build. I don't think it's the fault of this PR (I think it's npm-related) but Cypress tests alone won't say that everything is passing.

nickytonline:
> If you look at details you can see that several demo sites failed to build. I don't think it's the fault of this PR (I think it's npm-related) but Cypress tests alone won't say that everything is passing.

Are the Cypress tests passing because it's not using the `DEPLOY URL`? I remember @MarcL mentioning something about that on Slack.

PR #1780, opened 2022-11-16
fix: resolve all pages using nft

### LINKED ISSUE #1776: Use nft.json to resolve page dependencies
Currently we use a API route's nft.json files to trace dependencies for the function bundle. However we don't do this for the main handler and ODB functions. For these we instead trace all of the source files. This works in most cases, but means we miss some important files, particularly non-source files.

A good example is [this test](https://github.com/netlify/next-runtime/actions/runs/3463080955/jobs/5782821634#step:5:618) from #1773, which fails because we're not bundling the static `hello.txt`, even though it is in the nft.json file.

We should resolve dependencies for all routes in a similar way to how we do them for API routes. This would catch these cases.

ascorbic:
### Summary

Currently we don't use Next.js's generated `js.nft.json` files when determining the dependencies for pages - we only use them for advanced API routes. Instead we directly trace the source files themselves. This means that we may miss some dependencies that Next.js knows about. For example, see #1776

This PR changes the resolver to use nft.json files for everything, and refactors all of it to reuse a lot more of the logic and make it more testable. It adds some tests for the generated files - mainly to ensure that they actually exist.

It adds a demo page similar to the one in #1776, which has a static file read using fs. 

### Test plan

1. Ensure the unit tests pass and the default demo works
2. Check [the page that uses a static file](https://deploy-preview-1780--netlify-plugin-nextjs-demo.netlify.app/getServerSideProps/file/) and ensure it shows the word "world"

Fixes #1776

image

### Standard checks:

- [ ] Add docs when necessary

---


packages/runtime/src/index.ts
```diff
@@ -155,7 +155,7 @@ const plugin: NetlifyPlugin = {
     const apiRoutes = await getExtendedApiRouteConfigs(publish, appDir)
 
     await generateFunctions(constants, appDir, apiRoutes)
-    await generatePagesResolver({ target, constants })
+    await generatePagesResolver({ constants })
```
nickytonline, line 158: [dust] Since there's only one property in the object being destructured, consider passing `constants` as the single argument.

packages/runtime/src/templates/getPageResolver.ts
```diff
@@ -1,30 +1,43 @@
...
-  const functionDir = posix.resolve(posix.join('.netlify', 'functions', HANDLER_FUNCTION_NAME))
-  const root = posix.resolve(slash(publish), target === 'server' ? 'server' : 'serverless', 'pages')
+export const getUniqueDependencies = async (sourceFiles: Array<string>) => {
+  const dependencies = await Promise.all(sourceFiles.map((sourceFile) => getDependenciesOfFile(sourceFile)))
+  return [...new Set([...sourceFiles, ...dependencies.flat()])].sort()
+}
 
-  const pages = await glob('**/*.js', {
+export const getAllPageDependencies = async (publish: string) => {
+  const root = posix.resolve(slash(publish), 'server')
```
nickytonline, line 20: I know it's only `'server'` now, but I can't remember the reason why.

test/index.js
```diff
@@ -1247,6 +1267,29 @@ describe('function helpers', () => {
...
+        await nextRuntime.onBuild(defaultArgs)
+        const dependencies = await getAllPageDependencies(constants.PUBLISH_DIR)
+        const filesExist = await Promise.all(
+          dependencies.map(async (dep) => {
+            const exists = await pathExists(dep)
+            return [dep, exists]
+          }),
+        )
+        console.log(filesExist)
+        expect(filesExist.every(([dep, exists]) => exists)).toBeTruthy()
```
nickytonline, line 1289: Is returning the `dep` necessary in the tuple?

```suggestion
        const filesExist = await Promise.all(
          dependencies.map(async (dep) => {
            const exists = await pathExists(dep)
            return exists
          }),
        )

        expect(filesExist.every((exists) => exists)).toBeTruthy()
```

ascorbic: Oh crap. Good spot. I put that in for debugging.

packages/runtime/src/templates/getPageResolver.ts
ascorbic: We force the target to be server now

nickytonline (approved):
Thanks for making the changes! 🚀

ascorbic:
WTF Kodiak?

ascorbic:
Oh, it's because I mentioned in the other one that this PR fixes the test failure! GitHub thought I meant that that PR "fixes" this one

ascorbic:
And it fixes the failing test! https://github.com/netlify/next-runtime/actions/runs/3481118957/jobs/5821791833

PR #1773, opened 2022-11-14
chore: add Next.js e2e tooling

ascorbic:
### Summary

Adds test tooling to allow us to run e2e tests from the Next.js monorepo. Most of the changes in this PR are files copied from the Next.js monorepo. See next section for details.

### Test plan

The following directories do _not_ need reviewing:

- `test/e2e/getserversideprops`: test and fixture site copied from Next.js monorepo
- everything in `test/e2e/next-test-lib` except `test/e2e/next-test-lib/next-modes/next-deploy.ts`

[This e2e test](https://github.com/netlify/next-runtime/actions/runs/3463080955/jobs/5782821634#step:5:618) is failing because it's caught a bug already! Issue: https://github.com/netlify/next-runtime/issues/1776 Fix: https://github.com/netlify/next-runtime/pull/1780

### Relevant links (GitHub issues, Notion docs, etc.) or a picture of cute animal

smol capy

### Standard checks:

- [ ] Check the Deploy Preview's Demo site for your PR's functionality
- [ ] Add docs when necessary

---

🧪 Once merged, make sure to update the version if needed and that it was published correctly.

PR #1782, opened 2022-11-16
chore: add the Next.js e2e test suite

ascorbic:
### Summary

Adds most of the e2e test suite from the Next.js monorepo. This is a follow-on from #1773, which added the tooling and one suite.

The approach I took was to copy across all tests, fix any failures that were problems with the tooling or errors in the tests, disable any that aren't appropriate to us (e.g. they're testing client-side stuff, or dev mode), and then move them into one of three folders. The `tests` folder has suites that fully pass with no changes. 

The `modified-tests` folder has tests that have been changed in some way - whether to disable individual tests or to fix errors in the tests. This is so we don't try to update them from upstream without checking this. I have disabled any failing tests and commented each with `// NTL Fail`. The skipped tests can be run by setting the env var `RUN_SKIPPED_TESTS`.

Finally there are `disabled-tests`. These are suites that we do want to pass, but currently fail all tests or almost all tests. Each of these is usually because of a single bug or incompatibility. These are all tracked in the umbrella issue.

Depending on which suites are enabled, these tests run around 30-50 site deploys. I have made changes to enable this to happen efficiently. The way I have done this is by generating a matrix of GitHub Actions jobs, with one per test file (most of which would generate one site). I use `jest-junit` to report the results, and these are uploaded as task artifacts. There is then a "Report" job that runs after the matrix, which combines these reports and generates a summary of all of the results. This can be seen in the action summary page.

The way that the tests are run is by re-implementing the `NextDeployInstance` class from the Next.js monorepo. The original would deploy to Vercel. This instead deploys to Netlify, using the Netlify CLI. It uses `ntl deploy --build --json`, which allows it to extract the deploy preview URL, against which it runs the tests. The tests all use the same site for the deploy previews, which does not have a linked repo. 

The builds are performed in the GitHub Action rather than in Netlify. The workflow file has comments explaining how it distributes the sites across dozens of worker jobs. The test tooling generates the site from the local fixture in a tmp directory, with the Netlify Next runtime installed as a file dependency. It uses yarn for this as it handles these better. If you pass the env var SITE_URL then it will run the tests against this rather than deploying the site. This only makes sense when testing an individual site (i.e. running one suite) because each suite uses a different test site.

### Test plan

1. Have a look at the test summary page. You can find it by clicking on the "Next.js e2e test suite" link in the Checks section below, and then clicking "Summary" in the sidebar.
2. There should be no need to review the tests themselves - so anything in `test/e2e/*-tests`

### Standard checks:

- [ ] Add docs when necessary

---


packages/runtime/src/templates/ipx.ts
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable n/no-missing-import, import/no-unresolved, @typescript-eslint/ban-ts-comment  */
```
ascorbic, line 1: This was the eslint prepush hook being clever

orinokai (approved):
Looks good to me. This is a such a great improvement (also a technical marvel!). Just two questions:
1. I'm wondering how well this would scale so we can also test previous versions, e.g. Next 11 and 12?
2. Would we want to make the Report job a required pass?

ascorbic:
1. I think we'd need specific tests for those, as these do test 13-specific features.
2. Yes, I was going to do that once this was merged

PR #1787, opened 2022-11-18
chore: attempt to fix plugin wrapper deps

ascorbic:
### Summary

An attempt to fix the "npm ERR! Cannot set properties of null (setting 'dev')" error

### Test plan

### Standard checks:

- [ ] Add docs when necessary

---


LekoArts:
This PR seems to have gotten stale and we don't seem to actively use that demo page anyways. If we run into problems into the future, we can revisit fixing this later.

PR #1801, opened 2022-11-22
fix: better headers.getAll polyfill

ascorbic:
### Summary

Last week I added [a polyfill for headers.getAll](https://github.com/netlify/next-runtime/pull/1766). However Next.js has now added a check to the edge function return value to see if it's an instanceof `Response`. This means that our polyfilled version wouldn;t match if the user returned e.g. `Response.redirect`, or `fetch()`. This PR changes the polyfill from a subclass to augmenting the prototype of `Headers`. 😬 

### Test plan

This was caught by the e2e tests PR, but it can be checked for regression by looking at the middleware deploy preview

capy gamer

### Standard checks:

- [ ] Add docs when necessary

---


nickytonline (approved):
Looks good and the middleware deploy preview page for cookie looks good as well. 🚀

PR #1809, opened 2022-11-24
fix: transform Regex named capture groups for Golang

### LINKED ISSUE #1794: Middleware matchers generate invalid Golang regexp capture groups
We currently generate edge functions manifests with patterns generated from the middleware-manifest matcher regexp. The problem is that this now uses named capture groups, which have a different syntax in golang than JS.

In JS, a named capture group [has the form `(?<id> ...`](https://regex101.com/r/id6FsW/1) but Golang needs [the equivalent to be `(?P<id> ...`](https://regex101.com/r/7MRxN8/1). Right now when converting the JS pattern into Golang we just strip lookeaheads. It seems we may also need to convert these named capture groups too.

This is caught by the e2e test `edge-render-getserversideprops`, which gives a 500 error for `/123` as the edge function does not match, causing it to try to use a lambda, which has invalid syntax

The code is in here: https://github.com/netlify/next-runtime/blob/main/packages/runtime/src/helpers/matchers.ts

orinokai:
### Summary

Quick fix to transform named captured groups in Golang syntax so that edge functions manifests are generated correctly.

**Update: PR now simply transforms named capture groups to standard capture groups.**

~This fix requires the edge-bundler fix linked below before it can be merged.~

### Test plan

The previous failing test `edge-render-getserversideprops` is still failing, but now due to a different issue which is captured in https://github.com/netlify/next-runtime/issues/1890.

Fixes #1794
~Requires netlify/edge-bundler#226~


packages/runtime/src/helpers/matchers.ts
```diff
@@ -24,6 +24,9 @@ export const stripLookahead = (regex: string) => {
   }
 }
 
+// The Go regexp lib has alternative syntax for named capture groups
+export const transformCaptureGroups = (regex: string) => regex.replace('(?<', '(?P<')
```
ascorbic, line 28:
```suggestion
export const transformCaptureGroups = (regex: string) => regex.replace('(?<', '(')
```

sarahetter (approved):

ascorbic (approved):

ascorbic:
@orinokai can this be merged?

orinokai:
currently looking into a failing test (`edge-render-getserversideprops`) regarding a problem with data routes

sarahetter:
Hey @orinokai 
We should work with @netlify/runtime-pod-compute after this is merged - right now the system is ignoring these regex match errors (suppressing warnings/errors), but we'd like to not have to suppress them.

PR #1812, opened 2022-11-24
fix: fix Invalid URL error when using uppercase i18n url

sarahetter:
### Summary

When adding a locale such as `de-DE` to next.config.ts, we were giving an Invalid URL error as it our cdn lowercases paths. By adding a rewrite when these locale redirects are happening, we are able to point at the correct site.

### Test plan

1. Visit the middleware  Deploy Preview `/de-DE/static` route
2. Observe that the page shows up without an `Invalid URL` error

Fixes https://github.com/netlify/pod-ecosystem-frameworks/issues/300
Screen Shot 2022-11-24 at 2 59 27 PM

### Standard checks:

- [ ] Add docs when necessary

---


jestSetup.js
```diff
@@ -1,2 +1,2 @@
-// eslint-disable-next-line n/no-unpublished-require
+
```
sarahetter, line 1: This is happening when I run `npm i` in the package 😕

packages/next/src/middleware/request.ts
```diff
@@ -60,7 +56,16 @@ export class MiddlewareRequest extends Request {
    */
   async next(options?: NextOptions): Promise<MiddlewareResponse> {
     this.applyHeaders()
-    const response = await this.context.next(options)
+    let response = await this.context.next(options)
```
nickytonline, line 59: [dust] I imagine this can stay as `const`.

nickytonline (approved):
Looks good. Left a dusty comment.

jestSetup.js
nickytonline: Does it cause a lint error? If not, it's fine that it's removed.

sarahetter: No more lint errors than are currently in main :D

packages/next/src/middleware/request.ts
```diff
@@ -60,7 +56,16 @@ export class MiddlewareRequest extends Request {
...
+
+    // next i18n responses contain a relative location header, and a 301 status
```
ascorbic, line 61: [dust] this isn't specifically a Next i18n thing: this is our CDN redirecting to a lowercase version of any static file request. In this scenario the i18n file is statically-generated, so the cdn redirects to the lowercase version

packages/next/src/middleware/request.ts
sarahetter: It can't as it's potentially getting redefined below in the if statement.

nickytonline: Ahh, didn't see that. 🙃

sarahetter:
@ascorbic / @nickytonline Just looked with Nick - PR looks fine, just a few of the deploy previews are failing because it's unable to find the new `@netlify/edge-functions` (and some are able to find it properly 😕 ) 

I'm away for December, can one of you take care of this while I'm gone?

ascorbic:
I think I know what's up. I'll take a look

ascorbic (approved):

PR #1852, opened 2022-12-21
test: re-enable unit tests

ascorbic:
### Summary

When we added the Next e2e tests, a mistake in the jest config meant that the `index.js` test no longer matched, and so the tests weren't being run. This PR re-enables them. However the snapshots have changed since then, which wasn't caught because the tests weren't being run.

I have run `git bisect` to find the commits that broke each of the tests, and have commented with details. They all seem to be fine.


test/__snapshots__/index.spec.js.snap
```diff
@@ -757,8 +646,8 @@ Array [
   Object {
     "force": false,
     "from": "/_next/data/build-id/en/getStaticProps/:id.json",
-    "status": 200,
-    "to": "/.netlify/builders/___netlify-odb-handler",
+    "status": 404,
+    "to": "/server/pages/en/404.html",
```
ascorbic, line 650: This was introduced in 2aa02dbe8140b2107c312e757e9dc1609c3b4e70. Presumably this is as expected, @orinokai?

test/__snapshots__/index.spec.js.snap
```diff
@@ -335,22 +344,6 @@ Array [
     "pages/en/getStaticProps/static.json",
     "_next/data/build-id/en/getStaticProps/static.json",
   ],
-  Array [
```
ascorbic, line 338: Introduced in a5b80475a89f5ab2266059ad2e96c8786ff41421 (app dir)

ascorbic: OK, this is because we no longer move ISR pages, because they're needed for the first request. https://github.com/netlify/next-runtime/blob/main/packages/runtime/src/helpers/files.ts#L158

ericapisani (approved):

nickytonline (approved):
🚀

PR #1855, opened 2022-12-22
fix: correctly handle ISR for appDir pages

### LINKED ISSUE #1844: [Bug]: appDir pages not varying cache on RSC header
### Summary

I tried fetching in Postman and *sometimes* Netlify will make user download a file with following content: https://pastebin.com/raw/QMZ7wVuB

which I assume is Next.js's server component props

### Steps to reproduce

View the site

https://github.com/Enime-Project/enime.moe/tree/next

### Next Runtime version

4.29.3-appdir.0

### What OS are you using?

None

### Your netlify.toml file

<details>

```toml
```

</details>

<details>

```toml
```

</details>

<details>

```toml
```

</details>

<details>
  <summary>Build logs</summary>

```
# Paste logs here
```

</details>

### Function logs

<details>

```
# Paste logs here
```

</details>

### .next JSON files

<details>

```
```

</details>

ascorbic:
### Summary
Data routes are handled differently for appDir pages. Instead of a JSON file in `/_next/data`, the request is made to the same path, passing the `RSC` header. This returns the RSC data to stream the update. The fact this varies on a header means we can't use simple ODB handling for these pages, because it will lead to errors like #1844, where the RSC data is returned when HTML is requested, or vice versa. 

This PR fixes this by adding an edge function which routes appDir requests that include the `rsc` header to the appropriate `.rsc` path instead. The original request path is put into a header, which then replaces the `.rsc` path in the handler. These are real files but not normally requested directly. The reason for doing this is so that these requests have unique URLs, meaning they're cached separately as ODBs. 

There's a final extra part that is needed, which is to handle trailing slashes. If we normalise all of the requests to the matching `.rsc` file, then there will be no difference to requests with and without the trailing slash. This is a problem, because depending on whether the version with or without the slash is requested first, we may cache the redirect instead of the actual data (or vice versa). To handle this, we do a bit of trickery. If the request has a trailing slash, we include a trailing slash on the rewritten `.rsc`path. For example, while an rsc request to `/blog` will be rewritten to `/blog.rsc`, a request for `/blog/` (with trailing slash) will be rewritten to `/blog.rsc/`. This means that it will be treated as a separate ODB path, and will be cached differently.

Basically, all of this is to implement `Vary` in ODBs. 🤷🏻 

### Test plan

1. Visit [an appDir page on the default demo site](https://deploy-preview-1855--netlify-plugin-nextjs-demo.netlify.app/blog/erica/). Navigate around, checking in the network panel that SPA data requests are correctly returning `application/octet-stream` RSC data. 
2. Load any of the `/blog` pages directly and ensure that HTML is returned
3. Try loading them with and without the trailing slash and see if they redirect correctly.

Fixes #1844

capy wants an apple

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/.github/workflows/e2e-appdir.yml
+++ b/.github/workflows/e2e-appdir.yml
@@ -41,7 +41,6 @@ jobs:
         env:
           NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_BOT_AUTH_TOKEN }}
           NETLIFY_SITE_ID: 1d5a5c76-d445-4ae5-b694-b0d3f2e2c395
-          NEXT_TEST_VERSION: canary
       - uses: actions/upload-artifact@v3
         if: ${{ always() }}
         name: Upload test results
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,12 +15,6 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        node-version: [14, '*']
-        exclude:
-          - os: macOS-latest
-            node-version: 14
-          - os: windows-latest
-            node-version: 14
       fail-fast: false
 
     steps:
@@ -32,15 +26,8 @@ jobs:
           check-latest: true
       - name: NPM Install
         run: npm install
-      - name: Switching to Node.js ${{ matrix.node-version }} to run tests
-        uses: actions/setup-node@v2
-        if: "${{ matrix.node-version != 'lts/*' }}"
-        with:
-          node-version: ${{ matrix.node-version }}
-          check-latest: true
       - name: Linting
         run: npm run format:ci
-        if: "${{ matrix.node-version == 'lts/*' }}"
       - name: Run tests against next@latest
         run: npm test
   canary:
@@ -50,12 +37,6 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        node-version: [14, '*']
-        exclude:
-          - os: macOS-latest
-            node-version: 14
-          - os: windows-latest
-            node-version: 14
       fail-fast: false
 
     if: github.ref_name == 'main'
@@ -70,11 +51,5 @@ jobs:
         run: npm install
       - name: Install Next.js Canary
         run: npm install -D next@canary --legacy-peer-deps
-      - name: Switching to Node.js ${{ matrix.node-version }} to run tests
-        uses: actions/setup-node@v2
-        if: "${{ matrix.node-version != 'lts/*' }}"
-        with:
-          node-version: ${{ matrix.node-version }}
-          check-latest: true
       - name: Run tests against next@canary
         run: npm test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Installing with LTS Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: 'lts/*'
+          node-version: 16
           check-latest: true
       - name: NPM Install
         run: npm install
@@ -45,7 +45,7 @@ jobs:
       - name: Installing with LTS Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: 'lts/*'
+          node-version: 16
           check-latest: true
       - name: NPM Install
         run: npm install
--- a/cypress/integration/default/appdir.spec.ts
+++ b/cypress/integration/default/appdir.spec.ts
@@ -2,7 +2,7 @@ describe('appDir', () => {
   it('renders appdir pages using SSR', () => {
     cy.request('/blog/erica/first-post/').then((response) => {
       expect(response.headers).to.have.property('x-nf-render-mode', 'ssr')
-      expect(response.headers).to.have.property('context-type', 'text/html; charset=utf-8')
+      expect(response.headers).to.have.property('content-type', 'text/html; charset=utf-8')
     })
   })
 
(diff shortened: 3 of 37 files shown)


NADESHIKON:
When is this getting merged?

NADESHIKON:
When is this getting merged?

.github/workflows/test.yml
```diff
@@ -15,32 +15,19 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        node-version: [14, '*']
```
ascorbic, line 18: AppDir requires 16+

test/__snapshots__/index.spec.js.snap
```diff
@@ -54,9 +60,9 @@ exports[`onBuild() generates a file referencing all API route sources: for _api_
 exports.resolvePages = () => {
     try {
         require.resolve('../../../.next/package.json')
-        require.resolve('../../../.next/server/chunks/274.js')
-        require.resolve('../../../.next/server/chunks/7016.js')
-        require.resolve('../../../.next/server/chunks/7590.js')
+        require.resolve('../../../.next/server/chunks/CHUNK_ID.js')
```
ascorbic, line 63: These chunk name snapshots are too flaky

test/__snapshots__/index.spec.js.snap
```diff
@@ -1265,6 +1295,18 @@ Array [
     "status": 404,
     "to": "/404.html",
   },
+  Object {
+    "force": false,
+    "from": "/blog/:author",
+    "status": 200,
+    "to": "/.netlify/functions/___netlify-handler",
+  },
```
ascorbic, line 1303: RSC pages use SSR

test/index.spec.js
```diff
@@ -59,6 +59,8 @@ const utils = {
   },
 }
 
+const normalizeChunkNames = (source) => source.replaceAll(/\/chunks\/\d+\.js/g, '/chunks/CHUNK_ID.js')
```
ascorbic, line 62: The chunk id is too easy to break between platforms, and shouldn't indicate a test failure. This normalises them all to "CHUNK_ID"

.github/workflows/e2e-appdir.yml
```diff
@@ -41,7 +41,6 @@ jobs:
         env:
           NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_BOT_AUTH_TOKEN }}
           NETLIFY_SITE_ID: 1d5a5c76-d445-4ae5-b694-b0d3f2e2c395
-          NEXT_TEST_VERSION: canary
```
ascorbic, line 44: Run tests with the same version of next as the repo

cypress/integration/default/appdir.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  it('correctly redirects HTML requests for ISR pages', () => {
+    cy.request('/blog/erica').then((response) => {
+      expect(response.status).to.be('308')
+      expect(response.headers).to.have.property('location', '/blog/erica/')
+    })
+  })
+
+  // This needs trailing slash handling to be fixed
+  it.skip('correctly redirects HTML requests for static pages', () => {
```
ascorbic, line 61: Prerendered pages do the normal Netlify trailing slash handling. This shoudl work once we have the trialing slash edge funciton

packages/runtime/src/helpers/redirects.ts
```diff
@@ -142,30 +146,41 @@ const generateStaticIsrRewrites = ({
...
       staticIsrRewrites.push(
         // No i18n, because the route is already localized
-        ...redirectsForNextRoute({ route, basePath, to: ODB_FUNCTION_PATH, force: true, buildId, i18n: null }),
+        ...redirectsForNextRoute({
+          route,
+          basePath,
+          to: ODB_FUNCTION_PATH,
+          force: true,
+          buildId,
+          dataRoute: isAppDir ? dataRoute : null,
```
ascorbic, line 181: appDir routes provide their own data route value

orinokai:
@ascorbic is there a reason we can't just not normalise the request? so a request to /blog will be rewritten to /blog.rsc and a request to /blog/ will be rewritten to /blog.rsc/

it would make it easier to reason about when looking at logs, etc. but let me know if i've misunderstood something

ascorbic:
I tried that at first, and it didn't work. I can't for the life of me rememebr why! Let me see if I can reproduce it. You're right it would be simpler. I'm now wondering if other changes I made sinc ethen will make it work.

ascorbic:
@orinokai OK, so it didn't work right away, but with a few changes it does! Good suggestion. I've updated the description too.

packages/runtime/src/helpers/utils.ts
```diff
@@ -63,21 +63,38 @@ export const toNetlifyRoute = (nextRoute: string): Array<string> => {
...
-]
+export const generateNetlifyRoutes = ({
+  route,
+  dataRoute,
+  withData = true,
+}: {
+  route: string
+  dataRoute: string
+  withData: boolean
+}) => [...(withData ? toNetlifyRoute(dataRoute) : []), ...toNetlifyRoute(route)]
```
orinokai, line 74: Will be so good to replace all these helper functions with edge router!

orinokai (approved):
Amazing! Nice one for sticking with it. 🚢

gpoole:
@ascorbic the issue described in #1844 and fixed by this PR unfortunately seems to have surfaced again as #2089. It's happening with auto-detected version of the runtime with Next 13.5.6 (latest released v13 as of now). Could it be the same underlying problem and, if so, potentially a regression?

ascorbic:
@gpoole yes, this is related to internal changes in Next.js. We're working on some major changes that will fix this and many more issues. Beta will be available soon.

gpoole:
That's great, thanks for the update 👍
-------- Original Message --------

PR #1451, opened 2022-07-08
chore: enable minimal mode

ascorbic:
### Summary

Minimal mode is an experimental feature. See the Next.js GitHub discussion about this feature, https://github.com/vercel/next.js/discussions/29801

### Test plan

1. Visit the Deploy Preview ([insert link to specific page]()) ...

### Standard checks:

- [ ] Check the Deploy Preview's Demo site for your PR's functionality
- [ ] Add docs when necessary

---

🧪 Once merged, make sure to update the version if needed and that it was published correctly.

PR #1866, opened 2023-01-04
feat: add Next 13 request header mutation to middleware

ericapisani:
### Test plan

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "build": "npm run build -w packages/next -w packages/runtime",
     "postinstall": "run-s build install-husky",
     "install-husky": "if-env CI=1 || husky install node_modules/@netlify/eslint-config-node/.husky",
-    "test": "run-s build:demo test:jest",
+    "test": "run-s build:demo test:jest test:deno",
+    "test:deno": "deno test packages/runtime/src/templates/edge-shared/",
     "test:next": "jest -c test/e2e/jest.config.js",
     "test:next:disabled": "RUN_SKIPPED_TESTS=1 jest -c test/e2e/jest.config.disabled.js",
     "test:next:all": "RUN_SKIPPED_TESTS=1 jest -c test/e2e/jest.config.all.js",
@@ -101,11 +102,9 @@
       "**/test/**/*.spec.ts",
       "!**/test/e2e/**",
       "!**/test/fixtures/**",
-      "!**/test/sample/**"
+      "!**/test/sample/**",
+      "!**/test/templates/edge-shared/**"
     ],
-    "moduleNameMapper": {
-      "https://deno.land/x/html_rewriter@v0.1.0-pre.17": "<rootDir>/test/mockedDenoModules/html_rewriter"
-    },
     "transform": {
       "\\.[jt]sx?$": "babel-jest"
     },
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "@babel/preset-env": "^7.15.8",
     "@babel/preset-typescript": "^7.16.0",
     "@delucis/if-env": "^1.1.2",
-    "@netlify/build": "^29.1.3",
+    "@netlify/build": "^29.4.0",
     "@netlify/eslint-config-node": "^7.0.1",
     "@testing-library/cypress": "^8.0.1",
     "@types/fs-extra": "^9.0.13",
@@ -88,7 +88,7 @@
     "typescript": "^4.3.4"
   },
   "dependencies": {
-    "next": "^13.0.6"
+    "next": "^13.0.7"
   },
   "engines": {
     "node": ">=16.0.0"
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "build": "npm run build -w packages/next -w packages/runtime",
     "postinstall": "run-s build install-husky",
     "install-husky": "if-env CI=1 || husky install node_modules/@netlify/eslint-config-node/.husky",
-    "test": "run-s build:demo test:jest test:deno",
+    "test": "run-s build:demo test:jest",
     "test:deno": "deno test packages/runtime/src/templates/edge-shared/",
     "test:next": "jest -c test/e2e/jest.config.js",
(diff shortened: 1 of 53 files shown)


packages/runtime/src/templates/edge-shared/utils.ts
```diff
@@ -59,6 +59,34 @@ function isMiddlewareResponse(response: Response | MiddlewareResponse): response
...
 
+// Next 13 supports request header mutations and has the side effect of prepending header values with 'x-middleware-request'
+// as part of invoking NextResponse.next() in the middleware. We need to remove that before sending the response the user
+// as the code that removes it in Next isn't run based on how we handle the middleware
+//
+// Related Next.js code:
+// * https://github.com/vercel/next.js/blob/68d06fe015b28d8f81da52ca107a5f4bd72ab37c/packages/next/server/next-server.ts#L1918-L1928
+// * https://github.com/vercel/next.js/blob/43c9d8940dc42337dd2f7d66aa90e6abf952278e/packages/next/server/web/spec-extension/response.ts#L10-L27
+export function updateModifiedHeaders(response: Response) {
+  const overriddenHeaders = response.headers.get('x-middleware-override-headers') || ''
```
ascorbic, line 70: No need for the fallback if you're checking for falsiness right away

packages/runtime/src/templates/edge-shared/utils.ts
```diff
@@ -59,6 +59,34 @@ function isMiddlewareResponse(response: Response | MiddlewareResponse): response
...
+    return response
+  }
+
+  const headersToUpdate = overriddenHeaders.split(',')
+
+  for (const header of headersToUpdate) {
+    const oldHeaderKey = 'x-middleware-request-' + header
+    const headerValue = response.headers.get(oldHeaderKey) || ''
+
+    response.headers.set(header, headerValue)
```
ascorbic, line 82: We don't want to set these in the response - it's the request that needs them, so we should be calling this before calling `context.next()`

ascorbic (commented):
Sorry, I somehow missed that this was still in draft! Added a few comments.

package.json
```diff
@@ -103,6 +103,9 @@
       "!**/test/fixtures/**",
       "!**/test/sample/**"
     ],
+    "moduleNameMapper": {
+      "https://deno.land/x/html_rewriter@v0.1.0-pre.17": "<rootDir>/test/mockedDenoModules/html_rewriter"
+    },
```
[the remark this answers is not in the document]
ascorbic: There's an example in the edge router PR: https://github.com/netlify/next-runtime/pull/1451/files#diff-c02e108e555d63f8e66ff0b099106e84ef940d69b5cb0e1d2c12b2f79c1f6322

There's a workflow for it in there too

ericapisani:
@ascorbic I don't mind the early feedback, thank you! 😄

packages/runtime/src/templates/edge-shared/utils.ts
ericapisani: The `request` headers appear to be fine in that they don't have the `x-middleware-request` string prefixed to every header, nor does it contain the `x-middleware-override-headers` header - it's the `result.response` headers that have them and are not having them cleaned up.

Would be happy to pair on this tomorrow though as there's likely more insight that you have that I'm missing

packages/runtime/src/templates/edge-shared/utils.ts
```diff
@@ -68,6 +96,9 @@ export const buildResponse = async ({
   request: Request
   context: Context
 }) => {
+  debugger
```
MarcL, line 99: A missed debugger statement that needs removing?

packages/runtime/src/templates/edge-shared/utils.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    it("does not modify headers that are missing 'x-middleware-request' in the name", () => {
+      assertEquals(mockResult.response.headers.get('foo'), 'bar')
+    })
+
+    it("removes 'x-middleware-request-' from headers", () => {
+      assertEquals(mockResult.response.headers.get('x-middleware-request-hello'), null)
+      assertEquals(mockResult.response.headers.get('x-middleware-request-test'), null)
+
+      assertEquals(mockResult.response.headers.get('hello'), 'world')
+      assertEquals(mockResult.response.headers.get('test'), '123')
```
ascorbic, line 50: These should not be defined.

packages/runtime/src/templates/edge-shared/utils.ts
```diff
@@ -59,6 +59,34 @@ function isMiddlewareResponse(response: Response | MiddlewareResponse): response
...
 }
 
+// Next 13 supports request header mutations and has the side effect of prepending header values with 'x-middleware-request'
+// as part of invoking NextResponse.next() in the middleware. We need to remove that before sending the response the user
+// as the code that removes it in Next isn't run based on how we handle the middleware
+//
+// Related Next.js code:
+// * https://github.com/vercel/next.js/blob/68d06fe015b28d8f81da52ca107a5f4bd72ab37c/packages/next/server/next-server.ts#L1918-L1928
+// * https://github.com/vercel/next.js/blob/43c9d8940dc42337dd2f7d66aa90e6abf952278e/packages/next/server/web/spec-extension/response.ts#L10-L27
+export function updateModifiedHeaders(response: Response) {
```
ascorbic, line 69: You need to pass the request to this, and set the headers on it not the response

ascorbic:
You need to add a test like this one, for the /api/hello route: https://github.com/netlify/next-runtime/blob/main/cypress/integration/middleware/enhanced.spec.ts#L8-L12

ascorbic:
This is the failing test for this: https://cloud.cypress.io/projects/yn8qwi/runs/1235/overview/3ed1f6bf-da05-44e1-bc69-f5f2451bed7c

ascorbic (approved):
Great work! One [pebble]: move the Cypress test from "enhanced" to "standard" because it's now testing the core feature.

PR #1867, opened 2023-01-05
fix: correctly rewrite default locale ISR homepage to ODB handler

orinokai:
### Summary

ISR homepages with i18n were not correctly matched as static routes and were being served as SSR pages because no rewrite was created. This was because the matcher only worked for URLs with subpaths.

This PR updates the logic to match URLs prefixed with the default locale, with or without a subpath, and ensures that, when slicing off the locale, an index page at the root defaults to '/'.

### Test plan

Snapshots have been updated with the new redirects. Tests should pass as before.

Fixes netlify/pod-ecosystem-frameworks#233


packages/runtime/src/helpers/redirects.ts
```diff
@@ -143,8 +143,8 @@ const generateStaticIsrRewrites = ({
       return
     }
     // The default locale is served from the root, not the localised path
-    if (i18n?.defaultLocale && route.startsWith(`/${i18n.defaultLocale}/`)) {
-      route = route.slice(i18n.defaultLocale.length + 1)
+    if (i18n?.defaultLocale && route.startsWith(`/${i18n.defaultLocale}`)) {
```
ascorbic, line 146: Wouldn't this cause false positives? e.g. if there's a site with `en` default locale, but also a route called `/entry`

orinokai: hi @ascorbic, it doesn't appear to - both this and the previous condition would match `/en/entry`

ascorbic: Right, but would this match `/entry` and turn it into `try`?

orinokai: ah, i see what you mean - no it won't because with i18n the routes are always matched with the locale prefix, so for the route `/entry` it will be matched against `/en/entry` and `/fr/entry`, etc.

ascorbic: Aah, gotcha. Unfortunately I just checked the manifest for the default demo in my ISR branch and it looks like appDir pages _don't_ have the prefix.

orinokai: great catch, i'd forgotten that appDir doesn't implement i18n - i've made the condition more specific, which is a good thing anyway [7cef5fe](https://github.com/netlify/next-runtime/pull/1867/commits/7cef5fefc1841fce80b61a79d734d24a900517de)

ascorbic (approved):
🎉

PR #1870, opened 2023-01-06
fix: merge Middleware and API route response cookies

### LINKED ISSUE #1797: Middleware should merge API route response cookies
Currently, calling `res.cookies.set` in middleware replaces any cookies set by an API route. Instead they should be merged. 

e2e test: `skip-trailing-slash-redirect`

taty2010:
### Summary
When adding the Middleware headers to the response we ended up overriding the cookies set by the API route and returning only the cookies set within middleware. This fix ensures that the cookies set within the Middleware and API route are merged, by using 'append' instead of 'set', and that they are in the correct order. 

Fixes #1797 
### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/packages/runtime/src/templates/edge-shared/utils.ts
+++ b/packages/runtime/src/templates/edge-shared/utils.ts
@@ -31,7 +31,7 @@ export const addMiddlewareHeaders = async (
   // We need to await the response to get the origin headers, then we can add the ones from middleware.
   const res = await originResponse
   const response = new Response(res.body, res)
-  let originCookies = response.headers.get('set-cookie')
+  const originCookies = response.headers.get('set-cookie')
   middlewareResponse.headers.forEach((value, key) => {
     response.headers.set(key, value)
     // Append origin cookies after middleware cookies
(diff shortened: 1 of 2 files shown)


packages/runtime/src/templates/edge-shared/utils.ts
```diff
@@ -31,8 +31,13 @@ export const addMiddlewareHeaders = async (
   // We need to await the response to get the origin headers, then we can add the ones from middleware.
   const res = await originResponse
   const response = new Response(res.body, res)
+  let originCookies = response.headers.get('set-cookie')
```
ericapisani, line 34: Because `originCookies` isn't being modified after being set, this can instead be written as

```suggestion
  const originCookies = response.headers.get('set-cookie')
```

ericapisani (changes requested):
One small thing, but otherwise looks great!

packages/runtime/src/templates/edge-shared/utils.ts
```diff
@@ -31,8 +31,13 @@ export const addMiddlewareHeaders = async (
   // We need to await the response to get the origin headers, then we can add the ones from middleware.
   const res = await originResponse
   const response = new Response(res.body, res)
+  const originCookies = response.headers.get('set-cookie')
   middlewareResponse.headers.forEach((value, key) => {
     response.headers.set(key, value)
+    // Append origin cookies after middleware cookies
+    if (key == 'set-cookie' && originCookies) {
```
ascorbic, line 38:
```suggestion
    if (key === 'set-cookie' && originCookies) {
```
We should (almost) always use strict comparison

ascorbic (approved):
Great stuff!

ericapisani (approved):

PR #1889, opened 2023-01-17
fix: added polyfill for process.env

### LINKED ISSUE #1864: process.env not polyfilled for Edge Functions in CLI
### Describe the bug

This report is coming from a user here: https://answers.netlify.com/t/82464

They're trying to use environment variables in their Next.js code using `process.env` which is native to any Next.js developer. Since the middleware is running on Edge Functions (in production as well as in CLI), the correct syntax is `Deno.env` instead of `process.env`. In our production environment, we polyfill this, so users can get a seamless experience. However, it appears that the polyfill is missing in CLI.

### Steps to reproduce

There's a code example in the thread above, which can cause this issue.

### Configuration

_No response_

### Environment

Sharing my system details, but should happen on any system (with current latest version of CLI - v12.5.0):

```
System:
    OS: macOS 13.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 504.87 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.12.1 - ~/.asdf/installs/nodejs/18.12.1/bin/node
    npm: 9.2.0 - ~/.asdf/plugins/nodejs/shims/npm
```

taty2010:
### Summary

Using `process.env` to access env vars within an edge function currently returns an error when running `netlify dev`. Updated next-dev file within edge to now polyfill `process.env` and allow it to work within edge functions.

Fixes #1864 
### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/packages/runtime/src/templates/edge/next-dev.js
+++ b/packages/runtime/src/templates/edge/next-dev.js
@@ -5,6 +5,7 @@ import { buildResponse } from '../edge-shared/utils.ts'
 
 globalThis.NFRequestContextMap ||= new Map()
 globalThis.__dirname = fromFileUrl(new URL('./', import.meta.url)).slice(0, -1)
+globalThis.process ||= { env: Deno.env.toObject() }
 
 // Next.js uses this extension to the Headers API implemented by Cloudflare workerd
 if (!('getAll' in Headers.prototype)) {


ascorbic (approved):
Looks great!

nickytonline (approved):
Thanks @taty2010! 🚢

PR #1892, opened 2023-01-18
fix: match edge runtime pages with optional trailing slash

### LINKED ISSUE #1798: Edge runtime routes throw 500 error on trailing slash mismatch
If trailing slash in disabled, requesting an edge runtime route with a trailing slash will throw a 500 error. This is because the matcher pattern strictly matches just the route without the slash, meaning the site tries to serve the route from the origin which will fail. The expected behaviour is to redirect to the canonical version.

The fix is the trailing slash fix as indicated in #1448, however this may need to happen before the edge router is implemented.

e2e test: `streaming-ssr`

ascorbic:
### Summary

Currently appDir edge runtime pages use the `matcher` regex to generate edge function pattern. However this regex is designed to match a normalised path, so doesn't include a trailing slash. This meant that some pages with the edge runtime did not match the edge function, causing server errors. 

This PR changes edge functions to not use the regex from the matcher for pages, but instead use the regex from the routes manifest. This includes the optional trailing slash.

To make this work, I also refactored the handling of these matchers, splitting the generation of the functions from the creation of the matchers, and no longer use the same function for middleware and functions as they use different logic.

### Test plan

Fixes #1798 

image

### Standard checks:

- [ ] Add docs when necessary

---


orinokai (approved):
🚀

packages/runtime/src/templates/edge/function-runtime.ts
```diff
@@ -0,0 +1,23 @@
+import type { Context } from 'https://edge.netlify.com'
```
ascorbic, line 1: I've split this into a separate entrypoint, because functions don't need custom matcher handling or support for advanced middleware - but _do_ want trailing slash handling unlike middleware

packages/runtime/src/templates/edge/middleware-runtime.ts
```diff
@@ -1,49 +1,13 @@
...
-import nextConfig from '../edge-shared/nextConfig.json' assert { type: 'json' }
 import edgeFunction from './bundle.js'
-import { buildResponse } from '../edge-shared/utils.ts'
+import { buildNextRequest, buildResponse } from '../edge-shared/utils.ts'
 import { getMiddlewareRouteMatcher, MiddlewareRouteMatch, searchParamsToUrlQuery } from '../edge-shared/next-utils.ts'
+import nextConfig from '../edge-shared/nextConfig.json' assert { type: 'json' }
 
 const matchesMiddleware: MiddlewareRouteMatch = getMiddlewareRouteMatcher(matchers || [])
 
-export interface FetchEventResult {
```
ascorbic, line 11: Moved all of this into a shared file, so that we can have two separate entrypoints. They could go in a layer eventually.

nickytonline (approved):
Testsed this with https://deploy-preview-1892--next-plugin-edge-middleware.netlify.app/api/edge and https://deploy-preview-1892--next-plugin-edge-middleware.netlify.app/api/edge/ and they both load the hello world example page.

CleanShot 2023-01-19 at 08 45 49

and the redirect from https://deploy-preview-1892--next-plugin-edge-middleware.netlify.app/api/edge/ to https://deploy-preview-1892--next-plugin-edge-middleware.netlify.app/api/edge

CleanShot 2023-01-19 at 08 46 56


🚀

PR #1893, opened 2023-01-19
fix: custom headers for root route when using i18n

### LINKED ISSUE #1883: [Bug]: Header rules in 'next.config.js' not observed for root route when statically built and using localized routing
### Summary

In NextJS you can configure customized headers for matching routes by having something like this in your `next.config.js`:
```
headers: async () => {
	return [
		{
			source: '/:path*',
			headers: [
				{
					key: 'X-Content-Type-Options',
					value: 'nosniff',
				},
			],
		},
	];
},
```

In this example, the source `/:path*` is meant to include **_all_** potential routes. We are finding one specific scenario where the root route (`/`) is not affected by these custom header rules - but all other page routes are returning the headers properly.

Specifically, we are finding that when using localization and localized routes, this rule in `next.config.js` above turns into the following rule in the `netlify.toml`:
```
[[headers]]
for = "/:nextInternalLocale(en\\-US|fr|de)/*"

  [headers.values]
  X-Content-Type-Options = "nosniff"
```

And apparently this `/:nextInternalLocale(en\\-US|fr|de)/*` does not include the root route (`/`). When not using localization, the same rule above instead turns into this:
```
[[headers]]
for = "/*"

  [headers.values]
  X-Content-Type-Options = "nosniff"
```

Which *does* include the root route.

So for now, in order to have the root route covered by custom header rules, we have to also setup the same rules in the `netlify.toml` file explicitly for the root route. Ideally this should be unnecessary and the root route should be covered by the `next.config.js` wildcard as occurs when not using localization.

Side note - we had tried defining an additional explicit root route (`/`) in the `next.config.js`, but that did not seem to have an effect. This was the `next.config.js`:
```
headers: async () => {
	return [
		{
			source: '/:path*',
			headers: securityHeaders,
		},
		{
			source: '/',
			headers: securityHeaders,
		}
	];
},
```

.. and this was the `netlify.toml` output:
```
[[headers]]
for = "/:nextInternalLocale(en\\-US|fr|de)/*"

  [headers.values]
  X-Content-Type-Options = "nosniff"

[[headers]]
for = "/:nextInternalLocale(en\\-US|fr|de)"

  [headers.values]
  X-Content-Type-Options = "nosniff"
```

Note that it still injected the `/:nextInternalLocale(en\\-US|fr|de)` part, which apparently makes it not match the root route (`/`).

### Steps to reproduce

Please see the following Netlify support ticket:
https://netlify.zendesk.com/agent/tickets/126167

In that ticket will be a link to a test site where you can test all pages of the site and get the proper headers (as defined in the `next.config.js`), except for the root route, which will instead return a different set of headers (as defined in the `netlify.toml`). Also can provide access to a sample code repo through that ticket.

_No response_

### Next Runtime version

12.1.0

### What OS are you using?

None

### Your netlify.toml file

<details>

```toml
```

</details>

<details>

```toml
```

</details>

<details>

```toml
```

</details>

<details>
  <summary>Build logs</summary>

```
# Paste logs here
```

</details>

### Function logs

<details>

```
# Paste logs here
```

</details>

### .next JSON files

<details>

```
```

</details>

orinokai:
### Summary

When generating custom headers, the Runtime reads Next.js config rules and translates them into the required format for the TOML file. Part of that translation involves converting the `:nextInternalLocale` placeholder to a single localised rule for each locale.

However, the condition for the conversion was not being matched when one or more of the locales included a special regex character because Next.js escapes the locale names before writing the routes-manifest.json file and we were not doing the same before matching.

This PR lifts the `escapeStringRegexp` function from Next.js and uses it to escape the locales before creating the path, in the same way that Next.js does.

Note: in this case the problematic character was a `-`, which isn't a special character in regex, but one that Next.js escapes regardless.

### Test plan

1. Tests have been amended to catch this condition.

Fixes #1883


ascorbic (approved):
Great. For future reference, we also have `escapeStringRegexp` available in `next-utils` for edge functions

nickytonline (approved):
Thanks for the fix @orinokai! 🚢

Issue #1890, opened 2023-01-18
[Bug]: edge rendered dynamic routes do not correctly parse slugs into params


orinokai:
### Summary

Dynamic routes rendered on the edge with `runtime: 'experimental-edge'` do not correctly parse the slug into the page parameters.

This is demonstrated by the failing e2e test: `edge-render-getserversideprops`

### Steps to reproduce

1. Clone the repro and deploy to Netlify
2. Visit `/123?hello=world`

Expected: `{"url":"/123?hello=world","query": {"hello":"world" ,"id":"123"},"params": {"id":"123"},"now": 1673955487763}`

Actual: `{"url":"/123?hello=world","query": {"hello":"world"},"now": 1673955487763}`

### A link to a reproduction repository

https://github.com/orinokai/test-next-edge-slug

### Next Runtime version

4.30.0

### More information about your build

- [ ] I am building using the CLI
- [ ] I am building using file-based configuration (`netlify.toml`)

### What OS are you using?

None

taty2010:
Marc and I were looking into this issue and are needing more info on what's going on. @ascorbic are you already working on this?

ascorbic:
I started looking at it, and it's a tricky one. It might require edge router

PR #1894, opened 2023-01-19
test: re-disable edge-render-gssp test

ascorbic:
### Summary

The `edge-render-getserversideprops` test was previously disabled, but was accidentally re-enabled as part of another PR before it was fixed. This disables it again until we can deal with #1890.


orinokai (approved):
Thanks, can remain disabled until  https://github.com/netlify/next-runtime/issues/1890

PR #1896, opened 2023-01-20
fix: correctly match params in edge runtime

### LINKED ISSUE #1890: [Bug]: edge rendered dynamic routes do not correctly parse slugs into params

ascorbic:
### Summary

Next.js was not populating the params from dynamic edge runtime pages. This PR fixes that by populating the `x-matched-path` header, which along with setting minimal mode, makes Next.js parse the params as if we're on a certain triangle-shaped host. This includes correctly populating the params.

Fixes #1890

### Test plan

1. Visit `/edge/123?hello=world` in the default demo
2. Check that query is `{ id: '123', hello: 'world' }` and params is `{ id: '123' }`

pão em forma de capivara

### Standard checks:

- [ ] Add docs when necessary

---


orinokai (approved):
Nice one, looks/works great 🚢

test/e2e/next-test-lib/next-test-utils.js
```diff
@@ -128,7 +128,9 @@ export function renderViaHTTP(appPort, pathname, query, opts) {
 export async function fetchViaHTTP(appPort, pathname, query = undefined, opts = undefined, useUndici = false) {
   const url = `${pathname}${typeof query === 'string' ? query : query ? `?${qs.stringify(query)}` : ''}`
   const fetch = useUndici ? undiciFetch : nodeFetch
-  return fetch(getFullUrl(appPort, url), opts)
+  const fullUrl = getFullUrl(appPort, url)
+  console.log(`Fetching url: ${fullUrl}`)
```
nickytonline, line 132: Was this for debugging purposes only?

test/e2e/modified-tests/edge-render-getserversideprops/index.test.ts
```diff
@@ -36,7 +37,7 @@ describe('edge-render-getserversideprops', () => {
     expect(props.url).toBe('/123?hello=world')
   })
 
-  it('should have correct query/params on rewrite', async () => {
+  usuallySkip('should have correct query/params on rewrite', async () => {
```
nickytonline, line 40: Are all the `usuallySkip` too flaky at the moment? If so, can we create an issue to deflakify?

packages/runtime/src/helpers/edge.ts
```diff
@@ -470,6 +472,10 @@ export const writeEdgeFunctions = async ({
         ]),
       )
 
+      const dataRoutesMap = new Map(
```
nickytonline, line 475: Can you add some comments for the changes in this file? It looks like we create a map of data routes (if data routes exist) and set the values to the regex corresponding to the data route, and later on below, it looks like the edge function gets mapped to the specific data route?

And for the cache part below, that's for edge caching but only if using the app dir feature?

packages/runtime/src/templates/edge/function-runtime.ts
```diff
@@ -11,6 +11,7 @@ const handler = async (req: Request, context: Context) => {
     return redirect
   }
   const request = buildNextRequest(req, context, nextConfig)
+  request.headers['x-matched-path'] ||= edgeFunctionDefinition.page
```
nickytonline, line 14: Here, the `edgeFunctionDefinition.page` will populate the `x-matched-path` header only if it's not set, but from the issue this fixes, isn't the x-matched-path always not set, or is this because it's only when running on the edge in minimal mode?

test/e2e/next-test-lib/next-test-utils.js
ascorbic: No, I've added them to the test script because it makes it a lot easier to debug failures if you can see the actual URL that was loaded

packages/runtime/src/helpers/edge.ts
ascorbic: Sure thing. I'll add a comment. Re the caching, it's the same as the one right above: only enabled for appDir because it's experimental

packages/runtime/src/templates/edge/function-runtime.ts
ascorbic: This is for forward-compatibility: edge router will pre-populate this value, so we'll keep that value if it's set

test/e2e/modified-tests/edge-render-getserversideprops/index.test.ts
ascorbic: No, they're ones that haven't been implemented yet. This whole suite was previously disabled, but now the params work then we can enable it except for these two ones that need edge router. `usuallySkip` is just a way of having `it.skip` that can be enabled with an env var

nickytonline (approved):
Thanks for adding the comments @ascorbic! I tested this out per the test steps and works like a charm! 🚢 

image

PR #1903, opened 2023-01-24
fix: handle v1 edge function definition

ascorbic:
### Summary

Follow me, if you will, back to the summer of 2022.  Next.js middleware had gone to GA, which included updating the manifest format to include `functions` as well as `middleware`. Shortly after (the next patch release) they added support for middleware matchers, replacing the `regex` key on middleware defintions with an array of matchers, and doing the same for the new functions because why not add a new feature and make a breaking change to a format in a patch release. Anyway, for that brief period in June there was support for functions but no matchers. This PR makes those work.

### Test plan

1. Ensure the new unit tests pass. I'm not adding a demo site for a manifest format that was around for less time than the British prime minister who was running for office at that time. A unit test will have to do.

image

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/packages/runtime/src/helpers/edge.ts
+++ b/packages/runtime/src/helpers/edge.ts
@@ -348,12 +348,12 @@ export const writeRscDataEdgeFunction = async ({
   ]
 }
 
-const getEdgeFunctionPatternForPage = ({
+export const getEdgeFunctionPatternForPage = ({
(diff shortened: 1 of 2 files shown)


orinokai (approved):
🤣 at the description for this PR!

PR #1916, opened 2023-02-02
refactor: extract edge runtime shims into separate file

ascorbic:
### Summary

We provide a number of shims and polyfills to make our edge runtime work with Next.js. Currently they're a big string in a helper file. This PR extracts them into a file, meaning they can be properly typechecked.

### Test plan

This should have no effect on the output.

image

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/packages/runtime/src/helpers/edge.ts
+++ b/packages/runtime/src/helpers/edge.ts
@@ -4,7 +4,7 @@ import { resolve, join } from 'path'
 import type { NetlifyConfig, NetlifyPluginConstants } from '@netlify/build'
 import { greenBright } from 'chalk'
 import destr from 'destr'
-import { copy, copyFile, emptyDir, ensureDir, readJSON, readJson, writeJSON, writeJson } from 'fs-extra'
+import { copy, copyFile, emptyDir, ensureDir, readJSON, writeJSON, writeJson } from 'fs-extra'
 import type { PrerenderManifest } from 'next/dist/build'
 import type { MiddlewareManifest } from 'next/dist/build/webpack/plugins/middleware-plugin'
 import type { RouteHas } from 'next/dist/lib/load-custom-routes'
@@ -68,7 +68,7 @@ export interface FunctionManifest {
 
 const maybeLoadJson = <T>(path: string): Promise<T> | null => {
   if (existsSync(path)) {
-    return readJson(path)
+    return readJSON(path)
   }
 }
 export const isAppDirRoute = (route: string, appPathRoutesManifest: Record<string, string> | null): boolean =>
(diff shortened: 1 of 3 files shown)


packages/runtime/src/templates/edge/shims.ts
```diff
@@ -0,0 +1,59 @@
...
+  var process: {
+    env: Record<string, string>
+  }
+  var EdgeRuntime: string
+  var AsyncLocalStorage: typeof ALSCompat
+  var _ASSETS: Record<string, string>
+}
+
+// Deno defines "window", but naughty libraries think this means it's a browser
+delete (globalThis as Omit<typeof globalThis, 'window'> & Pick<Partial<typeof globalThis>, 'window'>).window
```
ascorbic, line 20: This type makes `window` optional

packages/runtime/src/templates/edge/shims.ts
```diff
@@ -0,0 +1,59 @@
...
+ * This file isn't imported, but is instead inlined along with other chunks into the edge bundle.
+ */
+
+declare global {
+  var process: {
+    env: Record<string, string>
+  }
+  var EdgeRuntime: string
+  var AsyncLocalStorage: typeof ALSCompat
+  var _ASSETS: Record<string, string>
```
ascorbic, line 16: This is defined by Next.js in the function bundle.

packages/runtime/src/helpers/edge.ts
```diff
@@ -4,7 +4,7 @@ import { resolve, join } from 'path'
 import type { NetlifyConfig, NetlifyPluginConstants } from '@netlify/build'
 import { greenBright } from 'chalk'
 import destr from 'destr'
-import { copy, copyFile, emptyDir, ensureDir, readJSON, readJson, writeJSON, writeJson } from 'fs-extra'
+import { copy, copyFile, emptyDir, ensureDir, readJSON, writeJSON, writeJson } from 'fs-extra'
```
nickytonline, line 7: I just notice there’s the same function name twice in fs-extra but with different casings. Odd.

nickytonline (approved):
🧹

packages/runtime/src/helpers/edge.ts
ascorbic: Yeah, I think that it's a legacy thing. Confusing, as I'm not sure which is "preferred"

MarcL:
Could we add tests for the function's code now? Worth doing that?

nickytonline:
> Could we add tests for the function's code now? Worth doing that?

I believe the snapshot should handle that?

ascorbic:
Yeah, these are all comprehensively tested elsewhere.

PR #1919, opened 2023-02-08
fix: correctly strip negative lookaheads from patterns with a forward slash

orinokai:
### Summary

The current code to strip regex lookaheads does not work when the regex pattern contains a forward slash. This is because the transform attempts to compile the regex from a string that is itself delimited by forward slashes.

This PR instead creates a RegEx object from the un-delimited string and passes that to the transform function.

### Test plan

The tests are also updated to use unescaped regex strings because this is how they are written in the `routes-manifest.json` file and why the original tests didn't catch the error.

Fixes https://github.com/netlify/pod-ecosystem-frameworks/issues/313


nickytonline (commented):
The fix looks good, but a bunch of deploy previews are failing.

nickytonline:
The windows tests failing is a known issue and can be ignored for now.

```
info  - Need to disable some ESLint rules? Learn more here: https://nextjs.org/docs/basic-features/eslint#disabling-rules
info  - Collecting page data...
ReferenceError: self is not defined
```

test/helpers/matchers.spec.ts
```diff
@@ -46,10 +46,16 @@ describe('the middleware path matcher', () => {
       '^(?:\\/(_next\\/data\\/[^/]{1,}))?(?:\\/([^/.]{1,}))\\/shows(?:\\/((?!99|88).*))(.json)?[\\/#\\?]?$',
       '^(?:\\/(_next\\/data\\/[^/]{1,}))?\\/shows(?:\\/((?!99|88).*))(.json)?[\\/#\\?]?$',
     ]
-    for (const regex of regexes) {
-      const stripped = stripLookahead(regex)
-      expect(regex).toMatch(/\(\?!/)
-      expect(stripped).not.toMatch(/\(\?!/)
+
+    const expected = [
```
nickytonline, line 50: I made these tests more explicit @orinokai.

test/e2e/app-dir/app-edge/pages/pages-edge.tsx
```diff
@@ -2,4 +2,4 @@ export default function Page() {
   return <p>pages-edge-ssr</p>
 }
 
-export const config = { runtime: 'edge' }
+export const config = { runtime: 'experimental-edge' }
```
nickytonline, line 5: @orinokai @sarahetter I put this back to `experimental-edge` as the build was complaining when I finally got this running locally. Maybe the Next version isn't up to date for the E2E tests we pull in that expects `runtime: 'edge'`? I'll take a peek at that and open an issue if that's the case.

test/helpers/matchers.spec.ts
pieh: I am able to get this test pass with current code that uses current
```
transform(`/${regex}/`
```

It doesn't seem like we have input that would actually validate the problem that we currently have?

nickytonline: Yeah I reverted it and it seems fine. even with the double escape. And the middleware demo still fails with a regex error (extra backslash), so still digging into the actual issue.

nickytonline:
The tests have been updated in here to be more explicit as mentioned, but I think the stripping of lookaheads was working prior to this PR. I think the main issue is in the edge bundler which I have a fix up for in https://github.com/netlify/edge-bundler/pull/376 .

LekoArts (changes requested):
Please note that in https://github.com/netlify/next-runtime/pull/2056/files#diff-7b3bd298efba211ef665c3eb9b70c0bb96b291a0c82a77780f4fde5b710c68d2 I actually shipped the initial fix of this PR.

IMO the current change of this PR then isn't necessary if the fix lies inside the edge-bundler. The tests as is are fine then. I don't like snapshots, in this case they add no value, since the first 5 regex don't contain a lookahead pattern.

So I'd propose to close this PR

nickytonline:
https://github.com/netlify/edge-bundler/pull/378 is merged, so going to go ahead close this.

Issue #1800, opened 2022-11-22
Middleware data redirects should normalize destination


ascorbic:
In middleware [we intercept redirects](https://github.com/netlify/next-runtime/blob/mk/all-the-tests/packages/runtime/src/templates/edge-shared/utils.ts#L163) inside data requests (i.e. those in `/_next/data/..`) and set the `x-next-redirect` header instead. This is because the router handles the redirect instead. However we're currently not normalizing the redirect target. The correct behaviour should be to always set it to the normalized path, even if the target is for the data URL. e.g. a redirect to `/_next/data/build-id/en-gb/hello.json?foo=bar` should be normalized to `/en-gb/hello?foo=bar`. Next.js implements this [here](https://github.com/vercel/next.js/blob/canary/packages/next/shared/lib/router/utils/get-next-pathname-info.ts#L46)

e2e test: `middleware-redirects`

taty2010:
Looks like the files have changed since this was created. 

First link: https://github.com/netlify/next-runtime/blob/main/packages/runtime/src/templates/edge-shared/utils.ts#L250

Second link: https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/router/utils/get-next-pathname-info.ts#L46

PR #1928, opened 2023-02-16
fix: updated redirect data urls

taty2010:
### Summary

This update will normalize data redirects 

https://github.com/netlify/next-runtime/issues/1800

### Test plan

### Standard checks:

- [ ] Add docs when necessary

---


packages/runtime/src/templates/edge-shared/utils.ts
```diff
@@ -6,7 +6,24 @@ export interface FetchEventResult {
   waitUntil: Promise<any>
 }
 
-type NextDataTransform = <T>(data: T) => T
+function normalizeDataUrl(redirect, url) {
```
nickytonline, line 9: I forgot to mention it on our Zoom, but it'd be good to have a link to where this is done in Next.js code to give more context.

taty2010: That's a great idea, thanks!

taty2010:
The middleware-redirect test is failing within the PR but not locally.

packages/runtime/src/templates/edge-shared/utils.ts
```diff
@@ -250,6 +266,12 @@ export const buildResponse = async ({
     res.headers.set('x-nextjs-redirect', relativizeURL(redirect, request.url))
   }
 
+  const nextRedirect = res.headers.get('x-nextjs-redirect')
```
nickytonline, line 269: We tested this together and it works well. I still have no idea why we have to set the header first, get it and the reset it. If anyone knows why, I'm curious. 🕵🏻

nickytonline (approved):
I see it's all good now. Ubuntu tests are still hanging, but given the other OSes for tests are passing, going to go ahead and approve. If the Ubuntu tests are still hanging, try restarting the job.

:shipit:

nickytonline:
Just checking in here @taty2010. Anything else need to be done here or are the test failures valid failures?

sarahetter (approved):

PR #1950, opened 2023-03-01
feat: refresh hooks api implementation

orinokai:
### Test plan

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/demos/default/pages/api/revalidate.js
+++ b/demos/default/pages/api/revalidate.js
@@ -1,0 +1,8 @@
+export default async function handler(req, res) {
+  try {
+    await res.revalidate('/getStaticProps/with-revalidate/')
+    return res.json({ revalidated: true })
+  } catch (err) {
+    return res.status(500).send('Error revalidating')
+  }
+}
--- a/packages/runtime/src/helpers/functions.ts
+++ b/packages/runtime/src/helpers/functions.ts
@@ -48,6 +48,10 @@ export const generateFunctions = async (
     await writeFile(join(functionsDir, functionName, `${functionName}.js`), apiHandlerSource)
     await copyFile(bridgeFile, join(functionsDir, functionName, 'bridge.js'))
     await copyFile(
+      join(__dirname, '..', '..', 'lib', 'templates', 'server.js'),
+      join(functionsDir, functionName, 'server.js'),
+    )
+    await copyFile(
       join(__dirname, '..', '..', 'lib', 'templates', 'handlerUtils.js'),
       join(functionsDir, functionName, 'handlerUtils.js'),
     )
@@ -67,6 +71,10 @@ export const generateFunctions = async (
     await ensureDir(join(functionsDir, functionName))
     await writeFile(join(functionsDir, functionName, `${functionName}.js`), handlerSource)
     await copyFile(bridgeFile, join(functionsDir, functionName, 'bridge.js'))
+    await copyFile(
+      join(__dirname, '..', '..', 'lib', 'templates', 'server.js'),
+      join(functionsDir, functionName, 'server.js'),
+    )
     await copyFile(
       join(__dirname, '..', '..', 'lib', 'templates', 'handlerUtils.js'),
       join(functionsDir, functionName, 'handlerUtils.js'),
--- a/packages/runtime/src/templates/getHandler.ts
+++ b/packages/runtime/src/templates/getHandler.ts
@@ -5,10 +5,7 @@ import { outdent as javascript } from 'outdent'
 
 import type { NextConfig } from '../helpers/config'
 
-import type { NextServerType } from './handlerUtils'
-
 /* eslint-disable @typescript-eslint/no-var-requires */
-
 const { promises } = require('fs')
 const { Server } = require('http')
 const path = require('path')
@@ -22,9 +19,9 @@ const {
   getMaxAge,
   getMultiValueHeaders,
   getPrefetchResponse,
-  getNextServer,
   normalizePath,
 } = require('./handlerUtils')
+const { NetlifyNextServer } = require('./server')
 /* eslint-enable @typescript-eslint/no-var-requires */
 
 type Mutable<T> = {
@@ -76,8 +73,7 @@ const makeHandler = (conf: NextConfig, app, pageRoot, staticManifest: Array<[str
     const port = Number.parseInt(url.port) || 80
     base = url.origin
 
-    const NextServer: NextServerType = getNextServer()
-    const nextServer = new NextServer({
+    const nextServer = new NetlifyNextServer({
       conf,
       dir,
       customServer: false,
@@ -180,6 +176,7 @@ export const getHandler = ({ isODB = false, publishDir = '../../../.next', appDi
   // We copy the file here rather than requiring from the node module
   const { Bridge } = require("./bridge");
   const { augmentFsModule, getMaxAge, getMultiValueHeaders, getPrefetchResponse, getNextServer, normalizePath } = require('./handlerUtils')
+  const { NetlifyNextServer } = require('./server')
 
   ${isODB ? `const { builder } = require("@netlify/functions")` : ''}
   const { config }  = require("${publishDir}/required-server-files.json")
--- a/packages/runtime/src/templates/getHandler.ts
+++ b/packages/runtime/src/templates/getHandler.ts
@@ -65,7 +65,7 @@ const makeHandler = (conf: NextConfig, app, pageRoot, staticManifest: Array<[str
   // We memoize this because it can be shared between requests, but don't instantiate it until
   // the first request because we need the host and port.
   let bridge: NodeBridge
-  const getBridge = (event: HandlerEvent): NodeBridge => {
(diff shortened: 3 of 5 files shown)


packages/runtime/src/templates/server.ts
```diff
@@ -0,0 +1,73 @@
...
+    super(options)
+    this.netlifyRevalidateToken = options.netlifyRevalidateToken
+  }
+
+  public getRequestHandler(): NodeRequestHandler {
+    const handler = super.getRequestHandler()
+    return async (req, res, parsedUrl) => {
+      if (req.headers['x-prerender-revalidate']) {
+        if (this.netlifyRevalidateToken) {
+          await this.netlifyRevalidate(req.url)
```
dwwoelfel, line 26: Will this revalidate both the `html` page and the `json` page, e.g. `/posts/1` and `/_next/data/xtcOaw2OikW05zh4TC3_7/posts/1.json`?

packages/runtime/src/templates/server.ts
```diff
@@ -0,0 +1,73 @@
...
+
+import { getNextServer, NextServerType } from './handlerUtils'
+
+const NextServer: NextServerType = getNextServer()
+
+interface NetlifyNextServerOptions extends Options {
+  netlifyRevalidateToken: string
+}
+
+class NetlifyNextServer extends NextServer {
```
nickytonline, line 13: This is looking good @orinokai. I haven't tried it, but I wonder if it makes more sense to use a [Proxy](https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/proxy) instead of extending the NextServer. 🤔

orinokai: I initially attempted to implement this with dependency injection, rather than inheritance, but the `NextServer` recursively calls `this.getRequestHandler` during the `res.revalidate` call and I need the nested call to still hit our overridden version. I couldn't work out how to do that with DI and I think it might be the same using a Proxy?

packages/runtime/src/templates/server.ts
orinokai: It won't at the moment, you're right. The next piece of work, for the next milestone, is to generate the right paths for data routes, i18n and RSC.

packages/runtime/src/templates/server.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    netlifyApiFetch: jest.fn(({ payload }) => {
+      switch (payload.paths[0]) {
+        case '/getStaticProps/with-revalidate/':
+          return Promise.resolve({ code: 200, message: 'Revalidated' })
+        case '/not-a-path/':
+          return Promise.resolve({ code: 404, message: '404' })
+        default:
+          return Promise.reject(new Error('Error'))
+      }
+    }),
```
MarcL, line 23: [dust] I'm not a big fan of test fixtures having logic in them as they're much harder to reason with. Maybe look to have per-test mocking instead?

packages/runtime/src/templates/getHandler.ts
```diff
@@ -22,17 +19,17 @@ const {
...
 /* eslint-enable @typescript-eslint/no-var-requires */
 
 type Mutable<T> = {
   -readonly [K in keyof T]: T[K]
 }
 
 // We return a function and then call `toString()` on it to serialise it as the launcher function
-// eslint-disable-next-line max-params
+// eslint-disable-next-line max-params, max-lines-per-function
 const makeHandler = (conf: NextConfig, app, pageRoot, staticManifest: Array<[string, string]> = [], mode = 'ssr') => {
```
nickytonline, line 33: Maybe for another PR, but this looks like this could be broken up as I see we disable the ESLint rule `max-lines-per-function`.

packages/runtime/src/templates/server.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const NextServer: NextServerType = getNextServer()
+
+interface NetlifyConfig {
+  revalidateToken?: string
+}
+
+class NetlifyNextServer extends NextServer {
+  private netlifyConfig: NetlifyConfig
+  private netlifyPrerenderManifest: PrerenderManifest
```
nickytonline, line 22: [dust] Opt for [private class members](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) instead of TypeScript's `private`.

```suggestion
  #netlifyConfig: NetlifyConfig
  #netlifyPrerenderManifest: PrerenderManifest
```

and then in the code, e.g. 

```javascript
    const manifest = this.getPrerenderManifest()
    this.#netlifyPrerenderManifest = {
      ...manifest,
      routes: { ...manifest.routes },
      dynamicRoutes: { ...manifest.dynamicRoutes },
    }
```

demos/default/pages/api/revalidate.js
```diff
@@ -0,0 +1,25 @@
+export default async function handler(req, res) {
+  const query = req.query
+  const select = Number(query.select) || 0
+
+  const paths = [
```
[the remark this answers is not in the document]
orinokai: thanks, comment added in https://github.com/netlify/next-runtime/pull/1950/commits/a19778353c23c837be2449823f4060394be49d93

packages/runtime/src/templates/getHandler.ts
orinokai: yes 100%, @ascorbic and I discussed this too - it would be good to refactor this handler so that it is more like the edge handler and is a module itself, rather than a string output - then we can refactor it more easily into manageable chunks

packages/runtime/src/templates/server.ts
orinokai: as this is a subclass, i decided to follow the convention in the Next.js Server itself, so i think it makes sense to stick with the current syntax for consistency

packages/runtime/src/templates/getHandler.ts
nickytonline: Cool cool. I'll create an issue for it.

packages/runtime/src/templates/server.ts
nickytonline: Gotcha. Thanks for the context!

packages/runtime/src/helpers/functions.ts
```diff
@@ -47,6 +47,10 @@ export const generateFunctions = async (
     await ensureDir(join(functionsDir, functionName))
     await writeFile(join(functionsDir, functionName, `${functionName}.js`), apiHandlerSource)
     await copyFile(bridgeFile, join(functionsDir, functionName, 'bridge.js'))
+    await copyFile(
+      join(__dirname, '..', '..', 'lib', 'templates', 'server.js'),
+      join(functionsDir, functionName, 'server.js'),
```
sarahetter, line 52: Can we get some comments on why we're copying this file? (This whole functions.ts file could use those tbh hah, but I won't put that on you)

packages/runtime/src/templates/handlerUtils.ts
```diff
@@ -222,3 +223,71 @@ export const normalizePath = (event: HandlerEvent) => {
...
+export const unlocalizeRoute = (route: string, i18n: { defaultLocale: string; locales: string[] }): string =>
+  isLocalized(route, i18n) ? `/${route.split('/').slice(2).join('/')}` : route
+
+// Add the default locale prefix to a route (if necessary)
+export const localizeRoute = (route: string, i18n: { defaultLocale: string; locales: string[] }): string =>
+  isLocalized(route, i18n) ? route : normalizeRoute(`/${i18n.defaultLocale}${route}`)
+
+// Normalize a data route to include the locale prefix and remove the index suffix
+export const localizeDataRoute = (dataRoute: string, localizedRoute: string): string => {
+  if (dataRoute.endsWith('.rsc')) return dataRoute
```
sarahetter, line 288: `.rsc` should probably be a constant at this point

packages/runtime/src/helpers/functions.ts
orinokai: yep, i totally agree! i don't know enough about some of the functions in there to comment it all, but i've added comments to `generateFunctions()` in https://github.com/netlify/next-runtime/pull/1950/commits/fd32f38d9d4e1f8c89f99770f82fb312cfab5998

packages/runtime/src/templates/handlerUtils.ts
orinokai: i'm not certain about this - there's a bunch of strings in that function that we could make into constants - but they are unlikely to change and i think it helps readability to know what they are in that context, especially since specifics like whether they have a `.` or a `/` are important with that string wrangling

sarahetter: Fair!

sarahetter (approved):
🎉  Yayy Rob!

charkour:
This is so exciting! Thank you Netlify team!

charkour:
I'm wondering how to update the [Netlify Docs](https://docs.netlify.com/integrations/frameworks/next-js/incremental-static-regeneration/#on-demand-isr) for Next.js On-Demand ISR. Any suggestions?

Issue #1953, opened 2023-03-01
modified-tests/edge-render-getserversideprops "should respond to _next/data for index correctly" test is flaky


sarahetter:
https://github.com/netlify/next-runtime/actions/runs/4305174453/jobs/7507961606

This test seems to mostly fail, but sometimes pass in CI. It always passes locally, sending `null` instead of `undefined` as the value.

ascorbic:
This is a genuine failure. It returns the wrong thing because one it's matching the wrong page. I think there's a ticket, but it's where a dynamic route is matching instead of the static one.

sarahetter:
@ascorbic is this a regression? I can't find a ticket related to this, and I'm not sure why it's not skipped.

ascorbic:
I think Rob was looking at it

sarahetter:
Added to another issue, closing

PR #1955, opened 2023-03-02
chore: disable edge-render tests

orinokai:
### Summary

Edge render tests are currently failing. There is an open issue #1953 to resolve this so disabling for now so that workflow is not affected in the meantime.


sarahetter (approved):
Thank you!

PR #1970, opened 2023-03-06
feat: add multiple set-cookie headers in middleware

sarahetter:
### Summary

Fixes an issue where [Deno Headers.set](https://deno.land/api@v1.31.1?s=Headers) overwrites the previously set header, instead of adding a new `set-cookie` header. This means that we only ever had one `set-cookie` header, which was always the most recently added cookie.

You are supposed to set multiple `set-cookie` headers, one per cookie (not a string of multiple cookies in one header). Deno's PR to add the ability to add multiple `set-cookie` headers uses the `append` method. See example in their [test files for the PR](https://github.com/denoland/deno/pull/5100/files). 

I removed @taty2010 's work to do with adding originCookies , as it turned out not to be required - the origin set-cookie headers were already present in the `response` object, but because we were doing `response.headers.set(key, value)` with a middleware `set-cookie` header, it removed all of the origin set-cookie headers and replaced them with the last-executed middleware `set-cookie` header. This PR is tested by the original tests in [Taty's original PR.](https://github.com/netlify/next-runtime/pull/1870/files)

This PR also solves an issue where you were only able to set one cookie in your middleware file, which we thought was a feature but was actually a bug 🐛 😅 

### Test plan

1. Add more than one `response.cookies.set('something', 'myvalue)` to a middleware.ts file
2. Add a cookie on a page in next.js
3. Validate that all `set-cookie` headers (multiple from middleware, one from origin) are present on the response.

Related: https://github.com/netlify/pillar-support/issues/306

### Standard checks:

- [ ] Add docs when necessary

---


taty2010 (commented):
lgtm

MarcL (approved):
Thanks for the quick debugging on this, @sarahetter! 🎉 
:shipit:

packages/runtime/src/templates/edge-shared/utils.ts
```diff
@@ -31,12 +31,11 @@ export const addMiddlewareHeaders = async (
...
-  const originCookies = response.headers.get('set-cookie')
   middlewareResponse.headers.forEach((value, key) => {
-    response.headers.set(key, value)
-    // Append origin cookies after middleware cookies
-    if (key === 'set-cookie' && originCookies) {
-      response.headers.append(key, originCookies)
+    if (key === 'set-cookie') {
+      response.headers.append(key, value)
+    } else {
+      response.headers.set(key, value)
```
MarcL, line 38: Re-reading the Deno docs, will the `.append` just set a new header anyway if it doesn't exist so we don't need the `.set` in the `else` statement? They already have logic for checking if it's a `set-cookie` header too.

sarahetter: Ah yes I thought of this, but I kept set in there as I think we may be potentially doing the `set` behaviour elsewhere (if a user is setting another type of header in their middleware) where we actually want it to work like `set` (not appending a second value to the header, as `append` will do with other types of headers)

MarcL: Ok thanks!

PR #1973, opened 2023-03-09
feat: add generator meta data for framework generated Netlify Functions

taty2010:
### Summary
https://github.com/netlify/pod-ecosystem-frameworks/issues/407

### Test plan

### Standard checks:

- [ ] Add docs when necessary

---


packages/runtime/src/helpers/functions.ts
```diff
@@ -71,10 +97,11 @@ export const generateFunctions = async (
       join(__dirname, '..', '..', 'lib', 'templates', 'handlerUtils.js'),
       join(functionsDir, functionName, 'handlerUtils.js'),
     )
+    writeGeneratorField(functionName, functionTitle, functionsDir)
   }
 
-  await writeHandler(HANDLER_FUNCTION_NAME, false)
-  await writeHandler(ODB_FUNCTION_NAME, true)
+  await writeHandler(HANDLER_FUNCTION_NAME, 'Next.js SSR handler', false)
+  await writeHandler(ODB_FUNCTION_NAME, 'Next.js ISR handler', true)
```
MarcL, line 104: [dust] Worth moving these into constants at the top of the module or in the constants file?

packages/runtime/src/helpers/functions.ts
```diff
@@ -22,6 +28,26 @@ export interface ApiRouteConfig {
   compiled: string
 }
 
+const writeGeneratorField = async (functionName: string, functionTitle: string, functionsDir: string) => {
+  const pluginPackagePath = '.netlify/plugins/package.json' || './package.json'
```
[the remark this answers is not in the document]
taty2010: The `.netlify/plugins/package.json` is where the nextjs plugin resides if the site does not have it set as a dependency so we will need to check if either exists. You've made me realize something though, the plugin package could still exist without having the nextjs plugin so I'll need to update my logic here.

test/functions.spec.ts
```diff
@@ -0,0 +1,105 @@
+import { readJSON } from 'fs-extra'
+import mock from 'mock-fs'
```
MarcL, line 2: Missing `mock-fs` from the dev dependencies in `package.json`?

packages/runtime/src/helpers/utilities/functions.ts
MarcL, line 1: [dust] Adding `utilities` under `helpers` seems like they mean the same thing. Worth renaming the module instead to something like `functionsMetadata.ts` or similar and adding it into the same directory?

packages/runtime/src/helpers/utilities/functions.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const metadata = {
+    config: {
+      name: functionTitle,
+      generator: nextPluginVersion ? `${NEXT_PLUGIN_NAME}@${nextPluginVersion}` : 'Next Runtime Version Not Found',
+    },
+    version: 1,
+  }
+
+  await writeFile(join(functionsDir, functionName, `${functionName}.json`), JSON.stringify(metadata))
```
MarcL, line 42: If we have a function `functionsDir/function.ts` then this is writing to `functionsDir/functionName/functionName.json`. Is this expected?
I think it's adding an extra directory that we don't need.

test/functions.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+        version: '1.0.0',
+        dependencies: {
+          '@netlify/plugin-nextjs': '29.3.4',
+        },
+      }),
+      'node_modules/@netlify/plugin-nextjs/package.json': JSON.stringify({
+        name: '@netlify/plugin-nextjs',
+        version: nextRuntimeVersion,
+      }),
+      '.netlify/functions/some-folder/someFunctionName/someFunctionName.json': '',
```
MarcL, line 27: This is writing a file to the mock file system. Might be better to just mock the directory?
```suggestion
      '.netlify/functions/some-folder/someFunctionName': {},
```

Also, see my comment above that I think we're adding an extra directory so we probably want this if the code is changed:

```suggestion
      '.netlify/functions/some-folder': {},
```

MarcL: Thinking about it a bit more does `fs.writeFile` write recursively? If so, do we need to mock the directory here? I would think that 'mock-fs' would allow you to write to it and then read from it. 🤔

test/functions.spec.ts
nickytonline: Yeah, just realized that. I’ll push it up in a bit.

nickytonline:
Closing in favour of #1999

PR #1980, opened 2023-03-13
fix: remove fs access which isn't available in edge functions

danez:
### Summary

Removes the `Deno.stat` call from the dev edge function.

It rethrows any errors besides the Module not found error for the `middleware.js` file. 

Because we have a number in the hash of the import it is also super easy to see if this error is from our import or not.

Fixes https://github.com/netlify/pod-ecosystem-frameworks/issues/422

### Picture of cute animal

61CNGisjWUL _SY450_

### Standard checks:

- [ ] Add docs when necessary

---


nickytonline (approved):
Thanks @danez! :shipit:

PR #1984, opened 2023-03-13
fix: use the same id for importing and checking error

danez:
The id in the import and in the catch was different. Now they are the same.


sarahetter (approved):

nickytonline (approved):
:shipit:

Issue #1983, opened 2023-03-13
[Bug]: [next-dev] TypeError: Module not found "file:///Users/lukasholzer/Sites/oepolsv/.netlify/middleware.js#0".


lukasholzer:
### Summary

When updating to the newest next runtime `"@netlify/plugin-nextjs": "^4.32.0",` and running `ntl dev` in my project I get the following console error – with `4.31.0` it's working

```
❯ ntl dev
◈ Netlify Dev ◈
◈ Ignored general context env var: LANG (defined in process)
◈ Injected .env file env var: STORYBLOK_TOKEN
◈ Injected .env file env var: STORYBLOK_PREVIEW_SECRET
◈ Injected .env file env var: IG_APP_ID
◈ Injected .env file env var: IG_SECRET
◈ Setting up local development server
Patching /Users/lukasholzer/Sites/oepolsv/node_modules/.pnpm/next@13.1.6_biqbaboplfbrettd7655fr4n2y/node_modules/next/dist/server/base-server.js
Done
Patching /Users/lukasholzer/Sites/oepolsv/node_modules/.pnpm/next@13.1.6_biqbaboplfbrettd7655fr4n2y/node_modules/next/dist/server/next-server.js
Done
◈ Starting Netlify Dev with Next.js
✔ Waiting for framework port 3000. This can be configured using the 'targetPort' property in the netlify.toml

   ┌──────────────────────────────────────────────────┐
   │                                                  │
   │   ◈ Server now ready on https://localhost:8888   │
   │                                                  │
   └──────────────────────────────────────────────────┘

◈ Loaded edge function netlify dev handler

> oepolsv@0.1.0 dev
> next dev

[next-dev] TypeError: Module not found "file:///Users/lukasholzer/Sites/oepolsv/.netlify/middleware.js#0".
    at async Object.handler [as function] (file:///Users/lukasholzer/Sites/oepolsv/.netlify/edge-functions/next-dev/index.js:38:28)
    at async FunctionChain.runFunction (https://63c7dea30898f20009dd4da8--edge.netlify.app/bootstrap/function_chain.ts:341:24)
    at async FunctionChain.run (https://63c7dea30898f20009dd4da8--edge.netlify.app/bootstrap/function_chain.ts:293:22)
    at async handleRequest (https://63c7dea30898f20009dd4da8--edge.netlify.app/bootstrap/handler.ts:63:22)
[...]

PR #1988, opened 2023-03-14
fix: make error message matching more generic

sarahetter:
### Summary

Some async business is messing with the idx value, where it's idx=z (strangely 1 more than we'd expect) when entering the catch condition, but idx=z-1 (correct value) when comparing in the if statement in the catch. 

We know that the correct values are being used when comparing, and the value of the middleware.js#x doesn't matter when the catch occurs, we just want to return silently. 

### Test plan

Fixes https://github.com/netlify/next-runtime/issues/1983
### Standard checks:

- [ ] Add docs when necessary

---


nickytonline (approved):
Looks good @sarahetter! :shipit:

PR #1993, opened 2023-03-15
test: added tests to check that the vary header contains 'RSC'

nickytonline:
### Summary

~Fixes an issue where requests coming in to the edge function that were handling React Server Components were always returning the first cached response causing issues with serving the content type `text/html` or `application/octet-stream`~

It looks like the `vary` header is already in the response, so I've added tests to ensure it remains.

Closes https://github.com/netlify/pod-ecosystem-frameworks/issues/352

### Test plan

There is an end-to-end test in this PR that tests that this works, but you can also curl the deploy preview to see it working.

1. Run `curl -v https://deploy-preview-1993--netlify-plugin-nextjs-demo.netlify.app/blog/erica/`. The content type returned is `content-type: text/html; charset=utf-8`
2. Run `curl -v -H "RSC: 1" https://deploy-preview-1993--netlify-plugin-nextjs-demo.netlify.app/blog/erica/`, The content type returned is `content-type: application/octet-stream`

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/packages/runtime/src/templates/edge-shared/rsc-data.ts
+++ b/packages/runtime/src/templates/edge-shared/rsc-data.ts
@@ -60,6 +60,12 @@ export const getRscDataRouter = ({ routes: staticRoutes, dynamicRoutes }: Preren
     const debug = request.headers.has('x-next-debug-logging')
     const log = debug ? (...args: unknown[]) => console.log(...args) : noop
     const url = new URL(request.url)
+
+    // Set the 'vary' header to 'RSC' to ensure that we cache correctly for the different
+    // possible mime types: application/octet-stream and text/html
+    // See https://github.com/netlify/pod-ecosystem-frameworks/issues/352#issuecomment-1450364417
+    request.headers.set('vary', 'RSC')
+
     // If this is a static RSC request, rewrite to the data route
     if (request.headers.get('rsc') === '1') {
       log('Is rsc request')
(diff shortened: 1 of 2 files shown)


sarahetter (approved):

taty2010 (dismissed):

packages/runtime/src/templates/edge-shared/rsc-data.ts
```diff
@@ -60,6 +60,11 @@ export const getRscDataRouter = ({ routes: staticRoutes, dynamicRoutes }: Preren
     const debug = request.headers.has('x-next-debug-logging')
     const log = debug ? (...args: unknown[]) => console.log(...args) : noop
     const url = new URL(request.url)
+
+    // Set the 'vary' header to 'RSC' to ensure that we cache correctly for the different
+    // possible content-types: application/octet-stream and text/html
+    request.headers.set('vary', 'RSC')
```
ascorbic, line 66: Shouldn't this be on the Response?

nickytonline: Yeah, you're right. Not sure why I put it on the request (late night coding 🙃 ), and also not sure why my tests and `curl`s pass. 🤔

ascorbic: It already worked. The problem isn't it returning the wrong data - that was already fixed. The problem is the browser caching the wrong data. The tests should be checking that the server is returning the `Vary` header, and you can manually test it in the browser with the instructions in https://github.com/netlify/pod-ecosystem-frameworks/issues/352

PR #2001, opened 2023-03-21
fix: use `IMAGE_FUNCTION_NAME` constant

LekoArts:
### Summary

While looking into https://github.com/netlify/pod-ecosystem-frameworks/issues/376 I searched for the usage of `IMAGE_FUNCTION_NAME` as the other constants showed all their usages. We had a couple of instances where we used `_ipx` instead of the constant and thus couldn't find it immediately.

So this is a small QoL PR to use the constant in all relevant places.


sarahetter (approved):
ty!

Issue #1629, opened 2022-09-20
[Bug]: Using `setPageProps` causes rendering mismatch when using with ampersands


ryanwalters:
### Summary

When using `setPageProp` in middleware, the response appears to become HTML encoded on the client side but not on the server, causing a `Error: Text content does not match server-rendered HTML.` 

This issue was seen on both Mac OS and Windows, Chrome, Edge, Firefox.

### Steps to reproduce

1. Pull down and run the sample repo
2. Go to http://localhost:8888

**Expected result**
A page displaying `Some plain text & an ampersand!`

**Actual result**
A page displaying `Some plain text &amp; an ampersand!` and a `Text content does not match server-rendered HTML.` error.

```
Warning: Text content did not match. Server: "Some plain text & an ampersand!" Client: "Some plain text &amp; an ampersand!"
```

Please note that commenting out `response.setPageProp` fixes the issue.

### A link to a reproduction repository

https://github.com/britannica/next-sandbox

### Next Runtime version

@netlify/next@1.3.0, @netlify/plugin-nextjs@latest (whatever is pulled down when configuring netlify.toml)

### More information about your build

- [ ] I am building using the CLI
- [ ] I am building using file-based configuration (`netlify.toml`)

### Your netlify.toml file

<details>
  <summary>`netlify.toml`</summary>

```toml
[build]
    command = "npm run build"
    publish = ".next"

[build.environment]
    NEXT_USE_NETLIFY_EDGE = "true"

[[plugins]]
    package = "@netlify/plugin-nextjs"
```

</details>

### Your public/_redirects file

<details>
  <summary>`_redirects`</summary>

```toml
# Paste content of your `_redirects` file here
```

</details>

### Your `next.config.js` file

<details>
  <summary>`next.config.js`</summary>

```toml
/** @type {import('next').NextConfig} */
const nextConfig = {
  reactStrictMode: true,
  swcMinify: true,
}

module.exports = nextConfig
```

</details>

### Builds logs (or link to your logs)

<details>
  <summary>Build logs</summary>

```
◈ Netlify Dev ◈
◈ Injected netlify.toml file env var: NEXT_USE_NETLIFY_EDGE
◈ Setting up local development server
​
────────────────────────────────────────────────────────────────
  Netlify Build
────────────────────────────────────────────────────────────────
​
> Version
  @netlify/build 27.18.3
​
> Flags
  {}
​
> Current directory
  H:\Projects\next-sandbox
​
> Config file
[...]

PR #2007, opened 2023-03-22
fix: do not escape HTML

LekoArts:
### Summary

When `setPageProp` is called it in turn calls `transformData` which pushes to the `dataTransforms` array. Inside the shared edge util it gets accessed:

https://github.com/netlify/next-runtime/blob/fb93b5469c570616134d53c10fbed1f7ef78e334/packages/runtime/src/templates/edge-shared/utils.ts#L203-L225

The HTML rewriter is a Deno port of https://developers.cloudflare.com/workers/runtime-apis/html-rewriter/.

By default it escapes HTML, but there is a `html` setting: https://developers.cloudflare.com/workers/runtime-apis/html-rewriter/#global-types

So we can just set this to `true` and fix the issue of incorrectly escaping raw HTML while rewriting the HTML.

### Test plan

See that the changed e2e test still passes.

Fixes https://github.com/netlify/next-runtime/issues/1629

### Standard checks:

---

### DIFF
--- a/demos/middleware/middleware.ts
+++ b/demos/middleware/middleware.ts
@@ -24,7 +24,7 @@ export async function middleware(req: NextRequest) {
   if (pathname.startsWith('/static')) {
     // Unlike NextResponse.next(), this actually sends the request to the origin
     const res = await request.next()
-    const message = `This was static & old but has been transformed in ${req.geo?.city}`
+    const message = `This was static (& escaping test &amp;) but has been transformed in ${req.geo?.city}`
 
     // Transform the response HTML and props
     res.replaceText('p[id=message]', message)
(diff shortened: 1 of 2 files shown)


sarahetter (approved):

demos/middleware/middleware.ts
```diff
@@ -24,7 +24,7 @@ export async function middleware(req: NextRequest) {
   if (pathname.startsWith('/static')) {
     // Unlike NextResponse.next(), this actually sends the request to the origin
     const res = await request.next()
-    const message = `This was static but has been transformed in ${req.geo?.city}`
+    const message = `This was static & old but has been transformed in ${req.geo?.city}`
```
nickytonline, line 27: These tests are great. I wonder if it makes sense to test that `&amp;` would render as the text `&amp;`? The only use case for that that I can think of is if a code example was in a prop being set. Maybe not worth it. 🤔

nickytonline (approved):
Left a suggestion, but this works great! :shipit:

demos/middleware/middleware.ts
LekoArts: Done!

PR #2018, opened 2023-03-27
fix: routes with null data routes can be filtered now

### LINKED ISSUE #1940: Bug: `appDir` applications can have `dataRoute: null`
When you run `next build` on https://github.com/redmonkez12/interactive-rating-component/tree/079f65a9b1d84efe2c1032dc01c022b46a7c0fb5, then `prerender-manifest.json` will contain

```json
"/api/hello": {
    "initialHeaders": { "content-type": "text/plain;charset=UTF-8" },
    "initialRevalidateSeconds": false,
    "srcRoute": "/api/hello",
    "dataRoute": null
  }
```

Our runtime currently can't handle that, and will throw an error here, because `dataRoute is null`:

https://github.com/netlify/next-runtime/blob/72ab895be865402fa0d92431f293129c70ae6c8b/packages/runtime/src/templates/edge-shared/rsc-data.ts#L38

To fix that, we should update the Next.js version that this repo is pulling TS types from, and fix all places where the types now fail.

nickytonline:
### Summary

Types in the project were not up to day and 

Closes #1940

### Test plan

1. Ensure the deploy previews are green.
2. There are tests that cover this fix.

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/cypress/integration/default/appdir.spec.ts
+++ b/cypress/integration/default/appdir.spec.ts
@@ -29,7 +29,7 @@ describe('appDir', () => {
     })
   })
 
-  it('returns a vary header for RSC data requests to ISR pages', () => {
+  it.skip('returns a vary header for RSC data requests to ISR pages', () => {
     cy.request({
       url: '/blog/erica/',
       followRedirect: false,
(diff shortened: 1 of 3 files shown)


orinokai (approved):
lgtm!

PR #2019, opened 2023-03-28
feat: add Generator to edge functions

taty2010:
### Summary

This PR inputs the runtime version into the new generator field within edge-functions to aid in observability.

### Test plan

### Standard checks:

- [ ] Add docs when necessary

---


packages/runtime/src/helpers/functionsMetaData.ts
```diff
@@ -15,6 +15,19 @@ const getNextRuntimeVersion = async (packageJsonPath: string, useNodeModulesPath
   return useNodeModulesPath ? packagePlugin.version : packagePlugin.dependencies[NEXT_PLUGIN]
 }
 
+const pluginPackagePath = '.netlify/plugins/package.json'
```
nickytonline, line 18: Use the constant format for constant variables to keep the codebase uniform.

```suggestion
const PLUGIN_PACKAGE_PATH = '.netlify/plugins/package.json'
```

nickytonline (changes requested):
Forgot to mark this as `Request changes`

test/functionsMetaData.spec.ts
```diff
@@ -103,3 +104,160 @@ describe('writeFunctionConfiguration', () => {
...
+      version:1
+      }
+  
+      const functionName = 'someFunction'
+      const name = 'someFunctionName'
+  
+      const pluginVersion = await writeEdgeFunctionConfiguration()
+  
+      manifest.functions.push({
+        functions: functionName,
```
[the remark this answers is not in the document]
taty2010: it's should be `function` thanks for the catch.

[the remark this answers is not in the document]
nickytonline: You can import the [FunctionManifest](https://github.com/netlify/next-runtime/blob/f69e2abe4040408039fc38b87ab85814e22a2914/packages/runtime/src/helpers/edge.ts#L52-L71) interface to catch this.

CleanShot 2023-04-03 at 16 17 00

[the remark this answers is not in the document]
nickytonline: This is making me think we should make TypeScript stricter for tests so as to catch things like this.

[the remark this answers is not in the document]
taty2010: I've added the functionManifest interface to the edge tests. I definitely think it's beneficial and I would have caught that little mistake beforehand.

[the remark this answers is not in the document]
orinokai: yes, I'd be happy with stricter TypeScript for things like this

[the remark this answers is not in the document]
nickytonline: Done. https://github.com/netlify/next-runtime/issues/2028

packages/runtime/src/helpers/functionsMetaData.ts
```diff
@@ -34,23 +47,20 @@ export interface FunctionInfo {
...
-      generator: `${NEXT_PLUGIN_NAME}@${nextPluginVersion}`,
+      generator: `${NEXT_PLUGIN_NAME}@${pluginVersion}`,
     },
     version: 1,
   }
 
   await writeFile(join(functionsDir, functionName, `${functionName}.json`), JSON.stringify(metadata))
 }
+
+export const writeEdgeFunctionConfiguration = async () => {
```
MarcL, line 63: This isn't writing anything so can we rename it to match its purpose?

packages/runtime/src/helpers/functionsMetaData.ts
```diff
@@ -34,23 +47,20 @@ export interface FunctionInfo {
...
-    (await getNextRuntimeVersion(pluginPackagePath, false)) ||
-    // The runtime version should always be available, but if it's not, return 'unknown'
-    'unknown'
+  const pluginVersion = await nextPluginVersion()
 
   const metadata = {
     config: {
       name: functionTitle,
-      generator: `${NEXT_PLUGIN_NAME}@${nextPluginVersion}`,
+      generator: `${NEXT_PLUGIN_NAME}@${pluginVersion}`,
```
MarcL, line 55: Could this call the function you've added below instead of a string template as that already generates the generator name with the plugin version?

test/functionsMetaData.spec.ts
```diff
@@ -103,3 +105,176 @@ describe('writeFunctionConfiguration', () => {
     expect(actual).toEqual(expected)
   })
 })
+
+describe('writeEdgeFunctionConfiguration', () => {
+  afterEach(() => {
+    mock.restore()
+  })
+
+  it('should write the configuration for an edge function using node modules version of @netlify/plugin-nextjs', async () => {
```
MarcL, line 114: These tests are testing more than `writeEdgeFunctionConfiguration` (as the function doesn't write anything). How about splitting up the tests?

taty2010: Hey, so I ended up removing the tests you referenced based on our convo today. Since we are now using the `getPluginVersion` within the `writeFunctionsConfiguration` the original tests that Nick worked on should be enough to check and make sure it is working as expected.

packages/runtime/src/helpers/functionsMetaData.ts
taty2010: This has been done, thanks Marc!

test/functionsMetaData.spec.ts
MarcL: Thanks! I won't stop this PR from merging but I think it would be good to get some tests around the new generator code that's been added (which uses the plugin version now). This might be more work to get them set up as none exist right now for the edge functions so maybe raise an issue to add them?

nickytonline: Since we're adding this code to generate meta data for edge functions, I think it'd be good to add the tests in this PR. We have our 1:1 today @taty2010, so we can take a peek at this then.

MarcL: That would be great. Thanks for helping @nickytonline.

taty2010:
Created new tests to check for the generator field within the manifest. Currently, all tests are passing locally but a few of them are failing within this PR and will need to look into why.

packages/runtime/src/helpers/edge.ts
```diff
@@ -300,11 +305,13 @@ export const writeRscDataEdgeFunction = async ({
       function: 'rsc-data',
       name: 'RSC data routing',
       path,
+      generator: `${nextjsPluginVersion}`,
```
MarcL, line 308: This doesn't need to be a template string now and you can just use the variable.

```suggestion
      generator: nextjsPluginVersion`,
```

packages/runtime/src/helpers/edge.ts
```diff
@@ -476,6 +489,7 @@ export const writeEdgeFunctions = async ({
       function: 'ipx',
       name: 'next/image handler',
       path: nextConfig.images.path || '/_next/image',
+      generator: `${nextjsPluginVersion}`,
```
MarcL, line 492: This doesn't need to be a template string now and you can just use the variable.

```suggestion
      generator: nextjsPluginVersion`,
```

packages/runtime/src/helpers/edge.ts
```diff
@@ -442,6 +453,7 @@ export const writeEdgeFunctions = async ({
         pattern,
         // cache: "manual" is currently experimental, so we restrict it to sites that use experimental appDir
         cache: usesAppDir ? 'manual' : undefined,
+        generator: `${nextjsPluginVersion}`,
```
MarcL, line 456: This doesn't need to be a template string now and you can just use the variable.

```suggestion
      generator: nextjsPluginVersion`,
```

packages/runtime/src/helpers/edge.ts
```diff
@@ -451,6 +463,7 @@ export const writeEdgeFunctions = async ({
           name: edgeFunctionDefinition.name,
           pattern: dataRoute,
           cache: usesAppDir ? 'manual' : undefined,
+          generator: `${nextjsPluginVersion}`,
```
MarcL, line 466: This doesn't need to be a template string now and you can just use the variable.

```suggestion
      generator: nextjsPluginVersion`,
```

taty2010:
Ended up having to sort the manifest data so that the snapshots would match.

packages/runtime/src/helpers/edge.ts
```diff
@@ -57,12 +57,14 @@ export interface FunctionManifest {
         name?: string
         path: string
         cache?: 'manual'
+        generator?: string
```
LekoArts, line 60: Isn't `generator` always defined?

LekoArts (changes requested):
Some smaller comments, not blocking

packages/runtime/src/helpers/edge.ts
taty2010: Yep generator will always be defined, thanks for spotting this.

test/index.spec.js
```diff
@@ -730,6 +738,32 @@ describe('onBuild()', () => {
...
+    expect(existsSync(path.join('.netlify', 'edge-functions', 'manifest.json'))).toBeTruthy()
+  })
+
+  test('generates generator field within the edge-functions manifest', async () => {
+    await moveNextDist()
+    await nextRuntime.onBuild(defaultArgs)
+    const manifestPath = await readJson(path.resolve('.netlify/edge-functions/manifest.json'))
+    const manifest = manifestPath.functions.sort()
+    
+    expect(manifest).toMatchSnapshot()
```
LekoArts, line 753: I think we're overusing snapshots in most of our tests, for me it's typically the last resort.

I'd use this instead:

```ts
expect(manifest).toEqual(
  expect.arrayContaining([
    expect.objectContaining({
      generator: '@netlify/next-runtime@1.0.0'
    })
  ])
)
```

Applies also to the test below.

You then can also drop the `sort()` as the order is not relevant anymore.

Some reading: https://www.emgoto.com/jest-partial-match/

LekoArts (changes requested):
Some more snapshots need an update: https://github.com/netlify/next-runtime/actions/runs/4725240207/jobs/8383749771?pr=2019#step:6:3932

test/index.spec.js
taty2010: This is really helpful, thanks Lennart!!

test/index.spec.js
```diff
@@ -730,6 +738,44 @@ describe('onBuild()', () => {
...
+    await moveNextDist()
+    await nextRuntime.onBuild(defaultArgs)
+    expect(existsSync(path.join('.netlify', 'edge-functions', 'manifest.json'))).toBeTruthy()
+  })
+
+  test('generates generator field within the edge-functions manifest', async () => {
+    await moveNextDist()
+    await nextRuntime.onBuild(defaultArgs)
+    const manifestPath = await readJson(path.resolve('.netlify/edge-functions/manifest.json'))
+    const manifest = manifestPath.functions.sort()
```
nickytonline, line 751: if the functions are sorted now, can we not use snapshots?

taty2010: The sort should have been deleted along with the snapshot, I will make that update.

LekoArts (approved):

nickytonline (approved):
:shipit:

PR #2030, opened 2023-04-05
fix: workaround netlify-cli not matching prod

pieh:
### Summary

On latest `netlify-cli` (`13.2.2`) running `netlify serve` fails with

```
Cannot read properties of undefined (reading 'odb_refresh_hooks')
```

My assumption is that we are not getting custom context in event handlers when running locally, so this just allows to run it locally for now

### Test plan

### Standard checks:

- [ ] Add docs when necessary

---


nickytonline:
Do refresh hooks work locally @orinokai or is it like ISR pages with `netlify serve`?

pieh:
> Do refresh hooks work locally @orinokai or is it like ISR pages with `netlify serve`?

From my manual local tests it looks like ODB doesn't act as ODB with `netlify serve` and rather as regular function locally, so it is always re-rendered so that alone will mean refresh hook wouldn't have any effect even if it didn't throw somewhere along the way.

Also refresh is done via hitting Netlify API and specifying hostname which would be just `localhost:8888` so Netlify platform wouldn't even know what to do about that (?)

orinokai:
Thanks for this @pieh, and yes you're right, ISR currently doesn't work locally (I'd like this to work, but we'd have to implement the ODB cache locally and ensure it has feature parity with production) so this fix is a good workaround for now.

LekoArts (approved):

PR #2041, opened 2023-04-13
fix: now data routes for dynamic routes filter even when null when creating the RSC data router

### LINKED ISSUE #1940: Bug: `appDir` applications can have `dataRoute: null`

nickytonline:
### Summary

There was a TypeScript error that was not picked up as we copy them for the build. I've added some tests to cover this like we did in #2018.

Closes #1940

### Test plan

1. Ensure the deploy previews are green.
2. There are tests that cover this fix.

### Standard checks:

- [ ] Add docs when necessary

---

someone petting an alapaca

### DIFF
--- a/packages/runtime/src/templates/edge-shared/rsc-data.ts
+++ b/packages/runtime/src/templates/edge-shared/rsc-data.ts
@@ -44,7 +44,7 @@ export const getRscDataRouter = ({ routes: staticRoutes, dynamicRoutes }: Preren
   )
 
   const dynamicRouteMatcher = Object.values(dynamicRoutes)
-    .filter(({ dataRoute }) => dataRoute.endsWith('.rsc'))
+    .filter(({ dataRoute }) => dataRoute?.endsWith('.rsc'))
     .map(({ routeRegex }) => new RegExp(routeRegex))
 
   const matchesDynamicRscDataRoute = (pathname: string) => {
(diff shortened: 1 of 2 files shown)


packages/runtime/src/templates/edge-shared/rsc-data.ts
```diff
@@ -44,7 +44,7 @@ export const getRscDataRouter = ({ routes: staticRoutes, dynamicRoutes }: Preren
   )
 
   const dynamicRouteMatcher = Object.values(dynamicRoutes)
-    .filter(({ dataRoute }) => dataRoute.endsWith('.rsc'))
+    .filter(({ dataRoute }) => dataRoute?.endsWith('.rsc'))
```
nickytonline, line 47: As mentioned in the description of the PR this wasn't being caught during compilation of the Next runtime as this file is copied.

sarahetter (approved):

PR #2044, opened 2023-04-14
fix: ensure appDir routes with null dataRoutes are not added to edge manifest for rsc-data

orinokai:
### Summary

In #1940 and #2041 we fixed an error that was occurring because the `rsc-data` edge function was unable to handle `null` `dataRoutes` in the `prerender-manifest.json`. However, the problem of null data routes also affected our own `.netlify/edge-functions/manifest.json` because the edge functions bundler always [expects a string](https://github.com/netlify/edge-bundler/blob/c91593f5e7f0831e4cfe3c69d240d6044c8869c8/node/manifest.ts#L171).

This PR adds a condition so that appDir routes with falsy dataRoutes are not added to the edge manifest (and thus will not invoke the `rsc-data` edge function, which is the expected result).

### Test plan

Tests added for the `writeRscDataEdgeFunction`

### DIFF
--- a/test/edge.spec.ts
+++ b/test/edge.spec.ts
@@ -1,0 +1,118 @@
+import { writeRscDataEdgeFunction } from '../packages/runtime/src/helpers/edge'
+import type { PrerenderManifest } from 'next/dist/build'
+
+const basePrerenderManifest: PrerenderManifest = {
+  version: 3,
+  routes: {},
+  dynamicRoutes: {},
+  notFoundRoutes: [],
+  preview: {
+    previewModeId: '',
+    previewModeSigningKey: '',
+    previewModeEncryptionKey: '',
+  },
+}
+
+describe('writeRscDataEdgeFunction', () => {
+  it('should return manifest entries for static appDir routes', async () => {
+    const prerenderManifest: PrerenderManifest = {
+      ...basePrerenderManifest,
+      routes: {
+        '/': {
+          initialRevalidateSeconds: false,
+          srcRoute: '/',
+          dataRoute: '/index.rsc',
+        },
+      },
+    }
+    const appPathRoutesManifest = {
+      '/page': '/',
+    }
+    const edgeManifest = await writeRscDataEdgeFunction({ prerenderManifest, appPathRoutesManifest })
+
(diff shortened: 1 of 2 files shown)


test/edge.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  dynamicRoutes: {},
+  notFoundRoutes: [],
+  preview: {
+    previewModeId: '',
+    previewModeSigningKey: '',
+    previewModeEncryptionKey: '',
+  },
+}
+
+describe('writeRscDataEdgeFunction', () => {
```
sarahetter, line 16: Sorry github won't let me comment in the other file as you didn't edit this part. Can we rename this function (`writeRscDataEdgeFunction`)? It doesn't actually doing the "write" in the function, it just returns an array.

test/edge.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    ])
+  })
+
+  it('should not return manifest entries for dynamic appDir routes without dataRoutes', async () => {
+    const prerenderManifest: PrerenderManifest = {
+      ...basePrerenderManifest,
+      dynamicRoutes: {
+        '/api/[endpoint]': {
+          routeRegex: '^/api/([^/]+?)(?:/)?$',
+          dataRoute: '/api/[endpoint].rsc',
```
sarahetter, line 104: Confused here, as it looks like there is a dataRoute?

pieh: `dynamicRoutes` (at least in our code) use `dataRouteRegex` field, not the `dataRoute`

test/edge.spec.ts
orinokai: sure thing, fixed in https://github.com/netlify/next-runtime/pull/2044/commits/1141f4ff6b2296f2d9fb85056beed29eb10cfe7e

test/edge.spec.ts
orinokai: thanks folks, i've updated the test name to make it a bit clearer https://github.com/netlify/next-runtime/pull/2044/commits/b82d3d98c994fe10c787010d8ff00acf8a5326d0

PR #2057, opened 2023-04-20
fix: make large lambda warning message more informative

biruwon:
### Summary

Currently there is a misleading about function's size larger than 50MB that will fail the deploy. This PR improves it, as the hard limit it is 250MB unzipped and we will display a warning for zipped files bigger than 50MB

### Test plan

### Standard checks:

- [ ] Add docs when necessary

---


biruwon:
@pieh not sure about the tests failing, they don't seem related to this change

packages/runtime/src/helpers/verification.ts
```diff
@@ -105,7 +105,11 @@ export const checkForRootPublish = ({
   }
 }
 
-export const checkZipSize = async (file: string, maxSize: number = LAMBDA_MAX_SIZE): Promise<void> => {
+export const checkZipSize = async (
```
nickytonline, line 108: Can we add a couple tests for this @biruwon? One for not calling console.log if `fileSize < warningSize` and one for checking that console.log is indeed called with the message or checking for a partial message if that's easier?

Currently we only have a test to check if the bundle size is disabled. Also, I can add those tests if you want.

See test/helpers/verification.spec.ts

pieh: @biruwon I pushed test fixes - it was mix of things:
 - spying correct `console.X` method
 - clearing mocks so previous tests don't impact future tests
 - resetting mocks so next `describe` blocks that didn't want mocks are not borked
 - disabling chalk's coloring for testing as I couldn't get it to match any other way

LekoArts:
@biruwon Please fill out the PR description so that we know in the future what we did and why :)

packages/runtime/src/helpers/verification.ts
biruwon: thanks @pieh for that! as I can't not approve my own PRs, can you approve it and move it forward? Like the thing you did for gastby

biruwon:
> @biruwon Please fill out the PR description so that we know in the future what we did and why :)

done

Issue #2024, opened 2023-04-02
Build Error - Plugin "@netlify/plugin-nextjs" internal error  


joeluna17:
Pushed new code and during deployment I'm receiving and error from the plugin. This worked last for me on 3/18/2023. I've tried deselecting the plugin from runtime, and then reselecting then recaching and redeploying with no luck. Below is the error output.

4:02:29 PM: build-image version: 6c82004ef0368b3a0232c6837cead23188bb347e (focal)
4:02:29 PM: buildbot version: 0dfbdb6e6e614fd26450cfc6bacb47c4c6fd9626
4:02:29 PM: Building without cache
4:02:29 PM: Starting to prepare the repo for build
4:02:29 PM: No cached dependencies found. Cloning fresh repo
4:02:29 PM: git clone --filter=blob:none git@gitlab.com:joelunaGetCheckApp/check-forms-app
4:02:30 PM: Preparing Git Reference refs/heads/main
4:02:32 PM: Parsing package.json dependencies
4:02:33 PM: Different build dir detected, going to use the one specified in the Netlify configuration file: 'apps' versus 'apps/' in the Netlify UI
4:02:33 PM: Different publish path detected, going to use the one specified in the Netlify configuration file: 'dist/apps/check-payments-marketing/.next' versus '../dist/apps/check-payments-marketing/.next' in the Netlify UI
4:02:33 PM: Starting to install dependencies
4:02:33 PM: Python version set to 3.8
4:02:33 PM: Started restoring cached go cache
4:02:33 PM: Finished restoring cached go cache
4:02:33 PM: Installing Go version 1.19.5 (requested 1.19.5)
4:02:38 PM: go version go1.19.5 linux/amd64
4:02:38 PM: Using PHP version 8.0
4:02:38 PM: v16.20.0 is already installed.
4:02:38 PM: Now using node v16.20.0 (npm v8.19.4)
4:02:39 PM: Enabling Node.js Corepack
4:02:39 PM: Started restoring cached build plugins
4:02:39 PM: Finished restoring cached build plugins
4:02:39 PM: Attempting Ruby version 2.7.2, read from environment
4:02:39 PM: Using Ruby version 2.7.2
4:02:40 PM: Install dependencies script success
4:02:40 PM: Starting build script
4:02:40 PM: Using stage install dependencies
4:02:41 PM: Detected 2 framework(s)
4:02:41 PM: "next-nx" at version "unknown"
[...]

PR #2059, opened 2023-04-21
fix: resolve next-server from next app directory and not from plugin

pieh:
### Summary

This PR adds functionality to resolve `next/dist/server/next-server` from the Next app directory (not to be confused with `appDir` inside pages), and not from the location of the `next-runtime`.

### Test plan

A new integration test was added.

Fixes https://github.com/netlify/next-runtime/issues/2024

### Standard checks:

- [ ] Add docs when necessary

---


packages/runtime/src/templates/getHandler.ts
```diff
@@ -86,7 +102,7 @@ const makeHandler = (conf: NextConfig, app, pageRoot, staticManifest: Array<[str
         port,
       },
       {
-        revalidateToken: customContext.odb_refresh_hooks,
+        revalidateToken: customContext?.odb_refresh_hooks,
```
pieh, line 105: not part of actual fix - just for local debugging (probably will remove it as I will be finishing this PR)

packages/runtime/src/helpers/functions.ts
```diff
@@ -41,6 +41,11 @@ export const generateFunctions = async (
   const functionDir = join(functionsDir, HANDLER_FUNCTION_NAME)
   const publishDir = relative(functionDir, publish)
 
+  const nextServerModuleAbsoluteLocation = getNextServerModulePath(appDir)
+  const nextServerModuleRelativeLocation = nextServerModuleAbsoluteLocation
+    ? relative(functionDir, nextServerModuleAbsoluteLocation)
+    : undefined
+
```
pieh, line 48: if `nextServerModuleAbsoluteLocation` is falsy we could warn here that non-SSG routes will fail (tho that might be noisy, so if we want warning like that maybe we add one once we know if we need lambdas for routes)

packages/runtime/src/templates/getHandler.ts
```diff
@@ -177,15 +190,25 @@ const makeHandler = (conf: NextConfig, app, pageRoot, staticManifest: Array<[str
...
+export const getHandler = ({
+  isODB = false,
+  publishDir = '../../../.next',
+  appDir = '../../..',
+  nextServerModuleRelativeLocation,
+}): string =>
   // This is a string, but if you have the right editor plugin it should format as js
   javascript/* javascript */ `
+  if (!${JSON.stringify(nextServerModuleRelativeLocation)}) {
+    throw new Error('Could not find Next.js server')
```
pieh, line 202: In current state of this PR, this is thrown on lambda invocation and not during the build, so it restores the that behavior prior to https://github.com/netlify/next-runtime/pull/1950 on when the error is thrown - not the hill I will die on and I can move it back to build time.

The other changes in this PR make it more likely to find `next-server` than before, but I don't have 100% confidence in it, so that's why I move this error throwing back here as that was behavior for a long time.

Without that move, builds for fully static sites that hit current error are just blocked until we implement logic to determine wheter we need page route handlers at all and skip it altogether if not needed

packages/runtime/src/helpers/files.ts
```diff
@@ -340,6 +340,54 @@ const getServerFile = (root: string, includeBase = true) => {
...
 }
 
+/**
+ * Try to find next-server module in few locations (to support different next versions) and in few context (try to resolve from app location and from this module)
+ */
+export const getNextServerModulePath = (root: string): string | null => {
+  // first let's try to use app location directory to find next-server
+  try {
+    const nextServerModuleLocation = getServerFile(root, false)
+    if (nextServerModuleLocation) {
```
nickytonline, line 350: For the guards we have here, would there ever be a case where `nextServerModuleLocation` would be falsy? Looking at the code, unless it throws because the module does not exist, we should always be get a value set in `nextServerModuleLocation` in all the try/catch if blocks here? Or maybe I'm missing a piece.

pieh: `getServerFile` and more specifically `findModuleFromBase` that is called there

https://github.com/netlify/next-runtime/blob/44dac405ca887fa6b6b3e14f831c376a68d102ed/packages/runtime/src/helpers/utils.ts#L257-L272

doesn't throw - instead it returns falsy `null` if it can't find a module

packages/runtime/src/templates/getHandler.ts
pieh: This was merged separately ( https://github.com/netlify/next-runtime/pull/2030 )

packages/runtime/src/templates/getApiHandler.ts
```diff
@@ -17,16 +17,16 @@ const { URLSearchParams, URL } = require('url')
...
 
 type Mutable<T> = {
   -readonly [K in keyof T]: T[K]
 }
 
 // We return a function and then call `toString()` on it to serialise it as the launcher function
-
-const makeHandler = (conf: NextConfig, app, pageRoot, page) => {
+// eslint-disable-next-line max-params
+const makeHandler = (conf: NextConfig, app, pageRoot, page, NextServer: NextServerType) => {
```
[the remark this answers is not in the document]
pieh: https://github.com/netlify/next-runtime/pull/2059/commits/610d73f7455cb6c19f639d6736711be26559b6db

packages/runtime/src/helpers/files.ts
pieh: This was refactored (that function was actually removed) since initial commit so I'll resolve this, but general sentiment is still that a lot of those helper functions don't throw and instead return falsy `null` if module can't be resolved

packages/runtime/src/helpers/analysis.ts
```diff
@@ -87,18 +83,25 @@ let hasWarnedAboutNextVersion = false
...
   }
 
   try {
     if (!extractConstValue) {
-      extractConstValue = require('next/dist/build/analysis/extract-const-value')
+      // eslint-disable-next-line import/no-dynamic-require
+      extractConstValue = require(findModuleFromBase({
+        paths: [appDir],
+        candidates: ['next/dist/build/analysis/extract-const-value'],
+      }) ?? 'next/dist/build/analysis/extract-const-value')
```
pieh, line 97: This is a bit messy - especially part after `??` - it's done because `findModuleFromBase` can return `null`, and calling `require(null)` won't result in `MODULE_NOT_FOUND` error that we handle below - so the reason for `??` part is to actually produce above error in case module can't be resolved

packages/runtime/src/helpers/types.ts
```diff
@@ -51,3 +51,10 @@ export interface RoutesManifest {
   i18n: I18n
   rewrites: Rewrites
 }
+
+// I have no idea what eslint is up to here but it gives an error
+// eslint-disable-next-line no-shadow
+export const enum ApiRouteType {
+  SCHEDULED = 'experimental-scheduled',
+  BACKGROUND = 'experimental-background',
+}
```
pieh, line 60: This (including comment) was moved from another module due to cyclic imports problem

packages/runtime/src/helpers/utils.ts
```diff
@@ -268,6 +267,17 @@ export const findModuleFromBase = ({ paths, candidates }): string | null => {
...
+  for (const candidate of candidates) {
+    try {
+      const modulePath = require.resolve(candidate)
+      if (modulePath) {
+        return modulePath
+      }
+    } catch {
+      // Ignore the error
+    }
+  }
```
orinokai, line 280: [sand] instead of an additional for loop, consider pushing the current directory `'.'` to the `paths` array

packages/runtime/src/helpers/analysis.ts
orinokai: [sand] you could potentially move the try/catch from `findModuleFromBase` to `getServerFile` (which is the only place that `findModuleFromBase` is used) and then you don't have to the juggling with `??`

packages/runtime/src/templates/server.ts
```diff
@@ -3,102 +3,105 @@ import { NodeRequestHandler, Options } from 'next/dist/server/next-server'
...
-const NextServer: NextServerType = getNextServer()
-
 interface NetlifyConfig {
   revalidateToken?: string
 }
 
-class NetlifyNextServer extends NextServer {
-  private netlifyConfig: NetlifyConfig
-  private netlifyPrerenderManifest: PrerenderManifest
+const getNetlifyNextServer = (NextServer: NextServerType) => {
```
orinokai, line 17: [dust] consider declaring `getNetlifyNextServer` separately rather than enclosing the class, but it's just a style thing so no worries if you disagree!

packages/runtime/src/templates/getHandler.ts
orinokai: Yeah, i think that makes sense tbh. We only need the server at runtime so let's go with this.

orinokai (approved):
Some minor suggestions, but take them or leave them. Otherwise looks great, thanks @pieh !

packages/runtime/src/templates/server.ts
pieh: I'm not sure I understand:

I think I do need to enclose the class inside a function because to declare `NetlifyNextServer` class I need a reference to `NextServer` class. And I can't have actual guaranteed working `NextServer` class with either static import or with executing some code in module scope (as both eveluate eagerly, before we even have access to `app` directory which is needed to reliably get a hold of `next-server` module in `next`).

Maybe it would be doable with some prototype shenanigans to declare `NetlifyNextServer` class that doesn't extend `NextServer` initially and apply needed stuff once we get hold of `NextServer`, but I think this would actually be worse and more confusing than current approach?

.github/workflows/test-integration.yml
```diff
@@ -0,0 +1,29 @@
+name: Next Runtime Integration Tests
```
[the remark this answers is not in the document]
pieh: Yeah, I think it make sense, but at this point I'd rather ship those as-is and do renaming in separate standalone PR to not keep the fix hostage

packages/runtime/src/helpers/analysis.ts
pieh: I decided to do https://github.com/netlify/next-runtime/pull/2059/commits/04f3164c275d50ac12f5958936b9ffe2f4c46d27 instead - I looked at how throwing would look like in `findModuleFromBase` and realized that removing `try/catch` there would result in throwing in first missed candidate, and to make it somewhat work it would be quite messy with storing exceptions and rethrowing them when we completed iterating over candidates and still didn't find one.

Also thinking more - this would change semantics of `findModuleFromBase` which is not end of the world, but I prefer to avoid changing shared utils behavior if it can be addressed in single place that would benefit from it (and possibly cause problems in all the other places)

packages/runtime/src/helpers/utils.ts
pieh: I tried, and it's actually not that simple. Using just `.`, `__dirname` etc in `paths` alone doesn't replicate behavior of `require.resolve` (when using default `paths`). We can get what default would use by calling `require.resolve.paths(candidate)` and we can push all that to our custom `paths` sure, but I never actually use that function before in production and I'm not sure of its sharp edges so would prefer to avoid going that route.

Just for some more details on that - here's example what default paths look like generally - it first starts with directory of current module + `/node_modules` and then go up directory tree, later seems to be checking some globals (??) etc:
```
> pwd
/Users/misiek/dev/next-runtime
> node -e "console.log(require.resolve.paths('next'))"
[
  '/Users/misiek/dev/next-runtime/node_modules',
  '/Users/misiek/dev/node_modules',
  '/Users/misiek/node_modules',
  '/Users/node_modules',
  '/node_modules',
  '/Users/misiek/.node_modules',
  '/Users/misiek/.node_libraries',
  '/Users/misiek/.nvm/versions/node/v18.14.0/lib/node'
]
```

So TL/DR it's not as simple as `.` and it might potentially introduce unknown problems that I would prefer to avoid

orinokai: Thanks for the explanation @pieh, got it 👍🏼

nickytonline (commented):
Not too much to say. This is looking good.

packages/runtime/src/helpers/types.ts
pieh:
```
// I have no idea what eslint is up to here but it gives an error
// eslint-disable-next-line no-shadow
```

This is needed to be moved as well and not just type - same error happen in current location when you remove comments

test/integration/fixtures/next-package-not-resolvable-in-base-dir/clear-node-modules-after-functions-bundling-plugin/index.js
```diff
@@ -0,0 +1,12 @@
+const fs = require(`fs`)
+const path = require(`path`)
+
+module.exports = {
+  onPostBuild: async () => {
+    const movedDir = path.join(process.cwd(), `next-app`, `node_modules2`)
```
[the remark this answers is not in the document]
pieh: Ah sorry - this is not pulled in test - it's just new one that uses `netlify serve` and this "build plugin" is meant to get rid of `node_modules` after functions were bundled but before they started to be served. However when I just deleted `node_modules` it was problematic, sometimes getting ENOENT errors so instead of deleting it I just move `node_modules` to `node_modules2` so it no longer is used. The reason to check if `node_modules2` exist and deleting it if it does is because running the same test twice in a row would crash otherwise.

And the reason to remove `node_modules` are just functions bundling quirks I encountered:

I did use absolute path to `next-server` (so something like `/Users/misiek/dev/repo/next-app/node_modules/next/<path-to-next-server>`) when generating lambda handler at some point and it worked locally, but not when deployed, because functions bundler doesn't seem to handle absolute paths correctly and left those imports as they are instead of bundling it. Things appeared to be working locally because those `node_modules` still existed.

packages/runtime/src/helpers/types.ts
pieh: https://github.com/typescript-eslint/typescript-eslint/issues/2483 is the reason - `enum` stuff is problematic with non typescript `no-shadow` rule - I can fix this either here or separately

pieh: Fixed in https://github.com/netlify/next-runtime/pull/2059/commits/66a9670e413aca7959efcdc65901721981a7ee64

LekoArts (approved):

PR #2061, opened 2023-04-21
chore: Update to cypress v12

taty2010:
### Summary

This PR updates our tests from cypress v9 to v12 [Migration Docs ](https://docs.cypress.io/guides/references/migration-guide)
- Updated the cypress-io/github-action to v5
- cy:open now defaults to Chrome and E2E
- 
closes https://github.com/netlify/pod-ecosystem-frameworks/issues/333

### Test plan

### Standard checks:

- [ ] Add docs when necessary

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "@babel/preset-env": "^7.15.8",
     "@babel/preset-typescript": "^7.16.0",
     "@delucis/if-env": "^1.1.2",
-    "@netlify/build": "^29.9.2",
+    "@netlify/build": "^29.10.0",
     "@netlify/eslint-config-node": "^7.0.1",
     "@testing-library/cypress": "^9.0.0",
     "@types/fs-extra": "^9.0.13",
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,4 +1,4 @@
 {
-  "packages/runtime": "4.35.0",
+  "packages/runtime": "4.36.0",
   "packages/next": "1.4.5"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,7 @@
         "@babel/preset-env": "^7.15.8",
         "@babel/preset-typescript": "^7.16.0",
         "@delucis/if-env": "^1.1.2",
-        "@netlify/build": "^29.9.2",
+        "@netlify/build": "^29.10.0",
         "@netlify/eslint-config-node": "^7.0.1",
         "@testing-library/cypress": "^9.0.0",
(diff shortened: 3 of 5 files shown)


sarahetter (approved):

.github/workflows/cypress-demo.yml
```diff
@@ -62,9 +62,9 @@ jobs:
           headless: true
           record: true
           parallel: true
-          config-file: cypress/config/ci.json
+          config-file: cypress/config/ci.config.ts
```
nickytonline, line 65: Cypress config is now a JS or TS file. See https://docs.cypress.io/guides/references/configuration#Configuration-File

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Run Next.js seamlessly on Netlify",
   "scripts": {
     "build:demo": "cd demos/default && npm run build",
-    "cy:open": "cypress open --config-file cypress/config/all.json",
+    "cy:open": "cypress open --config-file=cypress/config/all.config.ts",
```
LekoArts, line 7: Since Cypress now also has component testing, `cypress open` will put one into a screen where you have to choose e2e and your browser first.

You can short-circuit it like this:

```suggestion
    "cy:open": "cypress open --config-file=cypress/config/all.config.ts --browser chrome --e2e",
```

.github/workflows/cypress-canary.yml
LekoArts, line 1: If we're upgrading Cypress, I'd propose to also upgrade its GitHub action. We seem to currently use v2, while v5 is the latest: https://github.com/cypress-io/github-action/releases

taty2010: Glad you pointed that out looks like V2 was going to stop working after May 

> [GitHub announced](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) their plan to disable save-state and set-output commands by May 31, 2023. This will prevent [cypress-io/github-action](https://github.com/cypress-io/github-action) version [v4.2.1](https://github.com/cypress-io/github-action/releases/tag/v4.2.1), and earlier, running after this date since they use set-output. Affected users should update to using v5 of the [cypress-io/github-action](https://github.com/cypress-io/github-action) action before the deadline.

danez:
Just FYI: This PR made the integration tests in `zip-it-and-ship-it` fail. Not critical or problematic at all, but maybe we can add a comment somewhere here, which mentions that `zip-it-and-ship-it` needs to be updated too: https://github.com/netlify/zip-it-and-ship-it/pull/1412

PR #2074, opened 2023-04-27
Fix: Legacy fallback env var

taty2010:
### Summary

This PR allows users to opt-in to the legacy behavior for dynamic routes set to `fallback:false` by adding the env var `LEGACY_FALLBACK_FALSE='true'`. When hitting a non-existent path on a dynamic route, it will hit the origin server which will give users the ability to add redirects for those paths. 

closes https://github.com/netlify/pod-ecosystem-frameworks/issues/446

### Test plan

1. add env var `LEGACY_FALLBACK_FALSE='true'`
2. ntl serve
3. When checking Network tab within dev tools you should see header `x-nf-render-mode:odb`

### Standard checks:

- [ ] Add docs when necessary

---


test/helpers/utils.spec.ts
```diff
@@ -131,4 +142,121 @@ describe('redirectsForNext404Route', () => {
...
+        "page": "/getStaticProps/[id]",
+        "regex": "^/getStaticProps/([^/]+?)(?:/)?$",
+        "routeKeys": {
+            "nextParamid": "nextParamid"
+        },
+        "namedRegex": "^/getStaticProps/(?<nextParamid>[^/]+?)(?:/)?$"
+    },
+    ]
+
+    const middleware = await getMiddleware(path.resolve('.next'))
```
nickytonline, line 170: Does the middleware affect the tests in any way? If so consider tests with and without it.

test/helpers/utils.spec.ts
```diff
@@ -131,4 +142,121 @@ describe('redirectsForNext404Route', () => {
       { force: false, from: '/fr/test', status: 404, to: '/server/pages/fr/404.html' },
     ])
   })
+
+  it.only('returns static 404 redirects when LEGACY_FALLBACK_FALSE is not set', async () => {
```
nickytonline, line 146: remove the `.only`s from tests.

packages/runtime/src/helpers/redirects.ts
```diff
@@ -238,7 +238,11 @@ const generateDynamicRewrites = ({
             withData: true,
           }),
         )
-      } else if (prerenderedDynamicRoutes[route.page].fallback === false && !is404Isr) {
+      } else if (
+        prerenderedDynamicRoutes[route.page].fallback === false &&
+        !is404Isr &&
+        !process.env.LEGACY_FALLBACK_FALSE
```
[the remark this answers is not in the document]
taty2010: Thanks for the heads up Rob and the ref code.

test/helpers/utils.spec.ts
taty2010: I think if we wanted to test the matchesMiddleware functionality then it would matter. For these tests, we can just set the middleware to default to an empty array so that we can reach the logic we need to test related to the env var.

test/helpers/utils.spec.ts
```diff
@@ -132,4 +174,75 @@ describe('redirectsForNext404Route', () => {
...
+          from: '/getStaticProps/:id',
+          status: 404,
+          to: '/server/pages/404.html',
+        },
+      ],
+      dynamicRoutesThatMatchMiddleware: [],
+    }
+
+    expect(generateDynamicRewrites(route)).toStrictEqual(expected)
+  })
```
nickytonline, line 247: This test is not necessary.

```suggestion
```

test/helpers/utils.spec.ts
```diff
@@ -132,4 +174,75 @@ describe('redirectsForNext404Route', () => {
...
+
+  it('does not return static 404 redirects when LEGACY_FALLBACK_FALSE is true', async () => {
```
nickytonline, line 200: We only want to ensure that `LEGACY_FALLBACK_FALSE` has a value set. It doesn't matter what the value is.

```suggestion
  it('does not return static 404 redirects when LEGACY_FALLBACK_FALSE is set', async () => {
```

nickytonline (changes requested):

LekoArts:
Linting in unit tests isn't happy

PR #2084, opened 2023-05-04
feat: Use `none` bundler for SSR Routes

Skn0tt:
### Summary

This PR adds a new behaviour where the `none` bundler is used for SSR Routes. It leverages the `*.nft.json`-based file tracing logic added in previous PRs, and extends that to SSR Routes. Before, it was only applied to API Routes.

The new behaviour is behind a flag, which I have enabled for our test suite. It allows us to slowly roll out the change accross customers. 

When both the split-api behaviour and this new behaviour are used in conjunction, bundling times can be significantly reduced to a few seconds.

### Test plan

### Standard checks:

- [ ] Add docs when necessary

---


packages/runtime/src/helpers/flags.ts
```diff
@@ -29,3 +29,10 @@ export const splitApiRoutes = (featureFlags: Record<string, unknown>, publish: s
 
   return isEnabled
 }
+
+export const bundleBasedOnNftFiles = (featureFlags: Record<string, unknown>): boolean => {
```
Skn0tt, line 33: Note to self: We should probably make `splitApiRoute` another activation guard for this.

Skn0tt:
it looks like most of our tests are successful with this enabled! The failing "app-dir" tests look unrelated to this PR. @orinokai would love your review on this :D

packages/runtime/src/templates/getHandler.ts
```diff
@@ -206,6 +206,8 @@ export const getHandler = ({
     throw new Error('Could not find Next.js server')
   }
 
+  process.env.NODE_ENV = 'production';
```
orinokai, line 209: What is the reason for this? Just wondering because this will also be the case when running `netlify dev`.

packages/runtime/src/index.ts
```diff
@@ -50,6 +51,7 @@ const plugin: NetlifyPlugin = {
       cache,
     },
   }) {
+    console.log('using local version')
```
orinokai, line 54: does this need to be removed?

orinokai (approved):
Hey @Skn0tt, just a couple of questions, but otherwise this looks great and very exciting to see!

packages/runtime/src/templates/getHandler.ts
Skn0tt: We're filtering out `.development.js`, to reduce bundle size:

https://github.com/netlify/next-runtime/blob/main/packages/runtime/src/helpers/functions.ts#L274

(This needs to be done manually for whatever reason)

react-dom is an example for where this helps: https://www.npmjs.com/package/react-dom?activeTab=code

By setting `node_env` to `production`, we ensure these files are never `require`d.

> Just wondering because this will also be the case when running netlify dev.

AFAIK, we're directly proxying `next dev` there - so this runtime wouldn't run.

packages/runtime/src/index.ts
Skn0tt: oh absolutely! good catch
```suggestion
```

packages/runtime/src/templates/getHandler.ts
orinokai: Ah, I see. Thanks, that makes sense. And yes, you're absolutely right about proxying `next dev` - i was having a Friday moment and was mistakenly thinking about `netlify serve`!

orinokai (approved):
🚢 🚢 🚢

PR #2087, opened 2023-05-05
chore: update github workflows

LekoArts:
### Summary

This makes use of the correct concurrency setting as explained here: https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

This also updates the checkout and setup-node actions which were out of date.


nickytonline (approved):
Thanks for the update @LekoArts! :shipit:

PR #2104, opened 2023-05-11
fix: don't deploy metadata files to CDN

pieh:
### Summary

Current strategy of hiding metadata files (creating redirects to 404) has a flaw - redirects are case sensitive, but CDN is not so that means we can still actually reach those files by messing with route casing.

As example `/trace` is supposed to be hidden - and initially it appears to be: https://netlify-plugin-nextjs-demo.netlify.app/trace, however https://netlify-plugin-nextjs-demo.netlify.app/tRace allow to view the `trace` file. Preview Deploy for this PR actually make it impossible to serve that asset regardless of casing as it's just not deployed to CDN: https://deploy-preview-2104--netlify-plugin-nextjs-demo.netlify.app/tRaCe/

This PR changes "hiding metadata files" implementation to actually prevent those files from being deployed by deleting them after all the functions bundling (that rely on those metadata files) is finished.

This change also required adjustment to 404 documents - we were not moving them from `.next/server(less)/pages` before - now they are moved the same as regular html files and 404 redirects were updated to match it. This was needed because we do delete `.next/server(less)` before deployment so those 404 documents were just lost and not available to be served.

### Test plan

### Standard checks:

- [ ] Add docs when necessary

---

Ref: https://github.com/netlify/pod-ecosystem-frameworks/issues/491

### DIFF
--- a/packages/runtime/src/constants.ts
+++ b/packages/runtime/src/constants.ts
@@ -1,3 +1,5 @@
+import destr from 'destr'
+
 export const HANDLER_FUNCTION_NAME = '___netlify-handler'
 export const ODB_FUNCTION_NAME = '___netlify-odb-handler'
 export const IMAGE_FUNCTION_NAME = '_ipx'
@@ -6,7 +8,6 @@ export const NEXT_PLUGIN = '@netlify/plugin-nextjs'
 export const HANDLER_FUNCTION_TITLE = 'Next.js SSR handler'
 export const ODB_FUNCTION_TITLE = 'Next.js ISR handler'
 export const IMAGE_FUNCTION_TITLE = 'next/image handler'
-
 // These are paths in .next that shouldn't be publicly accessible
 export const HIDDEN_PATHS = [
   '/cache',
@@ -18,7 +19,7 @@ export const HIDDEN_PATHS = [
   '/build-manifest.json',
   '/prerender-manifest.json',
   '/react-loadable-manifest.json',
-  process.env.NODE_ENV === `test` ? `` : '/BUILD_ID',
+  destr(process.env.NEXT_KEEP_BUILD_ID) ? `` : '/BUILD_ID',
   '/app-build-manifest.json',
   '/app-path-routes-manifest.json',
   '/export-marker.json',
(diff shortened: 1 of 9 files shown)


packages/runtime/src/constants.ts
```diff
@@ -7,18 +9,29 @@ export const HANDLER_FUNCTION_TITLE = 'Next.js SSR handler'
...
-  '/serverless/*',
-  '/trace',
-  '/traces',
-  '/routes-manifest.json',
-  '/build-manifest.json',
-  '/prerender-manifest.json',
-  '/react-loadable-manifest.json',
-  '/BUILD_ID',
-]
+export const HIDDEN_PATHS = destr(process.env.NEXT_KEEP_METADATA_FILES)
```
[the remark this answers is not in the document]
nickytonline: Since it's for E2E/CI only (primarily), we already have a `CI` environment variable. See a usage of it here, https://github.com/netlify/next-runtime/blob/82c1ea371bbe2194058264d7cfad912343334809/package.json#L19.

```suggestion
export const HIDDEN_PATHS = destr(process.env.CI)
```

packages/runtime/src/helpers/files.ts
```diff
@@ -467,3 +478,10 @@ export const movePublicFiles = async ({
     await copy(publicDir, `${publish}${basePath}/`)
   }
 }
+
+export const removeMetadataFiles = async (publish: string) => {
```
nickytonline, line 482: Would it not be faster to `await Promise.all()` or `await Promise.allSettled()`?

pieh: Generally yes, but if we don't want to do 1 by 1 for perf reasons we should still limit concurrency as otherwise some starvation might happen and doing `Promise.all()` might actually be slower than doing it 1 by 1

For files moving we do have concurency of number of cpu cores, with minimum of 2 (if on single core):

https://github.com/netlify/next-runtime/blob/f051e227ca10ddc228b1d044eb835fe66fd1d733/packages/runtime/src/helpers/files.ts#L152-L153

So I would suggest we do the same here?

packages/runtime/src/constants.ts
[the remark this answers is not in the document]
pieh: `CI` env var is `true` on Netlify platform, so we can't rely on that, otherwise we won't actually delete those meta files :(

packages/runtime/src/helpers/files.ts
pieh: Swapped to `p-limit` (like snippet above) in https://github.com/netlify/next-runtime/pull/2104/commits/d709eb1dde0e2fccd2304b65de19eb1428ef17db

nickytonline (approved):
Thanks for this @pieh. The env var for preserving metadata files for E2E tests is not a blocker. I'll let you decide if you opt for the existing `CI` env var or not. 🚢

Issue #2106, opened 2023-05-12
[Bug]: Upgrading @netlify/plugin-nextjs to 4.36.1 breaks he usage with next-intl


KrustyC:
### Summary

I have a repo with NextJs and Netlify as well as next-intl installed. Up until the version 4.36.0, all worked fine, provided I set `NEXT_DISABLE_NETLIFY_EDGE` to `true`, but now when I upgrade it fails with an error `middleware is not a function` and with the following stack trace
```
[next-dev] TypeError: middleware is not a function
    at handler (file:///Users/davidecrestini/repo/sites/next-netlify-test/.netlify/edge-functions/next-dev/index.js:78:28)
    at async FunctionChain.runFunction (https://64523ab4e7865600087fc3df--edge.netlify.app/bootstrap/function_chain.ts:390:22)
    at async FunctionChain.run (https://64523ab4e7865600087fc3df--edge.netlify.app/bootstrap/function_chain.ts:310:20)
    at async handleRequest (https://64523ab4e7865600087fc3df--edge.netlify.app/bootstrap/handler.ts:80:22)
    at async Server.#respond (https://deno.land/std@0.170.0/http/server.ts:299:18)
```

I even installed `@netlify/next` but that didn't solve the issue either, neither when I run the app locally (with `netlify dev`), nor when I push and build it on Neltify.

### Steps to reproduce

1. clone https://github.com/KrustyC/next-netlify-test
2. installed the dependencies and run `netlify dev`; everything should work
3. stop the app
4. run  `npm i @netlify/next @netlify/plugin-nextjs@latest next@latest`
5. run `netlify dev`
6. open http://localhost:8888 on your browser

### A link to a reproduction repository

https://github.com/KrustyC/next-netlify-test

### Next Runtime version

13.3.1 and above

### Is your issue related to the `app` directory (beta)?

- [X] Yes, I am using the `app` directory

### More information about your build

- [X] I am building using the CLI
- [X] I am building using file-based configuration (`netlify.toml`)

### What OS are you using?

Mac OS

### Your netlify.toml file

<details>
  <summary>`netlify.toml`</summary>

```toml
[build]
  command = "npm run build"
  publish = ".next"
  environment = { NODE_VERSION = "16.15.0" }
 
[dev]
 command = "npm run dev"
 
[functions]
  directory = "netlify/functions"
  node_bundler = "esbuild"
  external_node_modules = ["sharp"]
  included_files = ["node_modules/sharp/**/*"]

[...]

PR #2116, opened 2023-05-16
fix: support named & default export for middleware in ntl dev

LekoArts:
## Description

As explained in https://github.com/amannn/next-intl/issues/290#issuecomment-1549110801 it seems that both a named `middleware` export and a `default` export is supported.

This allows both for `netlify dev` when we import the middleware.

### Documentation

N/A

## Tests

No tests were added but I tried the change manually on the repository shared in https://github.com/netlify/next-runtime/issues/2106

1. Existing E2E middleware should continue to work
2. https://github.com/netlify/next-runtime/issues/2106

Fixes https://github.com/netlify/next-runtime/issues/2106

### DIFF
--- a/packages/runtime/src/templates/edge/next-dev.js
+++ b/packages/runtime/src/templates/edge/next-dev.js
@@ -38,7 +38,11 @@ const handler = async (req, context) => {
     const nextMiddleware = await import(`../../middleware.js#${++idx}`)
 
     // The middleware file can export a named `middleware` export or a `default` export
-    middleware = isFunction(nextMiddleware.middleware) ? nextMiddleware.middleware : isFunction(nextMiddleware.default) ? nextMiddleware.default : undefined
+    middleware = isFunction(nextMiddleware.middleware)
+      ? nextMiddleware.middleware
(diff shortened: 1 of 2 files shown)


packages/runtime/src/templates/edge/next-dev.js
```diff
@@ -36,7 +36,9 @@ const handler = async (req, context) => {
     // We need to cache-bust the import because otherwise it will claim it
     // doesn't exist if the user creates it after the server starts
     const nextMiddleware = await import(`../../middleware.js#${++idx}`)
-    middleware = nextMiddleware.middleware
+
+    // The middleware file can export a named `middleware` export or a `default` export
+    middleware = isFunction(nextMiddleware.middleware) ? nextMiddleware.middleware : nextMiddleware.default
```
nickytonline, line 41: Is the function check necessary? I imagine `nextMiddleware.middleware` is `undefined` if the `default` exists?

```suggestion
    middleware = nextMiddleware.middleware ?? nextMiddleware.default
```

pieh (approved):

PR #2117, opened 2023-05-16
fix: ensure next-server patches only run on Netlify

zachleat:
This restores non-patched behavior for standalone next server.

## Description

When you run `next dev` standalone after `netlify dev` has patched Next files, it can skip middleware and other configuration. This makes the patches a little more defensive, requiring `process.env.NETLIFY` to be in place. This is important as the other ENV variable checks were `!== 'true'` which successfully passed when `next dev` was in use.

## Tests

This test case was validated https://github.com/anthonyholmes/next-netlify-middleware-bug (noting that step 6 in the readme is inaccurate)

image

Fixes netlify/pod-ecosystem-frameworks#474


orinokai (approved):
Nice! Thanks @zachleat, lgtm 🚢

PR #2121, opened 2023-05-17
fix: clientContext isn't available in background functions

Skn0tt:
## Description

This PR fixes a regression introduced in https://github.com/netlify/next-runtime/pull/2058. `clientContext` isn't available in backround functions, so destructuring on it throws an error.

## Tests

1. TODO

### DIFF
--- a/packages/runtime/src/templates/getApiHandler.ts
+++ b/packages/runtime/src/templates/getApiHandler.ts
@@ -70,9 +70,7 @@ const makeApiHandler = ({ conf, app, pageRoot, NextServer }: MakeApiHandlerParam
       return bridge
     }
 
-    const {
-      clientContext: { custom: customContext },
-    } = context
+    const customContext = context.clientContext?.custom
 
     // Scheduled functions don't have a URL, but we need to give one so Next knows the route to serve
     const url = event.rawUrl ? new URL(event.rawUrl) : new URL(path, process.env.URL || 'http://n')


packages/runtime/src/templates/getApiHandler.ts
```diff
@@ -70,9 +70,7 @@ const makeApiHandler = ({ conf, app, pageRoot, NextServer }: MakeApiHandlerParam
       return bridge
     }
 
-    const {
-      clientContext: { custom: customContext },
-    } = context
+    const customContext = context.clientContext?.custom
```
MarcL, line 73: Do we have any way to test this @Skn0tt?
I appreciate we might not have a good testing strategy yet for this.

Skn0tt: I don't think we have a good testing strategy yet, no. We could probably build something to test background functions using https://webhook.site/, but i'd like to address that in a separate PR/issue.

orinokai (approved):
lgtm - neater too!

packages/runtime/src/templates/getApiHandler.ts
Skn0tt: Tracked the need for an E2E test in https://github.com/netlify/next-runtime/issues/2122.

MarcL: Ok - thanks!

PR #2123, opened 2023-05-17
fix: failing test for background functions

### LINKED ISSUE #2122: Add E2E test for background functions
In #2058 we introduced a regression around `experimental-background` functions. It was fixed in https://github.com/netlify/next-runtime/pull/2121, and could've been caught by an E2E test.

We should build an E2E test that asserts on background functions running to completion, maybe using a tool like https://webhook.site/.

Skn0tt:
## Description

Closes #2122 

## Tests

1. TODO


nickytonline:
@Skn0tt, do you want us to take this over?

Skn0tt:
that'd be lovely!

PR #2124, opened 2023-05-17
fix: resolve follow-redirects from runtime package

Skn0tt:
## Description

If the Next Runtime is auto-installed, then `follow-redirects` (which is a dep of `next-runtime`) won't be resolvable from `.next` (the `publish` dir). This results in an error.

To fix this, I'm adding `__dirname` as a second resolution starting point. `__dirname` always points to a file within the `next-runtime` package, so `follow-redirects` should be resolvable from there.

Previously discussed in https://netlify.slack.com/archives/C02T8ASK2E9/p1684318093610159.

## Tests

1. TODO

### DIFF
--- a/packages/runtime/src/helpers/functions.ts
+++ b/packages/runtime/src/helpers/functions.ts
@@ -269,7 +269,7 @@ const traceNextServer = async (publish: string): Promise<string[]> => {
 
 export const traceNPMPackage = async (packageName: string, publish: string) => {
   try {
-    return await glob(join(dirname(require.resolve(packageName, { paths: [publish] })), '**', '*'), {
+    return await glob(join(dirname(require.resolve(packageName, { paths: [__dirname, publish] })), '**', '*'), {
       absolute: true,
     })
   } catch (error) {


packages/runtime/src/helpers/functions.ts
```diff
@@ -269,7 +269,7 @@ const traceNextServer = async (publish: string): Promise<string[]> => {
 
 export const traceNPMPackage = async (packageName: string, publish: string) => {
   try {
-    return await glob(join(dirname(require.resolve(packageName, { paths: [publish] })), '**', '*'), {
+    return await glob(join(dirname(require.resolve(packageName, { paths: [__dirname, publish] })), '**', '*'), {
```
danez, line 272: Are we sure this file is exectued as commonjs and not ESM? because `__dirname` is only available in commonjs.

danez: Okay I see that `__dirname` is also used on other locations in this file.

danez (approved):

packages/runtime/src/helpers/functions.ts
eduardoboucas: It better be, because `require.resolve` is also absent from ESM.

PR #2125, opened 2023-05-17
fix: guard split api routes behaviour by next version

Skn0tt:
## Description

We discovered that the Split-API-Routes behaviour introduced in #2058 depends on Next.js v12+. Because our Next Runtime needs to support everything from v10 up, I'm adding an additional guard that ensures versions sub v12 continue to use the current behaviour.

## Tests

1. TODO

### DIFF
--- a/packages/runtime/src/helpers/flags.ts
+++ b/packages/runtime/src/helpers/flags.ts
@@ -22,7 +22,7 @@ export const splitApiRoutes = (featureFlags: Record<string, unknown>, publish: s
 
   if (isEnabled && !existsSync(join(publish, 'next-server.js.nft.json'))) {
     console.warn(
-      'Trace-based bundling not possible on this version of Next.js. Speed up your builds significantly by upgrading to Next.js v12.',
+      'Trace-based bundling not possible on this version of Next.js. Speed up your builds significantly by upgrading to Next.js v12 or newer.',
     )
     return false
   }
--- a/packages/runtime/src/helpers/flags.ts
+++ b/packages/runtime/src/helpers/flags.ts
@@ -1,4 +1,6 @@
 import destr from 'destr'
+import { existsSync } from 'fs-extra'
+import { join } from 'pathe'
 
 /**
  * If this flag is enabled, we generate individual Lambda functions for API Routes.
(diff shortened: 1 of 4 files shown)


packages/runtime/src/helpers/flags.ts
```diff
@@ -11,7 +13,19 @@ import destr from 'destr'
...
  * Disabled by default. Can be overriden using the NEXT_SPLIT_API_ROUTES env var.
  */
-export const splitApiRoutes = (featureFlags: Record<string, unknown>): boolean =>
-  destr(process.env.NEXT_SPLIT_API_ROUTES) ?? featureFlags.next_split_api_routes ?? false
+export const splitApiRoutes = (featureFlags: Record<string, unknown>, publish: string): boolean => {
+  const isEnabled = destr(process.env.NEXT_SPLIT_API_ROUTES) ?? featureFlags.next_split_api_routes ?? false
+
+  if (isEnabled && !existsSync(join(publish, 'next-server.js.nft.json'))) {
+    console.warn(
+      'Trace-based bundling not possible on this version of Next.js. Speed up your builds significantly by upgrading to Next.js v12.',
```
LekoArts, line 25:
```suggestion
      'Trace-based bundling not possible on this version of Next.js. Speed up your builds significantly by upgrading to Next.js v12 or newer.',
```

packages/runtime/src/helpers/flags.ts
```diff
@@ -11,7 +13,19 @@ import destr from 'destr'
  * If disabled, we bundle all API Routes into a single function.
  * This is can lead to large bundle sizes.
  *
+ * Relies on `next-server.js.nft.json`, which is only supported in Next.js 12+.
+ *
  * Disabled by default. Can be overriden using the NEXT_SPLIT_API_ROUTES env var.
  */
-export const splitApiRoutes = (featureFlags: Record<string, unknown>): boolean =>
-  destr(process.env.NEXT_SPLIT_API_ROUTES) ?? featureFlags.next_split_api_routes ?? false
+export const splitApiRoutes = (featureFlags: Record<string, unknown>, publish: string): boolean => {
```
nickytonline, line 20: This looks good. Can you add some tests for this?

Skn0tt: done in https://github.com/netlify/next-runtime/pull/2125/commits/efa9787fd2c4adae4df8341ee545b4abb7596f9f.

orinokai (approved):
awesome stuff, thanks @Skn0tt

PR #2133, opened 2023-05-24
fix: remove edge functions beta label

Skn0tt:
## Description

see https://netlify.slack.com/archives/C0183KVDHFH/p1684936589886179 for reference

## Tests

1. TODO


nickytonline (approved):
🚢

PR #2136, opened 2023-05-25
fix: tracing root use `appDir` as fallback

Skn0tt:
## Description

We were relying on values `relativeAppDir` and `outputFileTracingRoot` that were optional, and don't exist for every version of Next.js.

`relativeAppDir` was introduced in https://github.com/vercel/next.js/pull/46393.
`appDir` was introduced in https://github.com/vercel/next.js/pull/22915.
`outputFileTracingRoot` was introduced in https://github.com/vercel/next.js/pull/22915.

This PR makes it so that we fall back to use `appDir` if `outputFileTracingRoot` and `relativeAppDir` aren't available.

## Tests

1. TODO

### DIFF
--- a/packages/runtime/src/helpers/functions.ts
+++ b/packages/runtime/src/helpers/functions.ts
@@ -231,11 +231,12 @@ export const setupImageFunction = async ({
 const traceRequiredServerFiles = async (publish: string): Promise<string[]> => {
   const requiredServerFiles = await getRequiredServerFiles(publish)
 
-  const relativeAppDir = requiredServerFiles.relativeAppDir ?? ''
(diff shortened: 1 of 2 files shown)


Skn0tt:
We could've caught this earlier, if `strictNullChecks` were enabled:

> https://github.com/netlify/next-runtime/blob/8e1936f352eb9b8b040e82a6c7e9713a02787333/tsconfig.json#L31

@orinokai do you think it'd make sense to enable TypeScript strict mode for this repo?

Skn0tt:
Ha, looks like there's failures! Turning into a draft.

pieh (approved):

PR #2138, opened 2023-05-25
fix: pass only first language to next-server to match platform redirects support

pieh:
## Description

Netlify locale redirects match only on first language listed in `Accept-Language` header. `next-server` handles all of them.
https://docs.netlify.com/routing/redirects/redirect-options/#redirect-by-country-or-language
> Language-based redirects always match against the first language reported by the browser in the Accept-Language header regardless of quality value weighting.

This means some requests might skip Netlify language redirects (when Automatic Locale Detection in next.js is not disabled) and hit lambda that could produce locale redirect. For SSR this might not a problem, but this is a problem for ODB, because that locale redirect gets cached for everyone.

This PR basically tricks `next-server` to only handle first language (to match platform capability) by removing all languages except of first one in `Accept-Language` header before passing it to `next-server`

## Tests

Added cypress test, but can be also verified manually like this:

Examine status code / response headers of requests to `/` with `__prerender_bypass` cookie (`/` is using ODB, so we want to ensure we execute it) with specifically crafter `Accept-Language` where first language on the list is not in i18n config, but there is a language in the list overall that is in i18n config - for example `Accept-Language: pl, fr-CH;q=0.9, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5` (`pl` not in i18n, `fr` is in i18n)

### `main` branch

Curl command:
```shell
curl --head --url https://netlify-plugin-nextjs-demo.netlify.app/ --header 'Accept-Language: pl, fr-CH;q=0.9, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5' --header 'cookie: __prerender_bypass=1' --header 'x-next-debug-logging: 1' --header 'x-nf-debug-logging: 1' --cookie __prerender_bypass=1
```

Result (truncated for readability):

```
HTTP/2 307
location: /fr/
```

Above is the problem because this would be cached for everyone if request like that happens before ODB cache is hydrated or after it expired. Note - 307 is redirect from `next-server`, our are 301s (separate discussion if those should be 301 or 302)

### Preview deploy for this PR:

Curl command:
```shell
curl --head --url https://deploy-preview-2138--netlify-plugin-nextjs-demo.netlify.app/ --header 'Accept-Language: pl, fr-CH;q=0.9, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5' --header 'cookie: __prerender_bypass=1' --header 'x-next-debug-logging: 1' --header 'x-nf-debug-logging: 1' --cookie __prerender_bypass=1
```
Result:

```
HTTP/2 200

```


packages/runtime/src/templates/getHandler.ts
```diff
@@ -130,6 +130,14 @@ const makeHandler = ({ conf, app, pageRoot, NextServer, staticManifest = [], mod
     const query = new URLSearchParams(event.queryStringParameters).toString()
     event.path = query ? `${event.path}?${query}` : event.path
 
+    if (event.headers['accept-language']) {
+      // keep just first language to match Netlify redirect limitation:
+      // https://docs.netlify.com/routing/redirects/redirect-options/#redirect-by-country-or-language
+      // > Language-based redirects always match against the first language reported by the browser in the Accept-Language header regardless of quality value weighting.
+      // If we wouldn't keep just first language, it's possible for `next-server` to generate locale redirect that could be cached by ODB
+      event.headers['accept-language'] = event.headers['accept-language'].replace(/\s*,.*$/, '')
```
pieh, line 138: maybe this should be done just for `odb` mode, and let `ssr` keep all languages, but it feels weird to change behavior between 2 of them?

orinokai: Having different behaviour between SSR/ODB feels wrong, and I guess it's because what we're doing is actually trying to workaround our lack of support by imposing the same limitation on Next, which would be a regression.

pieh: Yeah, to me it's picking one of 2 bad options and figuring it which one is worse/better. I do agree it's probably better to keep SSR pages behaving as they are (so supporting matching on all languages on the header, even if there are multiple source that can produce redirect) and limit the workaround to just ODB (it does technically make things worse to match on lowest common denominator to avoid caching wrong things).

In any case https://github.com/netlify/next-runtime/pull/2138/commits/4c85c21dc8198335612e24cb1556a0e81e18f5c3 limits the workaround just to ODB

orinokai (approved):
Thanks @pieh, lgtm 🚢

PR #2154, opened 2023-06-05
fix: bundle dynamic API routes correctly with split-api-routes

Skn0tt:
## Description

We discovered that dynamic API routes, e.g. `/api/shows/[id].js`, aren't bundled correctly. This is likely due to the square brackets being interpreted as special glob patterns, resulting in the underlying files not being included correctly.

This PR adds an e2e test that demonstrates the bug, and then fixes it by processing the glob patterns before passing them to the build process.

## Tests

1. TODO

https://github.com/netlify/pod-dev-foundations/issues/516

### DIFF
--- a/packages/runtime/src/helpers/config.ts
+++ b/packages/runtime/src/helpers/config.ts
@@ -96,6 +96,12 @@ export const hasManuallyAddedModule = ({
   )
 /* eslint-enable camelcase */
 
+/**
+ * Transforms `/api/shows/[id].js` into `/api/shows/*id*.js`,
+ * so that the file `[id].js` is matched correctly.
+ */
+const escapeGlob = (path: string) => path.replace(/\[/g, '*').replace(/]/g, '*')
+
 export const configureHandlerFunctions = async ({
   netlifyConfig,
   publish,
@@ -173,7 +179,7 @@ export const configureHandlerFunctions = async ({
       netlifyConfig.functions[functionName] ||= { included_files: [] }
       netlifyConfig.functions[functionName].node_bundler = 'none'
       netlifyConfig.functions[functionName].included_files ||= []
-      netlifyConfig.functions[functionName].included_files.push(...includedFiles)
+      netlifyConfig.functions[functionName].included_files.push(...includedFiles.map(escapeGlob))
     }
   } else {
     configureFunction('_api_*')
--- a/cypress/e2e/default/api.cy.ts
+++ b/cypress/e2e/default/api.cy.ts
@@ -6,6 +6,14 @@ describe('API routes', () => {
   })
 })
 
+describe('Dynamic API routes', () => {
+  it('are bundled correctly', () => {
+    cy.request('/api/shows/1').then((response) => {
+      expect(response.status).to.equal(200)
+    })
+  })
+})
+
 describe('Extended API routes', () => {
   it('returns HTTP 202 Accepted for background route', () => {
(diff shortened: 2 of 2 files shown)


Skn0tt:
Here's the failing test: https://github.com/netlify/next-runtime/actions/runs/5176735315/jobs/9325929677#step:10:236

Skn0tt:
https://github.com/netlify/next-runtime/pull/2154/commits/d19f422e1f5f239c502c2e07353cb3c8f08a577b should make it green 🟢

nickytonline (commented):
The fix looks good, but I'm trying to understand how our existing tests didn't fail as we have other dynamic routes. Is the API route splitting only enabled on the default demo site?

Also, it'd be good to test with and without the split API routes feature flag as it's not rolled out to everyone yet.

nickytonline (commented):
The fix looks good, but I'm trying to understand how our existing tests didn't fail as we have other dynamic routes. Is the API route splitting only enabled on the default demo site?

Also, it'd be good to test with and without the split API routes feature flag as it's not rolled out to everyone yet.

Skn0tt:
> The fix looks good, but I'm trying to understand how our existing tests didn't fail as we have other dynamic routes.

The bug only exists for API Routes, not for SSR routes, since we're currently only using the new bundling mechanism for API Routes. I don't think we tested any other dynamic API Routes before, or else we'd have caught this bug earlier.

> Is the API route splitting only enabled on the default demo site?

It's enabled in 4 demo sites, which I think are the sites that we're running e2e tests against: https://github.com/search?q=repo:netlify/next-runtime+NEXT_SPLIT_API_ROUTES&type=code

> Also, it'd be good to test with and without the split API routes feature flag as it's not rolled out to everyone yet.

With the way our e2e tests work, I don't see how we can do that. I think that's out-of-scope for this PR as well. I'd prefer that we get the split-api-routes work done quickly, and remove the featureflag before it becomes long-lived enough to warrant additional test cases.

packages/runtime/src/helpers/config.ts
```diff
@@ -96,6 +96,12 @@ export const hasManuallyAddedModule = ({
   )
 /* eslint-enable camelcase */
 
+/**
+ * Transforms `/api/shows/[id].js` into `/api/shows/*id*.js`,
+ * so that the file `[id].js` is matched correctly.
+ */
+const escapeGlob = (path: string) => path.replace(/\[/g, '*').replace(/]/g, '*')
```
orinokai, line 103: i think this is going to match unwanted files, e.g. `grid.js`
is it possible to escape the path, e.g. `\[id\].js` ?

Skn0tt: I tried escaping the path, and glob patterns do allow escaping the brackets - but I couldn't get that to work properly with how ZISI interprets the globs. Maybe we could open a follow-up issue to investigate this.

Yes, this matches too many files - but I don't think that it'll be a big problem in practice.

Skn0tt: Good to know it's fixed in v9! I'd prefer the follow-up. Let's open an issue for this.

LekoArts (approved):

packages/runtime/src/helpers/config.ts
LekoArts: Opened https://github.com/netlify/zip-it-and-ship-it/issues/1459

PR #2156, opened 2023-06-06
refactor: use LAMBDA_WARNING_SIZE as packing threshold

Skn0tt:
## Description

While going through the codebase, I discovered a magic number I left floating around. This PR replaces that with a reference to a constant we already had.

## Tests

1. TODO

### DIFF
--- a/test/helpers/functionsMetaData.spec.ts
+++ b/test/helpers/functionsMetaData.spec.ts
@@ -4,6 +4,7 @@ import { join } from 'pathe'
 
 import { NEXT_PLUGIN_NAME } from '../../packages/runtime/src/constants'
 import { writeFunctionConfiguration } from '../../packages/runtime/src/helpers/functionsMetaData'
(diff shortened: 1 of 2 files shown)


test/helpers/functionsMetaData.spec.ts
```diff
@@ -4,6 +4,7 @@ import { join } from 'pathe'
 
 import { NEXT_PLUGIN_NAME } from '../../packages/runtime/src/constants'
 import { writeFunctionConfiguration } from '../../packages/runtime/src/helpers/functionsMetaData'
+import { getFunctionNameForPage } from '../../packages/runtime/src/helpers/utils'
```
nickytonline, line 7: This appears to be unrelated to the PR and is currently unused in this PR

```suggestion
```

nickytonline (changes requested):
Looks good, one change request.

MarcL (approved):
LGTM. :shipit:

PR #2159, opened 2023-06-12
fix: Return MiddlewareResponse obj for rewrite

taty2010:
## Description
Fixes https://github.com/netlify/pod-ecosystem-frameworks/issues/510

We are now returning the MiddlewareResponse from the request.rewrite() instead of NextResponse. This will allow us to utilize the setPageProp and replaceText method.

This PR is to get the initial issue resolved, we will continue the work in another PR for the init parameter.

## Tests
This [page](https://deploy-preview-2159--next-plugin-edge-middleware.netlify.app/request-rewrite) should be rewritten with the cat ad and message with geo location

1. TODO


taty2010:
All the tests should be passing on this, there might have been an update recently with the app-dir tests. Will have to review once I'm back.

cypress/e2e/middleware/enhanced.cy.ts
```diff
@@ -1,5 +1,24 @@
...
+  it('rewrites the response body using request.rewrite()', () => {
+    cy.visit('/request-rewrite')
+    cy.get('#message').contains('This was static (& escaping test &amp;) but has been transformed in')
+    cy.contains("This is an ad that isn't shown by default")
+  })
+
+  it('modifies the page props when using request.rewrite()', () => {
+    cy.visit('/request-rewrite')
+    const data = cy.get('script#__NEXT_DATA__')
+    data.should('contain', '"message":"This was static (& escaping test &amp;) but has been transformed in')
```
nickytonline, line 11: Consider testing that the `props.pageProps.message` property includes that text instead of just checking the raw string.

```suggestion
    const data = cy.get('script#__NEXT_DATA__').then((element) => {
      const { props } = JSON.parse(element.text());
      expect(props.pageProps.message).to.include('This was static (& escaping test &amp;) but has been transformed in')
    })
```

cypress/e2e/middleware/enhanced.cy.ts
```diff
@@ -1,5 +1,24 @@
...
+  })
+
+  it.skip('passes in headers within request.rewrite()', () => {
+    cy.request('/request-rewrite').then((response) => {
+      expect(response.headers).to.have.property('x-rewrite-test', 'hello')
+    })
+  })
+
+
```
nickytonline, line 20: This can be removed for now as it will be sorted out in another PR.

demos/middleware/middleware.ts
```diff
@@ -27,7 +27,22 @@ export async function middleware(req: NextRequest) {
...
+    res.setPageProp('message', message)
+    res.setPageProp('showAd', true)
+
+    res.headers.set('x-modified-edge', 'true')
+    res.headers.set('x-is-deno', 'Deno' in globalThis ? 'true' : 'false')
+    return res
+  }
+
+  if (pathname.startsWith('/request-rewrite')) {
+    // request.rewrite() should return the MiddlewareResponse obj instead of the Response obj
```
nickytonline, line 40:
```suggestion
    // request.rewrite() should return the MiddlewareResponse object instead of the Response object.
```

demos/middleware/pages/request-rewrite.js
```diff
@@ -0,0 +1,10 @@
+const Rewrite = () => {
+    return (
+      <div>
+        <p>This should have been rewritten</p>
+      </div>
+    )
+  }
+  
+  export default Rewrite
+
```
nickytonline, line 10: nit: add a new line at the end of the file

test/e2e/next-test-lib/next-test-utils.js
```diff
@@ -2,7 +2,7 @@
 import spawn from 'cross-spawn'
 import { existsSync, readFileSync, unlinkSync, writeFileSync } from 'fs'
 import { writeFile } from 'fs-extra'
-import { fetch as undiciFetch } from 'undici'
+import { fetch as undiciFetch } from 'next/dist/compiled/undici'
```
nickytonline, line 5: I guess we never figured out why the `undici` dep was no longer included?

taty2010: It happened here https://github.com/netlify/next-runtime/commit/c295bc23bd769e0d946a40d75ee7340f4d5cdf9d
it got removed from package-lock but it looks like it was not set as a dep in the package.json.

nickytonline: Once we move to Node.js 18, `fetch` is native, so this import won't be required.

nickytonline (approved):
Looks great! :shipit:

PR #2080, opened 2023-05-01
chore: upgrade to next v13.3.4

taty2010:
### Test plan

1. Visit the Deploy Preview ([insert link to specific page]()) ...

### Standard checks:

- [ ] Check the Deploy Preview's Demo site for your PR's functionality
- [ ] Add docs when necessary

---

🧪 Once merged, make sure to update the version if needed and that it was published correctly.

PR #2176, opened 2023-06-21
chore: enable server tests

orinokai:
## Description

Server tests were previously disabled to ship the [13.4 update PR](https://github.com/netlify/next-runtime/pull/2080) while the on-demand revalidation fix happened in parallel.

We have now confirmed that this works so this PR re-enables the tests.


serhalp:
Superseded by v5.

PR #2179, opened 2023-06-22
fix: optionally chain server functions when selecting react version

orinokai:
## Description

The previous Next 13.4 [update PR](https://github.com/netlify/next-runtime/pull/2080/) introduced a route check to optionally use the prebundled react module, as required by Next.

This PR makes it backwards-compatible with Next 11.

## Tests

Automated tests for previous Next versions are currently in discussion.


MarcL (approved):
👍🏻

PR #2188, opened 2023-06-26
fix: edge cases for custom react module resolution

orinokai:
## Description

Fix to ensure that the default React module resolution is always used as a fallback when rendering Page Router routes.

This PR also guards against missing module paths in previous versions of Next by try/catching each call to `require.resolve()`.

In addition, this PR fixes a case where paths with query strings caused the route matching to fail.

### DIFF
--- a/packages/runtime/src/templates/requireHooks.ts
+++ b/packages/runtime/src/templates/requireHooks.ts
@@ -88,9 +88,13 @@ const resolveRequireHooks = () => {
       try {
         const resolvedPath = require.resolve(path, opts)
         mode.set(hook, resolvedPath)
-      } catch {
-        // module not present (older version of Next.js)
-        mode.delete(hook)
+      } catch (error) {
+        if (error.code === 'MODULE_NOT_FOUND') {
+          // module not present (older version of Next.js)
+          mode.delete(hook)
+        } else {
+          throw error
+        }
       }
     })
   })
--- a/packages/runtime/src/templates/requireHooks.ts
+++ b/packages/runtime/src/templates/requireHooks.ts
@@ -89,7 +89,7 @@ const resolveRequireHooks = () => {
         const resolvedPath = require.resolve(path, opts)
         mode.set(hook, resolvedPath)
       } catch {
-        // ignore
+        // module not present (older version of Next.js)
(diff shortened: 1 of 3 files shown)


packages/runtime/src/templates/requireHooks.ts
```diff
@@ -11,86 +11,89 @@ const resolveFilename = (mod as any)._resolveFilename
...
+  const opts = { paths: [process.cwd()] }
+
+  // resolve require hooks with module paths
+  requireHooks.forEach((mode) => {
+    mode.forEach((hook, path) => {
+      try {
+        const resolvedPath = require.resolve(path, opts)
+        mode.set(hook, resolvedPath)
+      } catch {
+        // ignore
```
MarcL, line 92: Do we need any logging here for our observability or is it safe to just ignore?

MarcL (approved):
Question about logging but looks good otherwise.

packages/runtime/src/templates/requireHooks.ts
orinokai: hey @MarcL, this is safe to ignore here because it's not actually an error - we want to delete the hook in this case because it happens when a module isn't found, i.e. an older version of Next
I have updated with a better comment, thanks

packages/runtime/src/templates/requireHooks.ts
```diff
@@ -11,86 +11,89 @@ const resolveFilename = (mod as any)._resolveFilename
...
+  // we may have changed the working directory in the handler
+  const opts = { paths: [process.cwd()] }
+
+  // resolve require hooks with module paths
+  requireHooks.forEach((mode) => {
+    mode.forEach((hook, path) => {
+      try {
+        const resolvedPath = require.resolve(path, opts)
+        mode.set(hook, resolvedPath)
+      } catch {
```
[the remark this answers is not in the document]
orinokai: @nickytonline good plan, have updated in [78fc4d6](https://github.com/netlify/next-runtime/pull/2188/commits/78fc4d69b69de595abc5f39e7185478e1da5075d)

nickytonline (dismissed):
🚢

LekoArts (approved):

PR #2189, opened 2023-06-26
fix: override headers within req.rewrite()

taty2010:
## Description

Based on the description provided [here](https://github.com/netlify/pod-ecosystem-frameworks/issues/183) this update makes it to where we can now override the headers within the req.rewrite() like in the example below.

```js
    const res = await request.rewrite('/static-rewrite',
    {
      headers: {
        'x-rewrite-test': 'hello',
        'x-rewrite-test-2': 'hello-2'
      }
    })
```

## Tests

1. Go to https://deploy-preview-2189--next-plugin-edge-middleware.netlify.app/request-rewrite
2. review the headers within the Network tab and should see the headers from the example code above.


cypress/e2e/middleware/enhanced.cy.ts
```diff
@@ -13,6 +13,12 @@ describe('Enhanced middleware', () => {
     })
   })
 
+  it.skip('passes in headers within request.rewrite()', () => {
```
MarcL, line 16: Should this be skipped?

MarcL (commented):
Think it just needs the test to be enabled and good to go.

cypress/e2e/middleware/enhanced.cy.ts
taty2010: ah this one should not be skipped, thanks!

nickytonline (approved):
Looks good! 🚢

PR #2206, opened 2023-07-10
fix: use native path matching for prebundled react conditional

orinokai:
## Description

During the upgrade to Next 13.4 we implemented a fix so that the correct prebundled version of React is chosen when the route is App vs Pages router. However, we were attempting to determine the type of route using a custom path matching algorithm and it was failing when the route used route groups.

This PR uses the Next native `getAppPathRoutes` function to normalize the route, removing route groups and other artifacts in the path.

## Tests

Unit tests have been added (and former tests restored as these were yet to be re-enabled after #2080 was merged).

Fixes netlify/pod-ecosystem-frameworks#489 (again)

### DIFF
--- a/packages/runtime/src/templates/server.ts
+++ b/packages/runtime/src/templates/server.ts
@@ -12,7 +12,6 @@ import {
   localizeRoute,
   localizeDataRoute,
   unlocalizeRoute,
-  joinPaths,
 } from './handlerUtils'
 
 interface NetlifyConfig {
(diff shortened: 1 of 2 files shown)


MarcL (approved):
Looks good to me but might be wise to get @pieh or @nickytonline to review it before merging.

packages/runtime/src/templates/server.ts
```diff
@@ -78,12 +77,11 @@ const getNetlifyNextServer = (NextServer: NextServerType) => {
     // doing what they do in https://github.com/vercel/vercel/blob/1663db7ca34d3dd99b57994f801fb30b72fbd2f3/packages/next/src/server-build.ts#L576-L580
     private netlifyPrebundleReact(path: string) {
       const routesManifest = this.getRoutesManifest?.()
-      const appPathsManifest = this.getAppPathsManifest?.()
+      const appPathsRoutes = this.getAppPathRoutes?.()
```
nickytonline, line 80: Makes sense since `this.getAppPathRoutes()` calls `this.getAppPathsManifest()`.

https://github.com/vercel/next.js/blob/4ddb6fc794201461bea80df5917b6f0a75f62986/packages/next/src/server/base-server.ts#L1042-L1053

nickytonline (approved):
🚢

packages/runtime/src/templates/server.ts
orinokai: Yeah exactly, it calls it and then does the normalization I was trying to do myself!

Issue #2220, opened 2023-07-19
Improve logging


MarcL:
The runtime currently uses `console.log`s for its logging and it's not easy to add additional logging for debugging purposes without cluttering the build logs.

We should add a logging library to enable us to use log levels for debugging and eventually be able to use structured logging.

MarcL:
Closing this as we move to v5 runtime.

PR #2221, opened 2023-07-19
chore: better logging

MarcL:
## Description

This PR adds the [winston](https://github.com/winstonjs/winston) logger to enable us to have clearer logging in the runtime. It will help us to add additional logging to help with debugging but avoid additional clutter in the logs. This is due to the default log level of `info` being set.

This is a suggestion to see if the team think it'll be useful. It will allow us to set a `LOG_LEVEL` environment variable to enable logging levels below the current default. With this, we can instrument the code and allow customers to enable a lower log level if we're struggling to debug issues.

I've added the `no-console` eslint rule to avoid accidental `console.*` statements but have not enabled it for tests. See the inline comments for more details.

I considered using [pino](https://github.com/pinojs/pino) too as it's faster. However, by default, it generates structured JSON logs which can't currently be ingested by Netlify's build process. You can use the [pino-pretty](https://github.com/pinojs/pino-pretty) package to display them nicely but it adds an additional timestamp. `winston` gave a simpler 1-line transport which has parity with our console logging so in the short term we can keep this.

### Documentation

N/A

## Tests

I've updated tests which spy or mock `console.*` to use the `logger.*` instead.
There is 1 failing snapshot test that we need to look at before merging.

image

#2220


packages/runtime/src/helpers/logger/index.ts
```diff
@@ -0,0 +1,3 @@
+// Winston supports JSON by default but this is using a custom formatter
+// It only gives the message and not the timestamp or log level
+export { logger as default } from './winston'
```
MarcL, line 3: I had this logger exposed via an index file as I was trying both `winston` and `pino`. Pino is meant to be faster but by default supports JSON logging. Even with their pretty formatter, it would make our logs a little strange with multiple timestamps.

We can move this to a single logger module that only includes `winston` if needed.

packages/runtime/src/helpers/logger/winston.ts
```diff
@@ -0,0 +1,11 @@
+import winston from 'winston'
+
+// Define log level from environment variable so we can change it if needed
+const logLevel = process.env.LOG_LEVEL ?? 'info'
```
MarcL, line 4: This will allow us to only display `logger.info` and higher levels (`.warn`, `.error`) into the user-facing logs.

We can use the environment variable with customers or support so we can use `logger.debug` or similar in the code.

orinokai (approved):
Thanks Marc! Seems like a no-brainer as it gives us extra flexibility and observability, plus we can avoid leaving accidental `console.log` in the code. Unless the performance difference is huge, I'd advocate for using Winston instead of developing a custom transport for Pino.

.eslintrc.js
```diff
@@ -30,6 +30,7 @@ module.exports = {
     'unicorn/no-array-push-push': 0,
     'unicorn/numeric-separators-style': 0,
     'max-lines': 0,
+    'no-console': 2,
```
MarcL, line 33: ESLint now errors for all `console.*` statements.

packages/runtime/src/helpers/files.ts
```diff
@@ -271,13 +272,16 @@ export const moveStaticPages = async ({
...
   if (shortRevalidateRoutes.length !== 0) {
-    console.log(outdent`
+    logger.info(outdent`
       The following routes use "revalidate" values of under ${MINIMUM_REVALIDATE_SECONDS} seconds, which is not supported.
       They will use a revalidate time of ${MINIMUM_REVALIDATE_SECONDS} seconds instead.
     `)
+
+    // Disabling lint rule as the logger doesn't have a table formatter
+    // eslint-disable-next-line no-console
     console.table(shortRevalidateRoutes)
```
MarcL, line 282: I wasn't sure what we wanted to do here. We'd have to write some code to format as a table so I've left `console.table` as it was.

packages/runtime/src/templates/getApiHandler.ts
```diff
@@ -1,3 +1,6 @@
+// This is a runtime function so assuming we don't want to add
+// an additional dependency on a logger (Winston) for runtime code
+/* eslint-disable no-console */
```
MarcL, line 3: My assumption for all of our runtime code is that we don't want to add the extra dependency of `winston` to avoid bundle bloat.

test/helpers/analysis.spec.ts
```diff
@@ -1,15 +1,18 @@
...
+import logger from '../../packages/runtime/src/helpers/logger'
 
 describe('static source analysis', () => {
   beforeEach(() => {
-    //  Spy on console.error
-    jest.spyOn(console, 'error').mockImplementation(() => {})
+    //  Spy on logger.error
+    // TODO: Request for PR comments - Do we need to mock this?
+    // We can just leave it as is, but it will print out the error message
+    jest.spyOn(logger, 'error')
```
MarcL, line 11: I wasn't sure what to do here.

`logger.error` expects a return type of a Winston `Logger`. We don't really need to mock this unless we care about the test output including the `logger.error` strings.

Thoughts?

test/index.spec.ts
```diff
@@ -1,3 +1,5 @@
+// Do we want logging in the tests?
+/* eslint-disable no-console */
```
MarcL, line 2: Not sure if we care about `console.*` statements in the tests.

Personally, I'm not sure we should be including additional logging in the tests as the tests should provide the information we need.

Thoughts?

test/index.spec.ts
```diff
@@ -810,9 +813,9 @@ describe('onPostBuild', () => {
       },
     ])
 
-    const oldLog = console.log
+    const oldLoggerInfo = logger.info
```
MarcL, line 816: I left this as was but we could probably just spy on the `logger.info` here to confirm it gave the message we expect.

MarcL:
I've updated the PR and added logging to all of the code.

MarcL:
Some test snapshots fail so I'll need to work with @nickytonline or @orinokai to understand why. They don't look like they're related to any of the changes in this PR. 🤔

test/helpers/analysis.spec.ts
MarcL: Updated to mock it and return the logger so we can chain if we needed to. Thanks, @orinokai for the help.

packages/runtime/src/templates/getApiHandler.ts
nickytonline: Yeah, the new logger should be just for build time, no runtime code for the reason you mention. I do wonder if we can use the system log in handlers we generate?

test/index.spec.ts
nickytonline: I don't think they're necessary in the tests either.

packages/runtime/src/helpers/verification.ts
```diff
@@ -210,7 +213,7 @@ export const warnForProblematicUserRewrites = ({
 
 export const warnForRootRedirects = ({ appDir }: { appDir: string }) => {
   if (existsSync(join(appDir, '_redirects'))) {
-    console.log(
+    logger.info(
       yellowBright(
```
nickytonline, line 217: do we need the colours still, or does winston provide different colours for different log levels?

packages/runtime/src/templates/getHandler.ts
```diff
@@ -1,3 +1,6 @@
+// This is a runtime function so assuming we don't want to add
```
nickytonline, line 1: double check, but I think you might be able to exclude the whole packages/runtime/src/templates/ folder from this eslint rule.

nickytonline (commented):
This is looking great @MarcL! Left some questions/comments.

packages/runtime/src/helpers/verification.ts
MarcL: Good idea. It does allow you to colour them but we'd have to add an additional config.

https://github.com/winstonjs/winston#colorizing-standard-logging-levels

Should we do this and remove our use of `chalk` or do we want to leave it as is so it matches the current behaviour?

packages/runtime/src/templates/getHandler.ts
MarcL: Great. I'll check this.

MarcL: Can you do this for a single rule @nickytonline?
I can see how you exclude linting a set of files but not for a single rule.

nickytonline: You can create overrides with globs. e.g.

https://github.com/netlify/next-runtime/blob/e1a3dc0a79a1ce388af306d0788077670507dd08/demos/nx-next-monorepo-demo/apps/demo-monorepo-e2e/.eslintrc.json#L1-L10

nickytonline: You can do it for a single rule. See https://eslint.org/docs/latest/use/configure/rules#using-configuration-files-1

packages/runtime/src/helpers/verification.ts
nickytonline: Maybe in a small follow up PR once the new logger is up and running. If it's not a big lift, go for it now.

MarcL: I took a look and we'd have to ensure we used the correct log levels. In some cases, we're changing colours but still using `logger.info`. Might be worth shipping this and then updating the log levels if we're considering some of them as different messages.

packages/runtime/src/templates/getHandler.ts
MarcL: Done! 👍🏻

packages/runtime/src/helpers/analysis.ts
```diff
@@ -100,7 +101,7 @@ export const extractConfigFromFile = async (apiFilePath: string, appDir: string)
 
   if (!extractConstValueModulePath || !parseModulePath) {
     if (!hasWarnedAboutNextVersion) {
-      console.log("This version of Next.js doesn't support advanced API routes. Skipping...")
+      logger.log("This version of Next.js doesn't support advanced API routes. Skipping...")
```
pieh, line 104:
```suggestion
      logger.info("This version of Next.js doesn't support advanced API routes. Skipping...")
```

`.info` was used as replacement in other places that replace `console.log`. `logger.log` currently cause TS errors:

```
src/helpers/analysis.ts(104,18): error TS2345: Argument of type 'string' is not assignable to parameter of type 'LogEntry'.
```
as `logger.log` has this signature:

```
  interface LogEntry {
    level: string;
    message: string;
    [optionName: string]: any;
  }

  interface LogMethod {
    (level: string, message: string, callback: LogCallback): Logger;
    (level: string, message: string, meta: any, callback: LogCallback): Logger;
    (level: string, message: string, ...meta: any[]): Logger;
    (entry: LogEntry): Logger;
    (level: string, message: any): Logger;
  }
```

packages/runtime/src/templates/getApiHandler.ts
MarcL: Let's look at that for the next PR. Sounds like a good idea.

PR #2224, opened 2023-07-21
chore: fix tests that rely on split route snapshots

MarcL:
## Description

When running the unit tests with `npm run build:demo && npm run test:jest`, they would fail locally if you haven't run `netlify serve` to generate the redirects.

This PR adds additional tests with the `NEXT_SPLIT_API_ROUTES` environment variable enabled and disabled to capture the different redirect generation behaviour.

## Tests

See changes for the additional tests.

1. `npm run build:demo && npm run test:jest`

image


test/index.spec.ts
```diff
@@ -738,20 +739,45 @@ describe('onBuild()', () => {
...
+
+    it('writes correct redirects to netlifyConfig', async () => {
+      await moveNextDist()
+
+      await nextRuntime.onBuild(defaultArgs)
+      // Not ideal, because it doesn't test precedence, but unfortunately the exact order seems to
+      // be non-deterministic, as it depends on filesystem globbing across platforms.
+      const sorted = [...netlifyConfig.redirects].sort((a, b) => a.from.localeCompare(b.from))
+
+      console.log(sorted)
```
pieh, line 779: should this `console.log` stay?

test/index.spec.ts
```diff
@@ -364,13 +364,14 @@ describe('onBuild()', () => {
     expect(existsSync(`.netlify/functions-internal/___netlify-odb-handler/handlerUtils.js`)).toBeTruthy()
   })
 
-  it('writes correct redirects to netlifyConfig', async () => {
+  it('when splitting API routes is disabled, it writes correct redirects to netlifyConfig', async () => {
```
pieh, line 367: this should set env var to `false`

currently Github Actions will have env var set so this test will actually run `onBuild` with API routes enabled

MarcL: 👍🏻

MarcL: Done.

test/index.spec.ts
```diff
@@ -738,20 +739,45 @@ describe('onBuild()', () => {
...
 
-    expect(functionsManifest).toEqual({
-      config: {
-        generator: '@netlify/next-runtime@unknown',
-        name: 'Next.js API handler',
-      },
-      version: 1,
+    afterEach(() => {
+      process.env = oldProcessEnv
+    })
```
pieh, line 757: this doesn't actually reset `process.env`, because `process.env` and `oldProcessEnv` is same object that is being mutated. 

You might want to store `process.env.NEXT_SPLIT_API_ROUTES` specifically and restore just that (tho side note if that env var would be undefined, restoring it would set `'undefined'` string because `process.env` is special thing that always coerce to string:

```
> node -e "process.env.WAT = undefined; console.log({ value: process.env.WAT });"
{ value: 'undefined' }
// ^ notice quotes around undefined
```

This won't cause problems because source code use `destr`, just quirk worth knowing about as in other situations could cause a lot of hair pulling

test/__snapshots__/index.spec.ts.snap
```diff
@@ -1326,7 +1326,1606 @@ Array [
...
+  Object {
+    "from": "/api/enterPreview",
+    "status": 200,
+    "to": "/.netlify/functions/___netlify-api-handler",
+  },
+  Object {
+    "from": "/api/exitPreview",
+    "status": 200,
+    "to": "/.netlify/functions/___netlify-api-handler",
+  },
```
pieh, line 2098: This snapshot look suspect to me - it tests SPLIT_API is disabled, but it still create those routes? (not sure if I read it right tho, just wanted to bring attention to it)

test/index.spec.ts
MarcL: Ah damn. Good point.
What about using a spread instead then?

```javascript
const oldProcessEnv = {...process.env}

// test code

process.env = oldProcessEnv
```

test/__snapshots__/index.spec.ts.snap
MarcL: Let me look.

test/index.spec.ts
```diff
@@ -364,14 +364,20 @@ describe('onBuild()', () => {
     expect(existsSync(`.netlify/functions-internal/___netlify-odb-handler/handlerUtils.js`)).toBeTruthy()
   })
 
-  it('writes correct redirects to netlifyConfig', async () => {
+  it('when splitting API routes is disabled, it writes correct redirects to netlifyConfig', async () => {
+    const oldProcessEnv = { ...process.env }
+    process.env.NEXT_SPLIT_API_ROUTES = 'false'
```
MarcL, line 369: It would help the snapshots if this was set to `false` and not `true` @pieh. 🤦🏻

test/__snapshots__/index.spec.ts.snap
MarcL: It helps if you set the env var to `false` and not `true`. Check that the snapshot seems sensible now. Thanks!

test/index.spec.ts
pieh: I missed that too lol 🤦

pieh (approved):
Thank you!

PR #2233, opened 2023-07-26
fix: ISR 404 pages should be cached ephemerally

orinokai:
## Description

In Next.js it's possible to make an ISR 404 page that revalidates after a period of time.

A 404 page will be served in a number of scenarios, including (but not limited to):
* a non-existing route, e.g. /blahblah
* an SSR route that returns `notFound: true`
* a non-prerendered path on a dynamic route with `fallback: false`, e.g. /posts/3 on the dynamic route /posts/[id] with `id` 1 and 2 prerendered and `fallback: false`
* a path on a dynamic route that returns `notFound: true`, e.g. /posts/4 on the dynamic route /posts/[id] where `id` 4 matches a condition that returns `notFound: true`

Currently, on Netlify, an ISR 404 page will behave like an SSR page in the first 2 scenarios, a static page in the 3rd scenario and an ISR page without a TTL in the last scenario (which is particularly problematic because the `notFound` logic might change due to some external state changing, e.g. CMS content, but the 404 will remain indefinitely and the route will never be revalidated).

These behaviors occur for the following reasons:
1. currently we can only revalidate a route that is detected as ISR at build time, therefore a non-existing route cannot be routed to the ODB handler
2. dynamic routes with `fallback: false` are only intended to serve prerendered paths on the route, therefore we currently redirect the entire route to the statically rendered 404 page on the CDN (relying on the CDN to serve the prerendered HTML pages instead of following the redirect)
3. dynamic routes that return `notFound: true` will serve an ISR 404 page, but Next.js does not use `stale-while-revalidate` headers in this case, unlike other ISR content, and this prevents us from identifying the response as ISR and extracting the TTL

This PR adds a special case to the TTL logic and assumes that if the route is an ODB route and the status code is 404, then we should only cache the response ephemerally for 60s. It is a stopgap solution designed to limit the impact of #3 above. With the PR, an ISR 404 will no longer be cached indefinitely, however it will still not respect that actual TTL specified in the `revalidate` value. This PR does not address #1 and #2 above, which will be tackled as part of forthcoming changes to the caching implementation.

## Tests

[TODO] Tests added

### DIFF
--- a/packages/runtime/src/templates/getHandler.ts
+++ b/packages/runtime/src/templates/getHandler.ts
@@ -163,11 +163,6 @@ const makeHandler = ({ conf, app, pageRoot, NextServer, staticManifest = [], mod
     // Sending SWR headers causes undefined behaviour with the Netlify CDN
     const cacheHeader = multiValueHeaders['cache-control']?.[0]
 
-    // special case for ISR 404s
-    if (result.statusCode === 404 && requestMode === 'odb') {
-      result.ttl = 60
-    }
-
     if (cacheHeader?.includes('stale-while-revalidate')) {
       if (requestMode === 'odb') {
         const ttl = getMaxAge(cacheHeader)
@@ -183,6 +178,11 @@ const makeHandler = ({ conf, app, pageRoot, NextServer, staticManifest = [], mod
         }
       }
       multiValueHeaders['cache-control'] = ['public, max-age=0, must-revalidate']
+    }
+
+    // ISR 404s are not served with SWR headers so we need to set the TTL here
+    if (requestMode === 'odb' && result.statusCode === 404) {
+      result.ttl = 60
     }
 
     if (result.ttl > 0) {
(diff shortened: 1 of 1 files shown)


packages/runtime/src/templates/getHandler.ts
```diff
@@ -176,12 +176,19 @@ const makeHandler = ({ conf, app, pageRoot, NextServer, staticManifest = [], mod
...
         }
-        if (result.ttl > 0) {
-          requestMode = `odb ttl=${result.ttl}`
-        }
       }
       multiValueHeaders['cache-control'] = ['public, max-age=0, must-revalidate']
     }
+
+    // ISR 404s are not served with SWR headers so we need to set the TTL here
+    if (requestMode === 'odb' && result.statusCode === 404) {
```
MarcL, line 184: [dust] If we can never have a 404 with an `swr` header then we don't need the 404 in `ephemeralCodes` above.

MarcL (approved):
This looks like a solid quick fix until we have proper cache headers. Thanks!

PR #2236, opened 2023-07-28
fix: improves support for __NEXT_PRIVATE_PREBUNDLED_REACT

nickytonline:
## Description

Improves support for setting `__NEXT_PRIVATE_PREBUNDLED_REACT`. We have an implementation that detects when to set `__NEXT_PRIVATE_PREBUNDLED_REACT` to `experimental` but there are edge cases. The fix is a naive approach where `__NEXT_PRIVATE_PREBUNDLED_REACT` is set to `experimental` if the request fails, and we reattempt the request.

### Documentation

Relates to https://github.com/netlify/pod-ecosystem-frameworks/issues/489

## Tests

1. TODO


packages/runtime/src/templates/server.ts
```diff
@@ -23,6 +23,10 @@ const getNetlifyNextServer = (NextServer: NextServerType) => {
     private netlifyConfig: NetlifyConfig
     private netlifyPrerenderManifest: PrerenderManifest
 
+    public getAppRouterReactVersion(): string {
```
nickytonline, line 26: Added this public method so we can get the React version to set `process.env.__NEXT_PRIVATE_PREBUNDLED_REACT` if the request fails with `[ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './ server.edge' is not defined`

taty2010 (dismissed):

nickytonline:
Gonna revisit this on Monday. I think I need to rerun the override hooks for this fix.

orinokai (approved):
Nice sleuthing, @taty2010 and thanks for all your work prior @nickytonline! Looks great to me.

PR #2249, opened 2023-08-08
fix: don't ever execute middleware in lambda

pieh:
## Description

This change make `next-server`'s `runMiddleware` no-op in lambdas. There are some scenarios where `next-server` would attempt that (example one seems to be rendering page in preview mode). The middleware is executed at the edge so executing it at lambda would cause:
 - it to run twice causing unknown effects
 - rather unknown behavior and support for various middleware APIs (geo? text rewrite)
 - crash when using `@netlify/next` enhanced middleware as that would execute in Nodejs env which cause early bail because it's not Deno env.
 - when SPLIT_* (api / render ) mode is enabled in next runtime we would not bundle required files to run middleware (common missing module is `.next/server/edge-runtime-webpack.js` and middleware module itself) - we could bundle those but because of all of above it seems better to not even try to run middleware at lambda.

Making `runMiddleware` always no-op in lambda seems safe for now as Next middleware is supposed to run only at edge, but https://github.com/vercel/next.js/discussions/46722 is a feature request asking to make middleware environment configurable, but at least as of now there was no reply from Next.js maintainers 

`next-server` in general does have capability to run in Nodejs env as that's what is used locally (`next start`) through Next.js sandbox environment ( https://github.com/vercel/next.js/tree/canary/packages/next/src/server/web/sandbox )

It does seem like there is some difference in behavior between recent next versions (I could reproduce missing module problem with `next@13.2.4`, but it didn't seem to happen with `next@13.4.13`), but even if things don't crash, it still seems valid to not run middleware in lambda ever if it runs at edge already.

## Tests

Added cypress tests for middleware demo site

https://github.com/netlify/pod-ecosystem-frameworks/issues/588

### DIFF
--- a/cypress/e2e/middleware/preview.cy.ts
+++ b/cypress/e2e/middleware/preview.cy.ts
@@ -1,0 +1,34 @@
+describe('Preview Mode', () => {
+  it('enters and exits preview mode', () => {
+    Cypress.Cookies.debug(true)
+    cy.getCookies().then((cookie) => cy.log('cookies', cookie))
+
+    cy.intercept('/previewTest', (req) => {
+      req.continue((res) => {
+        expect(res.headers?.['x-middleware-executed']).to.equal('true')
+      })
+    }).as('previewTestVisit')
+
+    // preview mode is off by default
+    cy.visit('/previewTest')
+    cy.findByText('Is preview? No', { selector: 'h1' })
+
+    // enter preview mode
+    cy.request('/api/enterPreview').then((response) => {
+      expect(response.body).to.have.property('name', 'preview mode')
+    })
+
+    // exptected content is rendered
+    cy.visit('/previewTest')
+    cy.findByText('Is preview? Yes!', { selector: 'h1' })
+
+    // exit preview mode
+    cy.request('/api/exitPreview')
+    cy.visit('/previewTest')
+    cy.findByText('Is preview? No', { selector: 'h1' })
+
+    // we should hit /previewTest 3 times (before entering preview, after entering preview, after exiting preview)
+    // this assertion is mainly to ensure interception works and assertion on response header is made
+    cy.get('@previewTestVisit.all').should('have.length', 3)
+  })
+})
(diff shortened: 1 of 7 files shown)


cypress/e2e/middleware/preview.cy.ts
```diff
@@ -0,0 +1,34 @@
...
+    // exit preview mode
+    cy.request('/api/exitPreview')
+    cy.visit('/previewTest')
+    cy.findByText('Is preview? No', { selector: 'h1' })
+
+    // we should hit /previewTest 3 times (before entering preview, after entering preview, after exiting preview)
+    // this assertion is mainly to ensure interception works and assertion on response header is made
+    cy.get('@previewTestVisit.all').should('have.length', 3)
+  })
+})
```
pieh, line 34: This test does assert things correctly, but the version of Next used in `middleware` demo actually doesn't have the bug included, so the fix is only relevant to older next versions which we don't run test against currently :/

I think it's still worthwhile to include the test, but we should figure out test setup to run against more versions of next.js

nickytonline: It's good to have this like you mentioned and if we get around to running these tests on older versions, this will definitely be helpful.

packages/runtime/src/templates/server.ts
```diff
@@ -118,6 +119,13 @@ const getNetlifyNextServer = (NextServer: NextServerType) => {
       }
     }
 
+    // eslint-disable-next-line class-methods-use-this, require-await
+    async runMiddleware(): Promise<{ finished: boolean }> {
```
nickytonline, line 123: Just linking to Next.js server's implementation for reference. https://github.com/vercel/next.js/blob/1286e145b058538e47b11cc1ec813a686cee1f68/packages/next/src/server/next-server.ts#L1474-L1589

nickytonline (approved):
Thanks for sorting this out @pieh! :shipit:

PR #2266, opened 2023-08-14
refactor: migrate demos/default to package path

lukasholzer:
## Tests

1. TODO

### DIFF
--- a/demos/default/netlify.toml
+++ b/demos/default/netlify.toml
@@ -1,7 +1,7 @@
 [build]
-command = "next build"
-publish = ".next"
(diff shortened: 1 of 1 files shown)


demos/default/netlify.toml
```diff
@@ -1,7 +1,7 @@
 [build]
-command = "next build"
-publish = ".next"
-ignore = "if [ $CACHED_COMMIT_REF == $COMMIT_REF ]; then (exit 1); else git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF ../..; fi;"
```
lukasholzer, line 4: We actually want to get a diff for the whole repo and not `git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF demos/canary` which would the regular diff for monorepos

serhalp:
Superseded by v5.

PR #2286, opened 2023-08-29
feat: cdn cache control support

pieh:
## Description

This explores using CDN Cache Control headers in `next-runtime`. Those are alternative to ODBs. This Pull Request contains changes from following PRs:
 - [fix: add text/x-component header to .rsc assets generated at build time, don't use rsc-data edge function when hitting .rsc path directly](https://github.com/netlify/next-runtime/pull/2290)
 - [test: setup cypress test running for variant of default demo with all flags enabled](https://github.com/netlify/next-runtime/pull/2289)

So it might be worth checking only last commit in this PR for just CDN Cache Control related changes

Things that are done here:
 - introduction of feature flag for cdn cache control
 - if that flag is enabled:
    - don't generate ODB handler at all, and all rewrites that previously would point to odb now point to ssr handler as it handle both SSR and ISR pages in this setup
    - setting up `netlify-cdn-cache-control` header to re-use value of `cache-control` header provided by Next.js, setting `x-nf-vary` from `vary` header provided by Next.js (syntax of value of those headers is different, so there is some manipulation of value going on)
    - `rsc-data` middleware is no longer used for RSC request if it is handled by lambda (we still need it for .rsc files that are produced at build time - we would need header conditions for rewrites to be able to get rid of it completely)

Things not yet done:
 - testing behavior related to i18n
 - implementation of on-demand isr revalidation (required APIs are not yet available)

## Tests

1. TODO

### DIFF
--- a/cypress/e2e/default/dynamic-routes.cy.ts
+++ b/cypress/e2e/default/dynamic-routes.cy.ts
@@ -16,11 +16,32 @@ describe('Static Routing', () => {
       expect(res.body).to.contain('Dancing with the Stars')
     })
   })
+  it('renders correct page via ODB on a static route (root)', () => {
+    cy.request({ url: '/', headers: { 'x-nf-debug-logging': '1' } }).then((res) => {
+      console.log(res.headers)
+      expect(res.status).to.eq(200)
+      if (CDNCacheControlEnabled) {
+        expect(res.headers).to.have.property('netlify-cdn-cache-control', 's-maxage=60, stale-while-revalidate')
+        expect(res.headers).to.have.property(
+          'netlify-vary',
+          'header=Accept-Encoding,language=en|es|fr,cookie=__prerender_bypass|__next_preview_data|NEXT_LOCALE',
+        )
+      } else {
+        expect(res.headers).to.have.property('x-nf-render-mode', 'odb ttl=60')
+      }
+      expect(res.body).to.contain('NextJS on Netlify')
+    })
+  })
+
   it('renders correct page via ODB on a static route', () => {
     cy.request({ url: '/getStaticProps/with-revalidate/', headers: { 'x-nf-debug-logging': '1' } }).then((res) => {
       expect(res.status).to.eq(200)
       if (CDNCacheControlEnabled) {
         expect(res.headers).to.have.property('netlify-cdn-cache-control', 's-maxage=1, stale-while-revalidate')
+        expect(res.headers).to.have.property(
+          'netlify-vary',
+          'header=Accept-Encoding,cookie=__prerender_bypass|__next_preview_data',
+        )
       } else {
         expect(res.headers).to.have.property('x-nf-render-mode', 'odb ttl=60')
       }
@@ -72,6 +93,10 @@ describe('Dynamic Routing', () => {
       expect(res.status).to.eq(200)
       if (CDNCacheControlEnabled) {
         expect(res.headers).to.have.property('netlify-cdn-cache-control', 's-maxage=31536000, stale-while-revalidate')
+        expect(res.headers).to.have.property(
+          'netlify-vary',
+          'header=Accept-Encoding,cookie=__prerender_bypass|__next_preview_data',
+        )
       } else {
         // expect 'odb' until https://github.com/netlify/pillar-runtime/issues/438 is fixed
         expect(res.headers).to.have.property('x-nf-render-mode', 'odb')
@@ -95,6 +120,10 @@ describe('Dynamic Routing', () => {
         expect(res.status).to.eq(200)
         if (CDNCacheControlEnabled) {
           expect(res.headers).to.have.property('netlify-cdn-cache-control', 's-maxage=31536000, stale-while-revalidate')
+          expect(res.headers).to.have.property(
+            'netlify-vary',
+            'header=Accept-Encoding,cookie=__prerender_bypass|__next_preview_data',
+          )
         } else {
           expect(res.headers).to.have.property('x-nf-render-mode', 'odb')
         }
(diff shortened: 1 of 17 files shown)


cypress/e2e/default/dynamic-routes.cy.ts
```diff
@@ -17,7 +19,11 @@ describe('Static Routing', () => {
   it('renders correct page via ODB on a static route', () => {
     cy.request({ url: '/getStaticProps/with-revalidate/', headers: { 'x-nf-debug-logging': '1' } }).then((res) => {
       expect(res.status).to.eq(200)
-      expect(res.headers).to.have.property('x-nf-render-mode', 'odb ttl=60')
+      if (CDNCacheControlEnabled) {
+        expect(res.headers).to.have.property('netlify-cdn-cache-control', 's-maxage=1, stale-while-revalidate')
+      } else {
+        expect(res.headers).to.have.property('x-nf-render-mode', 'odb ttl=60')
+      }
```
MarcL, line 26: I don't like test expectations that have logic in them as I feel it's harder to parse. What do you think about having a separate test suite that uses the same tests but enables the feature flag and this one disables it?

packages/runtime/src/templates/edge-shared/rsc-data.ts
```diff
@@ -28,10 +28,9 @@ export declare type PrerenderManifest = {
 
 const noop = () => {}
 
-// Ensure that routes with and without a trailing slash map to different ODB paths
 const rscifyPath = (route: string) => {
   if (route.endsWith('/')) {
-    return route.slice(0, -1) + '.rsc/'
+    return route.slice(0, -1) + '.rsc'
```
MarcL, line 33: Have we got a test for this? Do we know why we had the trailing slash in the first place?

packages/runtime/src/templates/getHandler.ts
```diff
@@ -160,32 +169,49 @@ const makeHandler = ({ conf, app, pageRoot, NextServer, staticManifest = [], mod
     }
 
     // Sending SWR headers causes undefined behaviour with the Netlify CDN
```
MarcL, line 171: Worth deleting this comment as it probably doesn't hold true now?

test/helpers/edge.spec.ts
```diff
@@ -46,12 +46,6 @@ describe('generateRscDataEdgeManifest', () => {
         name: 'RSC data routing',
         path: '/',
       },
-      {
-        function: 'rsc-data',
-        generator: '@netlify/next-runtime@1.0.0',
-        name: 'RSC data routing',
-        path: '/index.rsc',
-      },
```
MarcL, line 54: Is this removed due to the cache control? Just checking it needs to be deleted for both the feature flag on and off.

test/helpers/edge.spec.ts
```diff
@@ -99,12 +93,6 @@ describe('generateRscDataEdgeManifest', () => {
         name: 'RSC data routing',
         pattern: '^/blog/([^/]+?)(?:/)?$',
       },
-      {
-        function: 'rsc-data',
-        generator: '@netlify/next-runtime@1.0.0',
-        name: 'RSC data routing',
-        pattern: '^/blog/([^/]+?)\\.rsc$',
-      },
```
MarcL, line 107: Is this removed due to the cache control? Just checking it needs to be deleted for both the feature flag on and off.

packages/runtime/src/templates/edge-shared/rsc-data.ts
pieh: We do have existing tests for rsc in https://github.com/netlify/next-runtime/blob/main/cypress/e2e/default/appdir.cy.ts 

Overall I did lift RSC related changes to separate standalone PR ( https://github.com/netlify/next-runtime/pull/2290 ) where we can focus on them - I only hit some problems with RSC in this work after I made changes to not use RSC rewrites edge middleware to lambdas (as we finally could vary on RSC header there) which exposed some problems with it

packages/runtime/src/templates/getHandler.ts
```diff
@@ -160,32 +169,49 @@ const makeHandler = ({ conf, app, pageRoot, NextServer, staticManifest = [], mod
...
-        const ttl = getMaxAge(cacheHeader)
-        // Long-expiry TTL is basically no TTL, so we'll skip it
-        if (ttl > 0 && ttl < ONE_YEAR_IN_SECONDS) {
-          // ODBs currently have a minimum TTL of 60 seconds
-          result.ttl = Math.max(ttl, 60)
+    const cacheControlHeader = multiValueHeaders['cache-control']?.[0]
+
+    if (useCDNCacheControl) {
+      if (cacheControlHeader?.includes('stale-while-revalidate')) {
+        multiValueHeaders[`Netlify-Cdn-Cache-Control`] = [cacheControlHeader]
```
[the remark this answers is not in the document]
pieh: It seems to work with this casing, but I can try lowercasing it to keep header names style consistent in our code if possible

packages/runtime/src/templates/getHandler.ts
pieh: Probably worth updating or maybe revisiting. As it's not just CDN behavior that it would affect - it would also affect browser cache behavior - if we had both SWR on CDN and in browser, it would mean that content that CDN sends back that might already be stale (while it starts to revalidate it in background) would be fresh for browser and browser would potentially cache response it sees as fresh but might actually already be stale (in CDN terms)

test/helpers/edge.spec.ts
pieh: More context on those in https://github.com/netlify/next-runtime/pull/2290 which has RSC changes from this PR extracted to standalone change

cypress/e2e/default/dynamic-routes.cy.ts
pieh: Moving them to separate file make keeping things in sync harder and more manual work on trying to keep them in sync  (i.e. if we add new test case, we should add to both, but if you add to just one of them, nothing will signal that things are not in sync anymore).

While I'm not huge fan of doing what I did, I think it's realistically least bad option if we want have (and maintain) parity on what behaviors we test

packages/runtime/src/templates/getHandler.ts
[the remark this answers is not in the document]
pieh: I did change to lowercases everywhere and it works, so I'll keep it lowercased (and resolve that comment)

cypress/e2e/default/dynamic-routes.cy.ts
MarcL: Makes sense. Thanks.

serhalp:
Superseded by v5.

PR #2287, opened 2023-08-29
feat: add blob storage

nickytonline:
## Description

Uses Netlify blob storage for prerendered content instead of storing prerendered content in the generated lambdas, assuming blob storage is configured for the site.

There are 2 distinct and somewhat separate changes here:
 1. (minimal code changes) we don't force fresh/revalidated content on first ODB request anymore (we use `x-prerender-revalidate` header to force revalidation) - this applies also when blob storage is not used (then it will use files from lambda instead of blob storage)
    - note here - this change makes `fallback: true` pages behavior really bad, because ODB would cache those fallback (think Skeleton) so for users they would always first see skeleton page before browser client would re-render with actual content (if navigated to a page directly, and not using client navigation like with `@next/link`).
    
      So to address that this also makes `fallback: true` behave more like `fallback: blocking` (which is behavior we do have today, so we are not changing it really)
 2. Use blob storage for ISR assets generated at build time:
    - we compile ESM-only `@netlify/blobs` to CJS via `build:blob` script - runtime is CJS and although there are tricks and hacks to still be able to use ESM version there are a lot of problems with them (`await import` is compiled to `require` using our current `tsconfig`, any `eval` tricks to avoid that later make lambda bundling problematic etc)
    - we check a feature flag (`next-runtime-use-blobs-for-isr-assets`) and if that is enabled we also test for availability of blob store to determine wether to use it (there are no additional feature flags)
    - if blob store is available:
      - (at build time) when we iterate over generated files after Next build, if file is connected to ISR page we upload the file content to blob storage and delete original file (currently we just keep the file as-is and those are later bundled into lambdas)
      - (at lambda request time) `fs.readFIle` will download a file from blob store (if the file to read was uploaded at build time). We already had mechanism like that that would pull files from CDN for SSG assets (like for example `next-server` deciding to serve 404 page) - that mechanism now can also pull from blob store as well.

## Tests

1. TODO

- https://www.notion.so/netlify/IncrementalCache-implementation-for-Next-js-runtime-1d2c12a7e0e04e698f52bcb955d3bcca?pvs=9
- https://github.com/netlify/next-runtime/pull/2288

## Fixes

Fixes https://github.com/netlify/pod-ecosystem-frameworks/issues/568
Fixes FRA-34


packages/runtime/src/helpers/functions.ts
```diff
@@ -364,58 +369,154 @@ const changeExtension = (file: string, extension: string) => {
   return join(dirname(file), base + extension)
 }
 
-const getSSRDependencies = async (publish: string): Promise<string[]> => {
+const getPrerenderManifest = async (publish: string) => {
```
nickytonline, line 372: I extracted getting the prerenderManifest out of the prerender content functions so one these could be more testable potentially, but also, the job of those functions are more to iterated through the content of the prerendermanifest, not load it.

packages/runtime/src/helpers/functions.ts
```diff
@@ -364,58 +369,154 @@ const changeExtension = (file: string, extension: string) => {
...
 
-  return [
-    ...Object.entries(prerenderManifest.routes).flatMap(([route, ssgRoute]) => {
-      if (ssgRoute.initialRevalidateSeconds === false) {
-        return []
-      }
+  return prerenderManifest
+}
+
+type CachedPageValue = Extract<ResponseCacheValue, { kind: 'PAGE' }> | { html: object | string }
```
nickytonline, line 378: `ResponseCacheValue` is an exported union type from Next.js, however it doesn't export the types in the union. I use extract to get the actual type I need.

packages/runtime/src/helpers/functions.ts
```diff
@@ -364,58 +369,154 @@ const changeExtension = (file: string, extension: string) => {
...
+type PageCacheData = {
+  value: CachedPageValue
+  lastModified: number
+}
+
+const getPrerenderedBlobStoreContent = async (
+  prerenderManifest: PrerenderManifest,
+  publish: string,
+): Promise<Array<{ key: string; data: PageCacheData }>> => {
+  const limit = pLimit(Math.max(2, cpus().length))
```
nickytonline, line 388: Going with the same logic used in https://github.com/netlify/next-runtime/blob/a30ffbace64ffb447c9a8c118cc1faac63cd8db9/packages/runtime/src/helpers/files.ts#L153

packages/runtime/src/helpers/functions.ts
```diff
@@ -364,58 +369,154 @@ const changeExtension = (file: string, extension: string) => {
...
   const commonDependencies = await getCommonDependencies(publish)
   const ssrRoutes = await getSSRRoutes(publish)
 
   // TODO: for now, they're the same - but we should separate them
   const nonOdbRoutes = ssrRoutes
   const odbRoutes = ssrRoutes
+  const { NETLIFY_API_HOST, NETLIFY_API_TOKEN, SITE_ID } = constants
+
+  // This check could be improved via featureFlags exposed in the build for blob storage
+  const isUsingBlobStorage =
```
nickytonline, line 468: I added checks for the deploy ID because blob storage will not be available if you run only `ntl build`, but it will be available once @eduardoboucas makes some changes for blob storage to be supported via `ntl deploy`.

packages/runtime/src/helpers/blobStorage.ts
```diff
@@ -0,0 +1,34 @@
...
+  token,
+  siteID,
+  deployId,
+}: {
+  apiHost: string
+  token: string
+  siteID: string
+  deployId: string
+}): Promise<Blobs> => {
+  // eslint-disable-next-line no-new-func
```
nickytonline, line 16: This is a workaround to allow us to use `@netlify/blobs` which is currently ESM only. Hat tip to @pieh for the hack.

nickytonline:
When iterating through prerendered content, this error seems to happen about every 1 out of 3 deploys

```
3:13:59 PM:   Error: ENOENT: no such file or directory, open '/opt/build/repo/demos/default/.next/server/pages/en/getStaticProps/with-revalidate-404.json'
```

I'm not sure why it isn't generated sometimes, but that is something to look into. Here's an example build failure.https://app.netlify.com/sites/netlify-plugin-nextjs-demo/deploys/64fb72486da6ec00084e4573

packages/runtime/src/helpers/config.ts
```diff
@@ -1,4 +1,4 @@
-import type { NetlifyConfig } from '@netlify/build'
+import type { NetlifyConfig } from '@netlify/build/types'
```
nickytonline, line 1: I don't know why, but for `@netlify/build/29.20.15` I had to import from the types folder directly even though the editor/TS language server was able to resolve the types from `@netlify/build`. I was even able to go to definition no problem. IT was only when I compiled that it would complain. Setting it to `@netlify/build/types` made tsc happy.

packages/runtime/src/templates/getCacheHandler.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import { copyFile, writeFile } from 'fs-extra'
+import { outdent as javascript } from 'outdent'
+
+export const generateCacheHandler = async (publish: string): Promise<void> => {
+  const blobCache = getCacheHandler()
+
+  await writeFile(join(publish, 'netlify-incremental-cache.js'), blobCache)
+  // node_modules is relative to the root of the project or monorepo
+  await copyFile(
```
nickytonline, line 11: This file copy works locally, but the `process.cwd` is the `/Users/.../packages/next-runtime` folder. Just getting this sorted as it should be relative to the project root or monorepo root.

packages/runtime/src/templates/getCacheHandler.ts
```diff
@@ -0,0 +1,48 @@
...
+    const blobAuthOptions = globalThis.blobContext
+
+    console.log('loaded incremental cache handler')
+
+    module.exports = class CacheHandler {
+        async get(key) {
+          // getBlobStorage is memoized. See below for converstation about getBlobStorage
+          const netliBlob = await getBlobStorage(blobAuthOptions)
+
+          return await netliBlob.get(key.slice(1))
```
nickytonline, line 36: We're stripping starting slashes in keys atm because of https://github.com/netlify/cli/issues/5988

packages/runtime/src/helpers/config.ts
```diff
@@ -34,18 +34,52 @@ const defaultFailBuild = (message: string, { error }): never => {
...
+    const requiredServerFiles: RequiredServerFiles = await readJSON(join(publish, 'required-server-files.json'))
+    const { config, appDir, ignore } = requiredServerFiles
+
     if (!config) {
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-ignore
       return failBuild('Error loading your Next config')
     }
 
+    if ('incrementalCacheHandlerPath' in config.experimental) {
```
nickytonline, line 68: It's unlikely someone deploying to Netlify would add a custom incremental cache handler, but I added this warning in case.

packages/runtime/package.json
```diff
@@ -11,7 +11,8 @@
     "manifest.yml"
   ],
   "dependencies": {
-    "@netlify/esbuild": "0.14.39",
+    "@netlify/blobs": "^1.6.1",
+    "@netlify/esbuild": "0.14.39-1",
```
lukasholzer, line 15: TODO: THis fork is already deprecated we should use esbuild directly! (follow up PR)

cypress/e2e/default/dynamic-routes.cy.ts
```diff
@@ -61,12 +61,13 @@ describe('Dynamic Routing', () => {
...
+    // unfortunately there is a problem with `fallback: true` in ODB context - the fallback would be cached indefinitely
+    // so visits to those pages would always render the fallback first and browser client would later re-render with correct
+    // content. As this is not ideal the `fallback: true` is treated as `fallback: blocking`.
     cy.request({ url: '/getStaticProps/withFallback/3/', headers: { 'x-nf-debug-logging': '1' } }).then((res) => {
       expect(res.status).to.eq(200)
-      // expect 'odb' until https://github.com/netlify/pillar-runtime/issues/438 is fixed
       expect(res.headers).to.have.property('x-nf-render-mode', 'odb')
-      // expect 'Bitten' until the above is fixed and we can test for fallback 'Loading...' message
       expect(res.body).to.contain('Bitten')
     })
```
pieh, line 72: This is actually not changing any assertions, just update the comment why the assertions will stay as they are. Visiting https://651d4678cf52d90008b40118--netlify-plugin-nextjs-demo.netlify.app/getStaticProps/withRevalidate/withFallback/3/ (older preview) shows the problematic behavior (that fallback that flashes before browser client update) - you can also just check with JS disabled in browser and then you never get to see actual proper content

packages/runtime/blob-cjs-build.mjs
```diff
@@ -0,0 +1,20 @@
...
+}
+
+await esbuild.build({
+  entryPoints: ['src/templates/netliblob.mts'],
+  format: 'cjs',
+  bundle: true,
+  outfile: 'src/templates/blob.js',
+  platform: 'node',
+  plugins: [stripNodeColonPlugin],
+})
```
pieh, line 20: this compiles ESM-only `@netlify/blobs` to CJS and removes `node:` prefixes in import/requires (like from `node:fs`) to make resulting code compatible with our minimal node version

packages/runtime/src/templates/getHandler.ts
```diff
@@ -140,6 +146,19 @@ const makeHandler = ({ conf, app, pageRoot, NextServer, staticManifest = [], mod
...
+      const rawData = Buffer.from(context.clientContext.custom.blobs, 'base64')
+      const data = JSON.parse(rawData.toString('ascii'))
+      setBlobInit({
+        authentication: {
+          contextURL: data.url,
+          token: data.token,
+        },
+        context: `deploy:${event.headers['x-nf-deploy-id']}`,
+        siteID: event.headers['x-nf-site-id'],
+      })
```
pieh, line 159: We use Blob Storage in a place that doesn't have direct access to `event`/`context` which is needed to initialize the store, so instead we globally set init argument on each request. This means it's possible that requests would use init data we got from another one (depending on timing), but this is acceptable/ok to do and only note is that there is expiration time on token so that's why this being refreshed on each request and not just once.

test/e2e/next-test-lib/fetch-polyfill.js
```diff
@@ -0,0 +1,8 @@
+if (!globalThis.fetch) {
+  const fetch = require('node-fetch')
+
+  globalThis.fetch = fetch
+  globalThis.Headers = fetch.Headers
+  globalThis.Request = fetch.Request
+  globalThis.Response = fetch.Response
+}
```
pieh, line 8: this is used just in Next's e2e tests setup, everywhere else there is global fetch (`@netlify/blobs` require `fetch` to be either passed in or to be globally available)

packages/runtime/src/helpers/files.ts
```diff
@@ -1,8 +1,9 @@
 import { cpus } from 'os'
 
+import type { Blobs } from '@netlify/blobs/dist/src/main'
```
ascorbic, line 3: Why can't this be imported from the package root?

pieh: Addressed in https://github.com/netlify/next-runtime/pull/2287/commits/239e0443833e1fd4ba9a46d69ebae9c045099ba1

packages/runtime/src/templates/blobStorage.ts
```diff
@@ -0,0 +1,33 @@
...
+let blobInit: BlobsInit
+export const setBlobInit = (init: BlobsInit): void => {
+  blobInit = init
+}
+export const getBlobInit = (): BlobsInit => blobInit
+
+/**
+ * @netlify/blobs ATM has some limitation to keys, so we need to normalize it for now (they will be resolved so we will be able to remove this code)
+ */
+export const getNormalizedBlobKey = (key: string): string => Buffer.from(key).toString('base64url')
```
ascorbic, line 31: Is this fixed now? I know soemthing about encoding keys has landed

ascorbic (approved):

packages/runtime/src/templates/blobStorage.ts
pieh: It's not yet fully fixed - some fixes did land, but for now result is that keys starting with `/` are no longer throwing, but trying to read those keys result in `null` values, so we still need this for time being

packages/runtime/src/templates/server.ts
```diff
@@ -42,6 +42,28 @@ const getNetlifyNextServer = (NextServer: NextServerType) => {
...
+      if (typeof manifest?.dynamicRoutes === 'object') {
+        for (const route of Object.values(manifest.dynamicRoutes)) {
+          // 'fallback' property is:
+          //   - a string when fallback: true is used
+          //   - `null` when fallback: blocking is used
+          //   - `false` when fallback: false is used
+          // `fallback: true` is not working correctly with ODBs
+          // as we will cache fallback html forever, so
+          // we are treating those as `fallback: blocking`
+          // by editing the manifest
```
orinokai, line 57: Thanks for documenting all these!

orinokai (approved):
Wow, this was quite an undertaking! Thanks @pieh, looks great.

nickytonline:
I still get notified since I opened the PR. Great work @pieh!

orinokai (approved):

nickytonline (commented):
:shipit:

orinokai (dismissed):
All looks dandy 🚢

MarcL (approved):
:shipit:

PR #2288, opened 2023-08-31
feat: implemented incremental cache

nickytonline:
## Tests

1. TODO

## Fixes

Fixes https://github.com/netlify/pod-ecosystem-frameworks/issues/568


packages/runtime/src/helpers/blobStorage.ts
```diff
@@ -0,0 +1,34 @@
+import type { Blobs } from '@netlify/blobs/dist/src/main'
```
nickytonline, line 1: This and memoize won't end up being part of this PR as they're already in #2287.

demos/default/next.config.js
```diff
@@ -89,5 +89,6 @@ module.exports = {
   },
   experimental: {
     optimizeCss: false,
+    incrementalCacheHandlerPath: require.resolve('./cache-handler.js'),
```
nickytonline, line 92: This won't be hard-coded. This is hard-coded for the moment for testing purposes. The real implementation will need to generate this and modify the Next.js config, pre-build.

demos/default/cache-handler.js
```diff
@@ -0,0 +1,19 @@
+const cache = new Map()
```
nickytonline, line 1: This uses an in memory cache for the moment for debugging purposes to see key and values in the cache.

nickytonline:
I merged this branch into #2287.

PR #2289, opened 2023-09-01
test: setup cypress test running for variant of default demo with all flags enabled

pieh:
## Description

This is extracted from https://github.com/netlify/next-runtime/pull/2286 to standalone PR to prepare setup in repo to run cypress default demo tests with all upcoming feature flags enabled (as there will be some assertions that need to be conditional or potentially tests / entire describe blocks skipped). As-is it will just do same thing as current default demo and extracting it to separate PR just limits to diff on my PR + make the setup available for blobs store related work (if needed).

New Netlify site was created ( https://app.netlify.com/sites/netlify-plugin-nextjs-demo-all-flags/overview ) that is build from same directory as current default demo site demo. As new feature flags are added - we should be adding those via Environment Variables to new test site in Netlify UI (`netlify.toml` is shared so we can't add them there without making default demo also using all the new feature flags)

## Tests

1. TODO

### DIFF
--- a/.github/workflows/cypress-demo-all-flags.yml
+++ b/.github/workflows/cypress-demo-all-flags.yml
@@ -1,0 +1,73 @@
+name: Run e2e (default demo with all feature flags enabled)
+on:
+  pull_request:
+    types: [opened, synchronize]
+  push:
+    branches:
+      - main
+    paths:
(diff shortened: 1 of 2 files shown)


serhalp:
Superseded by v5.

PR #2290, opened 2023-09-01
fix: add text/x-component header to .rsc assets generated at build time, don't use rsc-data edge function when hitting .rsc path directly

pieh:
## Description

When working on something else I noticed quite the weirdness with how `rsc-data` middleware behaves when it comes to routing:

```
# hitting /blog/erica/first-post
Aug 31, 06:02:32 PM: 01H9659Z info   Is rsc request
Aug 31, 06:02:32 PM: 01H9659Z info   Rewriting to /blog/erica/first-post.rsc
# hitting /blog/erica/first-post/
Aug 31, 06:06:23 PM: 01H965H0 info   Is rsc request
Aug 31, 06:06:23 PM: 01H965H0 info   Rewriting to /blog/erica/first-post.rsc/
# hitting /blog/erica/first-post.rsc
Aug 31, 06:07:18 PM: 01H965JP info   Is rsc request
Aug 31, 06:07:18 PM: 01H965JP info   Rewriting to /blog/erica/first-post.rsc.rsc
```

as `/blog/erica/first-post.rsc` is actually asset produced at build time:
image
so only the first request from logs above one worked correctly. Second one did seem to work, but instead of using static asset on CDN it wen't to lambda to generate it and third one I don't even remember but it definitely seems wrong.

I'm not exactly sure why we make edge middleware to handle requests for `.rsc` paths - when trying that path we should just use path as-is - it will either use static asset from CDN or go to lambda to generate it (and possibly cache with with ODB). I did remove that from this PR as it didn't make sense to me, especially as it was redirecting to `/blog/erica/first-post.rsc.rsc` which seems all kind of wrong thing to do.

The statically produced .rsc files also didn't get expected `text/x-component` header (we only get this header right now when request is handled by lambda), so I also added header rule to ensure those get expected headers.

And finally I'm not sure why the rsc-data would want to rewrite to paths that end with trailing slash (`.rsc/`) - the comment say to differentiate those for ODB, but with no clarity why you want to do that? It also meant that if user hit path with trailing slash it would never use static assets on CDN and would always need to go through lambda

## Tests

1. TODO

### DIFF
--- a/packages/runtime/src/helpers/config.ts
+++ b/packages/runtime/src/helpers/config.ts
@@ -218,6 +218,19 @@ const buildHeader = (buildHeaderParams: BuildHeaderParams) => {
 const sanitizePath = (path: string) => path.replace(/:[^*/]+\*$/, '*')
 
 /**
+ * Sets Content-type header for static RSC assets (text/x-component), so application/octet-stream is not used when serving them
+ * @param netlifyHeaders - Existing headers that are already configured in the Netlify configuration
+ */
+export const addContentTypeHeaderToStaticRSCAssets = (netlifyHeaders: NetlifyHeaders = []) => {
+  netlifyHeaders.push({
+    for: `*.rsc`,
+    values: {
+      'Content-Type': 'text/x-component',
+    },
+  })
+}
+
+/**
  * Persist Next.js custom headers to the Netlify configuration so the headers work with static files
  * See {@link https://nextjs.org/docs/api-reference/next.config.js/headers} for more information on custom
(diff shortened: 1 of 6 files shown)


ascorbic:
I think the reason was because at the time at least Next didn't request `.rsc` files directly. We were adding the `.rsc` to we could either load the static asset, or get a different asset from the ODB as we couldn't vary on the `RSC` header. I dunno which of those have now changed

serhalp:
Superseded by v5.

PR #2302, opened 2023-09-25
fix: vendor deno dependencies

Skn0tt:
## Description

This PR vendors all dependencies of the edge function templates into the `src/templates/vendor` directory, and sets the `functions[*].deno_import_map` directive to use it at build-time. This should shield next-runtime builds against deno.land outages.

## Tests

I didn't add any tests - hoping our existing E2E suite will capture any regressions.

Closes https://github.com/netlify/pod-ecosystem-frameworks/issues/404.


packages/runtime/package.json
```diff
@@ -52,7 +52,9 @@
     "publish:install": "npm ci",
     "publish:test": "cd .. && npm ci && npm test",
     "clean": "rimraf lib dist-types",
-    "build": "tsc",
+    "build": "run-s build:*",
+    "build:vendor": "rimraf src/templates/vendor && deno vendor src/templates/edge/shims.js src/templates/edge-shared/next-utils.ts https://deno.land/x/html_rewriter@v0.1.0-pre.17/index.ts --output=src/templates/vendor",
```
Skn0tt, line 56: Any new files that import from deno.land will need to be listed here.

Skn0tt:
[This](https://github.com/netlify/next-runtime/files/12713643/4c43d9923afb6f0d38710df525e1d55173b233a6b3fafe26e074357ba8486e57.zip) is an eszip file that I created using the changes in this PR. Opening it on https://eszip-viewer.deno.dev/, you can see that it indeed picks up the vendored files from the local directory:

Screenshot 2023-09-25 at 11 40 17

packages/runtime/package.json
MarcL: Is there a reason we add the `html_rewriter` explicit URL rather than `src/templates/edge-shared/utils.ts` package that references it?

Skn0tt: For some reason, Deno isn't able to vendor `https://edge.netlify.com` correctly. The file that contains `html_rewriter` also references `https://edge.netlify.com`, though, so referencing the file will break :/

Skn0tt:
For some reason, this import map breaks the config detection. I'll need to figure out what's going on there before we can merge this. Turning the PR into draft for now.

```cli
Edge Functions bundling                                       
────────────────────────────────────────────────────────────────

Packaging Edge Functions from .netlify/edge-functions directory:
 - next_middleware
 - next_pages_api_edge
Could not load configuration for edge function at '/Users/skn0tt/dev/netlify/next-runtime/demos/middleware/.netlify/edge-functions/next_pages_api_edge/index.ts'
error: Module not found "file:///Users/skn0tt/dev/netlify/next-runtime/demos/middleware/.netlify/edge-functions/vendor/deno.land/std@0.170.0/flags/mod.ts".
    at https://64e8753eae24930008fac6d9--edge.netlify.com/bootstrap/server.ts:1:23
Could not load configuration for edge function at '/Users/skn0tt/dev/netlify/next-runtime/demos/middleware/.netlify/edge-functions/next_middleware/index.ts'
error: Module not found "file:///Users/skn0tt/dev/netlify/next-runtime/demos/middleware/.netlify/edge-functions/vendor/deno.land/std@0.170.0/flags/mod.ts".
    at https://64e8753eae24930008fac6d9--edge.netlify.com/bootstrap/server.ts:1:23
```

packages/runtime/package.json
MarcL: Thanks!

packages/runtime/src/helpers/edge.ts
```diff
@@ -378,6 +378,11 @@ export const writeEdgeFunctions = async ({
   const nextConfig = nextConfigFile.config
   const usesAppDir = nextConfig.experimental?.appDir
 
+  await copy(getEdgeTemplatePath('../vendor'), join(edgeFunctionRoot, 'vendor'))
+  netlifyConfig.functions['*'] = {
+    deno_import_map: join(INTERNAL_EDGE_FUNCTIONS_SRC, 'vendor', 'import_map.json'),
```
MarcL, line 383: This gives a TypeScript error as `deno_import_map` doesn't exist in `FunctionsObject`. Do we need to update the type to support import maps?

Skn0tt: yup, good catch! doing that in https://github.com/netlify/build/pull/5305

pieh:
> For some reason, this import map breaks the config detection. I'll need to figure out what's going on there before we can merge this. Turning the PR into draft for now.
 
```cli
Edge Functions bundling                                       
────────────────────────────────────────────────────────────────

Packaging Edge Functions from .netlify/edge-functions directory:
 - next_middleware
 - next_pages_api_edge
Could not load configuration for edge function at '/Users/skn0tt/dev/netlify/next-runtime/demos/middleware/.netlify/edge-functions/next_pages_api_edge/index.ts'
error: Module not found "file:///Users/skn0tt/dev/netlify/next-runtime/demos/middleware/.netlify/edge-functions/vendor/deno.land/std@0.170.0/flags/mod.ts".
    at https://64e8753eae24930008fac6d9--edge.netlify.com/bootstrap/server.ts:1:23
Could not load configuration for edge function at '/Users/skn0tt/dev/netlify/next-runtime/demos/middleware/.netlify/edge-functions/next_middleware/index.ts'
error: Module not found "file:///Users/skn0tt/dev/netlify/next-runtime/demos/middleware/.netlify/edge-functions/vendor/deno.land/std@0.170.0/flags/mod.ts".
    at https://64e8753eae24930008fac6d9--edge.netlify.com/bootstrap/server.ts:1:23
```

This seems like `deno.land` dep of netlify.edge? ( https://github.com/netlify/edge-functions-bootstrap/blob/721c691484c132849bca4b71339017ff93302e3a/src/bootstrap/server.ts#L1 ) and because we don't vendor that module here, but we point entire `deno.land` to stuff we vendored we just miss that module? Would it be possible to also add netlify edge to list of modules to be vendored? alternatively massage import map to target only modules that runtime uses and not entire `deno.land`?

Skn0tt:
> Would it be possible to also add netlify edge to list of modules to be vendored?

No, because it's a moving target and when we update the version here:

https://github.com/netlify/edge-bundler/blob/d5d427ab74a98d1bcd49e7d84fe227426f74f57d/deno/config.ts#L2

we might also update the stdlib modules under the hood, and then it'll break again since Next-runtime has other modules bundled.

> alternatively massage import map to target only modules that runtime uses and not entire deno.land?

We could implement this using a scoped import map. I've tried that out with https://github.com/netlify/next-runtime/pull/2302/commits/30416b5ec4d08783b358a5966196096c0a2c2c56, but wasn't able to get it working. 

I'm now going down a straight-forward path, and am updating all imports to straight-up import from the `vendor` directory. Not the most beautiful, but workable until we figure out what's breaking the scoped import maps.

packages/runtime/html_rewriter.patch
```diff
@@ -0,0 +1,4 @@
+56c56
+<       fetch(new URL("./vendor/html_rewriter_bg.wasm", import.meta.url).href)
+---
```
Skn0tt, line 3: This looks like we're replacing the from-disk fetch with a HTTP request at runtime here. This is required because `deno vendor` doesn't download the WASM onto disk (for whatever reason),  so fetching this from disk would fail. This was already the case before this PR, but before this file was imported via `https://`, meaning that `import.meta.url` was something like `https://deno.land/x/...` - now that we're accessing this file from disk, it's more like `file://.netlify/edge-functions/vendor...`. Before, with the HTTPS address, `fetch` would fall back to an HTTP request anyways.

So TL;DR: This looks like we're replacing a solid disk-read with a brittle network fetch. In reality, it's always been a brittle network fetch, so this is fine!

pieh: Can we download that wasm file ourselves and manually vendor it? i.e. manually pull from https://github.com/worker-tools/html-rewriter/blob/v0.1.0-pre.17/vendor/html_rewriter_bg.wasm and avoid doing a  patch (it's bit icky to me, but manual download is icky as well, so I guess - pick your poison). This way at least we would avoid this brittle network fetch?

Skn0tt: Tried that, but the problem is *not only* that `deno vendor` doesn't download WASM files - Deno's eszip tool (that we use for bundling) also isn't able to detect the import, so it doesn't bundle the file. This will probably change once https://github.com/tc39/proposal-source-phase-imports lands, but until then the HTTP fetch is the only thing we can do :/

packages/runtime/src/templates/vendor.ts
```diff
@@ -0,0 +1,19 @@
+// this file isn't meant to be imported.
+// it's a list of all external modules that we use,
+// and we vendor those into the `vendor/` directory
+// for consumption in other files.
```
pieh, line 4: Just for future folks so they don't have to figure this out on their own - just short manual / explainer on how this is setup / how to interact:
```suggestion
// for consumption in other files.
// Vendoring happens automatically as part of runtime `build` script.
// You can trigger just vendoring without full package build by running `build:vendor` script.
```

pieh (approved):
:shipit:

PR #2303, opened 2023-09-26
Optimizing Performance: Concurrent Execution of Async Functions with `Promise.all` ✈

sanjaiyan-dev:
Hi, this PR introduces a very minor performance gain by running async functions concurrently using `Promise.all`.

Extremely sorry, if I made any mistakes :(


orinokai:
Hello @sanjaiyan-dev, thank you for your contribution! However, AFAIK, this will not result in a performance optimization because file IO is blocking - so only one file can be moved at a time - and there is nothing else for the process to do while it is awaiting.

sanjaiyan-dev:
> Hello @sanjaiyan-dev, thank you for your contribution! However, AFAIK, this will not result in a performance optimization because file IO is blocking - so only one file can be moved at a time - and there is nothing else for the process to do while it is awaiting.

Hi @orinokai , I understand your concern, but currently, we are moving public files and static pages. With Node.js default 4-thread pool, we can utilize parallel processing to potentially optimize performance. For more information on how code runs in the thread pool, you can refer to this link: https://nodejs.org/en/docs/guides/dont-block-the-event-loop#what-code-runs-on-the-worker-pool

Extremely sorry if I am wrong :)

orinokai:
thanks @sanjaiyan-dev but, even if each file copy/rename is happening in a separate thread, i still think there are hardware filesystem limitations to the performance gain here, but we're actually running a benchmark in a separate project for something similar so i'll let you know

sanjaiyan-dev:
> thanks @sanjaiyan-dev but, even if each file copy/rename is happening in a separate thread, i still think there are hardware filesystem limitations to the performance gain here, but we're actually running a benchmark in a separate project for something similar so i'll let you know

Okay 🤝

serhalp:
Superseded by v5 rewrite.

PR #2304, opened 2023-09-27
chore:  update demos

taty2010:
## Tests

1. TODO


packages/runtime/src/helpers/config.ts
```diff
@@ -164,6 +166,17 @@ export const configureHandlerFunctions = async ({
...
+    try {
+      // on Next 13.5+ there is no longer statically analyzable import to styled-jsx/style
+      // so lambda fails to bundle it. Next require hooks actually try to resolve it
+      // and fail if it is not bundled, so we forcefully add it to lambda.
+
+      // eslint-disable-next-line n/no-unsupported-features/node-builtins
+      const nextRequire = mod.createRequire(require.resolve(`next`))
+      const styledJsxPath = nextRequire.resolve(`styled-jsx/style`)
+      netlifyConfig.functions[functionName].included_files.push(styledJsxPath)
+    } catch {}
```
MarcL, line 178: Do we need to log anything in the `catch` here? Or do we just need it to fail silently for < Next 13.5+ if the module isn't found?

serhalp:
Superseded by v5.

PR #2305, opened 2023-09-28
chore: bump cypress to latest v12

pieh:
## Description

CI is having troubles running cypress currently and I think https://github.com/cypress-io/cypress/issues/27915 is issue we are facing, so this tries to bump cypress to be `>12.15.0` as https://github.com/cypress-io/cypress/issues/27915#issuecomment-1737288722 suggests 🤞 (there is version 13 already out, but https://github.com/netlify/next-runtime/pull/2296 which bumps to it shows failing tests, so at least some migration would be needed - I just bump to latest of same major we are already on)

## Tests

1. TODO

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,7 +40,7 @@
         "chance": "^1.1.8",
         "cheerio": "^1.0.0-rc.12",
         "cpy": "^8.1.2",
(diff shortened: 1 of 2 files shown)


orinokai (approved):
🚢

PR #2306, opened 2023-09-29
[do-not-merge] blob test variants

pieh:
## Description

This is temporary draft PR just to get deploys that contains https://github.com/netlify/next-runtime/pull/2287 + some default demo config changes to ensure base PR handles all the cases and to not litter main PR git history with those tests

## Tests

1. TODO


pieh:
This served it's purpose with testing variants, closing.

PR #2313, opened 2023-10-05
fix: improve support for new requireHooks update

taty2010:
## Description

With the release of Next 13.5+ there were some major updates done within require-hooks along with the removal of appDir from the experimental config. 

This fix should:
- Only run our requireHooks when Next version is 13.3.x-13.4.9 and based on if experimental.appDir exists.
- Resolve the styled-jsx/style issue

## Tests

1. TODO


packages/runtime/src/templates/getHandler.ts
```diff
@@ -57,10 +58,13 @@ const makeHandler = ({ conf, app, pageRoot, NextServer, staticManifest = [], mod
...
 
+  const { appDir }: ExperimentalConfigWithLegacy = conf.experimental
   // Next 13.4 conditionally uses different React versions and we need to make sure we use the same one
-  overrideRequireHooks(conf)
+  // With the release of 13.5 experimental.appDir is no longer used.
+  // we will need to check if appDir exists to run requireHooks for older versions
+  if (appDir) overrideRequireHooks(conf.experimental)
   const NetlifyNextServer: NetlifyNextServerType = getNetlifyNextServer(NextServer)
-  applyRequireHooks()
+  if (appDir) applyRequireHooks()
```
MarcL, line 67: Do we want consistent brackets or are we happy with single-line `if` statements?

packages/runtime/src/helpers/edge.ts
```diff
@@ -463,7 +463,7 @@ export const writeEdgeFunctions = async ({
         name: edgeFunctionDefinition.name,
         pattern,
         // cache: "manual" is currently experimental, so we restrict it to sites that use experimental appDir
-        cache: usesAppDir ? 'manual' : undefined,
+        cache: 'manual',
```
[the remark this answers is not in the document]
taty2010: Thanks, I removed it

packages/runtime/src/templates/getHandler.ts
taty2010: Figured since the logic and return were pretty simple having it within a single line wouldn't cause confusion

orinokai: If we do we should probably update eslint, but IMO it's a lesser concern at the moment

orinokai (approved):
this is a chunky fix, nice one @taty2010 - looks good to me

packages/runtime/src/templates/getHandler.ts
MarcL: Yes - agreed.

PR #2341, opened 2023-10-18
fix: copy `vendor` over for local dev

Skn0tt:
Closes https://linear.app/netlify/issue/FRA-53/netlify-dev-fails-to-run-middleware.

In https://github.com/netlify/next-runtime/pull/2302, we started vendoring Deno dependencies, but we forgot about copying them over for local dev. This PR fixes that.

The main change is a one-liner, but there was a problem within the way `esm.sh` bundles these two dependencies, and the way we import them via `../vendor`:

https://github.com/netlify/next-runtime/blob/5d71bfbba214e88cf59ed335a786806c947a2109/packages/runtime/src/templates/edge/next-dev.js#L1-L2

This was fixed by registering the `import_map.json` created by Deno for local dev.  There was a bug in how that works though, that was only fixed in a [recent release](https://github.com/netlify/cli/releases/tag/v16.8.0) of the CLI. This means customers will need to update to that to get local dev working properly.

I've tested this locally with `demos/middleware`, and it's working well!

### DIFF
--- a/packages/runtime/patch-import-map.mjs
+++ b/packages/runtime/patch-import-map.mjs
@@ -1,0 +1,13 @@
+import { readFile, writeFile } from "fs/promises"
+
+const importMapPath = new URL("src/templates/vendor/import_map.json", import.meta.url)
+const importMap = JSON.parse(await readFile(importMapPath, "utf8"))
+
+const newMap = {
+  scopes: {
+    "../": importMap.imports,
+    ...importMap.scopes
+  }
+}
+
+await writeFile(importMapPath, JSON.stringify(newMap, null, 2))
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -55,7 +55,7 @@
     "publish:test": "cd .. && npm ci && npm test",
     "clean": "rimraf lib dist-types",
     "build": "run-s build:*",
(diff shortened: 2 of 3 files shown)


Skn0tt:
One weird thing is that when I go to `http://localhost:8888/shows/100` in the `demos/middleware` site, Next shows me this:

Screenshot 2023-10-18 at 18 00 16

I can't imagine how this would relate to my changes, so i'll assume this also exists on `main`. @orinokai is this a known thing?

packages/runtime/patch-import-map.mjs
```diff
@@ -0,0 +1,13 @@
+import { readFile, writeFile } from "fs/promises"
+
+const importMapPath = new URL("src/templates/vendor/import_map.json", import.meta.url)
+const importMap = JSON.parse(await readFile(importMapPath, "utf8"))
+
+const newMap = {
+  scopes: {
+    "../": importMap.imports,
```
Skn0tt, line 8: We're moving the `imports` field generated by Deno into the `../` scope, so it only applies to our Next code and not to other, Netlify-managed code like the bootstrap layer.

Skn0tt:
Once we have the import map in place, we should be able to simplify the `vendor.ts` file a bit, and can make this change in a lot of our files:

```diff
- ../vendor/esm.sh/v91/next@12.2.5/deno/dist/server/web/spec-extension/request.js
+ https://esm.sh/v91/next@12.2.5/deno/dist/server/web/spec-extension/request.js
```

That'll be a small QoL improvement, since it'll make it easier to update versions of the imported files. Gonna do this in a follow-up!

PR #2393, opened 2023-12-08
fix: check type before using string-only method

pieh:
## Description

Our override of `fs/promises`->`readFile` assumes that first argument is a `string`. `readFile` does accept multiple types - including `URL` that is used by Node.js itself (https://github.com/nodejs/node/blob/8cdb7cae98468efee222727ad9a4096e415f3ef6/lib/internal/modules/esm/load.js#L44-L46) when loading ESM module.

This adds typeguard to prevent trying to use `.startsWith` if argument is not a string. As we only care here about specific context of `readFile` usage - this continues to use a string, so we don't have to try to handle all the possible types that `readFile` accept and only need to handle actual types used in that context. If type is not string, we will do passthrough to original `fs/promises.readFile` function

## Tests

1. TODO

https://answers.netlify.com/t/all-nextjs-builds-returning-500-error-with-typeerror-file-startswith-is-not-a-function/107758/10


demos/default/.nvmrc
```diff
@@ -0,0 +1 @@
+20
```
pieh, line 1: can remove this if anyone feel strongly about not bumping node version for default demo - was part if initial exploration and was just left in here

MarcL (dismissed):
Looks good. I assume we're ok to bump the demos to Node 20.

MarcL (approved):

PR #2396, opened 2023-12-14
feat: add waf bypass header to CDN download function

orinokai:
## Description

Pass through the WAF Bypass Token header from the initial request to the fetch call that downloads files from the CDN when not present in the lambda (e.g. static files such as 500.html files).

## Tests

There is currently no easy way to add a test with the current scaffolding. Given the current priorities, I'd be happy to push this through with manual testing for now.

Main PR to implement the WAF bypass: https://github.com/netlify/pod-adn/issues/82
Prior art: https://github.com/netlify/netlify-ipx/blob/main/src/index.ts#L106-L111

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,6 +47,7 @@
         "eslint-plugin-promise": "^6.0.0",
         "eslint-plugin-unicorn": "^43.0.2",
         "execa": "^5.1.1",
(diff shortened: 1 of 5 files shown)


serhalp:
Superseded by v5.

PR #2421, opened 2024-02-08
feat: support Netlify Image CDN

ascorbic:
## Tests

1. TODO


packages/runtime/package.json
```diff
@@ -15,7 +15,6 @@
     "@netlify/blobs": "^2.2.0",
     "@netlify/esbuild": "0.14.39",
     "@netlify/functions": "^1.6.0",
-    "@netlify/ipx": "^1.4.6",
```
ascorbic, line 18: bye bye

MarcL: 🙌🏻  Hurrah! 🎉

serhalp:
Superseded by v5.

ascorbic:
@serhalp This was specifically to enable it for the legacy adapter so we can deprecate netlify/ipx for old versions too

ascorbic:
It was. This was blocked because of the issue with image CDN not working with private sources, but that's no longer the case. I think it would be worth doing this backport if possible.

PR #2480, opened 2024-06-14
fix: update cache handler to accomodate changes in next@canary

pieh:
## Description

There were quite a bit of changes in `next@canary` that we currently don't handle:

1. They separated cache entries for App router pages and Pages router pages and added new cache kind: `APP_PAGE` - https://github.com/vercel/next.js/pull/65988

This one means we need to update both our CacheHandler to support that kind (should be done) and our build time handling of prerendered content (not fully done yet - missing part is correct condition to use current or new one cache kind - right now I'm relying on our test env var to at least test that blob shape is correct)

2. They moved away from mutating `PrerenderManifest` when doing staleness calculations. Instead they now use `SharedRevalidateTimings` for that ( https://github.com/vercel/next.js/pull/64370 ) and made `PrerenderManifest` frozen  ( https://github.com/vercel/next.js/pull/64313 ) which means we can't mutate it and so https://github.com/netlify/next-runtime/commit/c91e2575dd5b5faee6d9cb67f92def5171462015 is causing actually more problems there (I'm also trying to add some graceful error handling with addition of try/catches so in case the hacks that we have stop working it's not completely borked)

## Tests

1. TODO


package.json
```diff
@@ -79,7 +79,7 @@
     "memfs": "^4.9.2",
     "mock-require": "^3.0.3",
     "msw": "^2.0.7",
-    "next": "^14.0.4",
+    "next": "^15.0.0-canary.28",
```
pieh, line 82: I did this to get updates Cache related types from `next`, otherwise all the `APP_PAGE` related additions were causing a lot of TS issues.

I'm not sure if doing this is the best or wether I should maybe copy over relevant types until those become available in latest `next`?

src/run/config.ts
```diff
@@ -25,6 +25,8 @@ export const setRunConfig = (config: NextConfigComplete) => {
   // set the path to the cache handler
   config.experimental = {
     ...config.experimental,
+    // @ts-expect-error incrementalCacheHandlerPath was removed from config type
+    // but we still need to set it for older Next.js versions
     incrementalCacheHandlerPath: cacheHandler,
```
pieh, line 30: one of type changes in next canary/rc :(

src/build/content/prerendered.ts
```diff
@@ -103,6 +120,18 @@ export const copyPrerenderedContent = async (ctx: PluginContext): Promise<void>
...
       const limitConcurrentPrerenderContentHandling = pLimit(10)
 
+      // https://github.com/vercel/next.js/pull/65988 introduced Cache kind specific to pages in App Router (`APP_PAGE`).
+      // Before this change there was common kind for both Pages router and App router pages
+      // so we check Next.js version to decide how to generate cache values for App Router pages.
+      // Note: at time of writing this code, released 15@rc uses old kind for App Router pages, while 15.0.0@canary.13 and newer canaries use new kind.
+      // Looking at 15@rc release branch it was merging `canary` branch in, so the version constraint assumes that future 15@rc (and 15@latest) versions
+      // will use new kind for App Router pages.
+      const shouldUseAppPageKind = ctx.nextVersion
+        ? satisfies(ctx.nextVersion, '>=15.0.0-canary.13 <15.0.0-d || >15.0.0-rc.0', {
```
pieh, line 130: This is a bit tricky as it tries to predict what future rc/stable releases of Next would likely contain and this is my best guess in hope that version released with this code would be usable with future next versions.

The alternative I was wondering about is doing a test by checking some internal modules, but the Next.js change only added some helper exports ( i.e. https://github.com/vercel/next.js/pull/65988/files#diff-17519b0578e6b6493027a0faa49f2aaab1a70d81216339a93a8d87c68a861f0d ) and only clear `APP_PAGE` things were types which I can't check in runtime :/ But also this test would have to maintained into the future, so this overall is not that great of an idea (IMO)

orinokai: I agree with you about this being difficult to maintain. I did a bit of spelunking and found this util that we could potentially use to ascertain whether `APP_PAGE` is supported https://github.com/vercel/next.js/blob/658037358cf4c9813648e3a9b1c70d7e414c87ff/packages/next/src/server/response-cache/utils.ts#L5

pieh: How would we use that to test it? This helper was added long ago so it couldn't be just testing if there is export like that?

orinokai: We could pass a mock `ResponseCacheEntry` object to it with `value.kind = 'APP_PAGE'` and `value.rscData` set. If `APP_PAGE` is supported it will return an `IncrementalCacheItem` with `value.rscData` still set (canary behaviour) and if not it will return a `ResponseCacheEntry` without `value.rscData`.

pieh: Ok, that makes sense, but do we want to introduce test like (relying on specific behavior of Next.js internal helper) that that we will need to possibly maintain (including finding alternatives if it ever changes)? It feels like more work over-time than possibly adjusting version range above if expectation of the change shipping with `-rc.1` will be incorrect?

orinokai: Maybe a better alternative would be to abstract out the version management, so we can just call something like `hasAppPage()` in the cache handler and the abstraction takes care of the version detection and mapping to version numbers. Then we just accept that we'll need to monitor versions and make changes, but at least it would just be in one place? This could be a follow up PR, so I'm approving this one.

orinokai (approved):
I think this is ready to go, albeit with a new comment below about the version handling.

src/build/content/prerendered.ts
pieh: `CacheHandler` doesn't need to have such check. The `PAGE` `kind` is the same so all the handling for that remain and `APP_PAGE` `kind` has separate handling. The only place that needs conditional logic like that would be when we prepare cache entries from prerendered content when deciding which `kind` the content should be and where to put rsc data in the cache entry + wether it should be plain text or encoded content

PR #2490, opened 2024-06-17
fix: track revalidate / cdn purge to ensure it finishes execution and is not suspended mid-execution

pieh:
## Description

With the changes some time ago to streaming functions to not await untracked background work we might have not handled all cases of background work we should be tracking. We do track background regeneration (~SWR), but on-demand revalidation code paths were not tracked possibly resulting in flaky on-demand revalidate behaviors

## Tests

Not clear how to exactly test this without adding some TEST specific timeous to runtime, if anyone has ideas on what kind of tests could assert that we indeed are awaiting/tracking relevant on-demand revalidation work, happy to add or modify tests.

We do have tests for on-demand revalidation which are a bit flaky. Maybe this will un-flake them (or at least decrease flakiness)?

- https://linear.app/netlify/issue/FRA-566/nextjs-routes-revalidating-when-not-required#comment-1bf42469 (+ my next comment)
   // notes:
   my hunch in here it looks like Next cache is being correctly invalidated, but then CDN purge is not going through resulting in CDN nodes still serving whatever it had cached prior to purge - optionally CDN nodes that didn't have response cached would serve fresh leading to undeterministic response (depending on CDN node that handled request)
   
 - https://linear.app/netlify/issue/FRA-568/help-understanding-cache-tag-usage#comment-169cc400
   // notes:
   App Router `revalidateTag` (or `revalidatePath` for that matter) don't return promises that user can await on, so execution can be suspended without going through <- I'm not exactly the most confident that changes here will fully resolve what this comment talks about, but from my own testing those changes seem required

### DIFF
--- a/src/run/handlers/cache.cts
+++ b/src/run/handlers/cache.cts
@@ -244,12 +244,14 @@ export class NetlifyCacheHandler implements CacheHandler {
         if (requestContext?.didPagesRouterOnDemandRevalidate) {
           const tag = `_N_T_${key === '/index' ? '/' : key}`
           getLogger().debug(`Purging CDN cache for: [${tag}]`)
-          purgeCache({ tags: [tag] }).catch((error) => {
-            // TODO: add reporting here
-            getLogger()
-              .withError(error)
-              .error(`[NetlifyCacheHandler]: Purging the cache for tag ${tag} failed`)
-          })
+          requestContext.trackBackgroundWork(
+            purgeCache({ tags: [tag] }).catch((error) => {
+              // TODO: add reporting here
+              getLogger()
+                .withError(error)
+                .error(`[NetlifyCacheHandler]: Purging the cache for tag ${tag} failed`)
+            }),
+          )
         }
       }
     })
@@ -257,6 +259,18 @@ export class NetlifyCacheHandler implements CacheHandler {
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   async revalidateTag(tagOrTags: string | string[], ...args: any) {
+    const revalidateTagPromise = this.doRevalidateTag(tagOrTags, ...args)
+
+    const requestContext = getRequestContext()
+    if (requestContext) {
+      requestContext.trackBackgroundWork(revalidateTagPromise)
+    }
+
+    return revalidateTagPromise
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private async doRevalidateTag(tagOrTags: string | string[], ...args: any) {
     getLogger().withFields({ tagOrTags, args }).debug('NetlifyCacheHandler.revalidateTag')
 
     const tags = Array.isArray(tagOrTags) ? tagOrTags : [tagOrTags]
@@ -275,7 +289,7 @@ export class NetlifyCacheHandler implements CacheHandler {
       }),
     )
 
-    purgeCache({ tags }).catch((error) => {
+    await purgeCache({ tags }).catch((error) => {
       // TODO: add reporting here
       getLogger()
         .withError(error)
--- a/src/run/revalidate.ts
+++ b/src/run/revalidate.ts
@@ -1,4 +1,5 @@
 import type { ServerResponse } from 'node:http'
+import { isPromise } from 'node:util/types'
 
 import type { RequestContext } from './handlers/request-context.cjs'
 
@@ -12,7 +13,13 @@ export const nextResponseProxy = (res: ServerResponse, requestContext: RequestCo
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         return async function newRevalidate(...args: any[]) {
           requestContext.didPagesRouterOnDemandRevalidate = true
-          return originalValue?.apply(target, args)
+
+          const result = originalValue?.apply(target, args)
+          if (result && isPromise(result)) {
+            requestContext.trackBackgroundWork(result)
+          }
+
+          return result
         }
       }
       return originalValue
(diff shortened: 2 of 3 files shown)


src/run/handlers/cache.cts
```diff
@@ -244,19 +244,33 @@ export class NetlifyCacheHandler implements CacheHandler {
...
-              .error(`[NetlifyCacheHandler]: Purging the cache for tag ${tag} failed`)
-          })
+          requestContext.trackBackgroundWork(
+            purgeCache({ tags: [tag] }).catch((error) => {
+              // TODO: add reporting here
+              getLogger()
+                .withError(error)
+                .error(`[NetlifyCacheHandler]: Purging the cache for tag ${tag} failed`)
+            }),
+          )
```
pieh, line 254: making sure we track `purgeCache` part of the pages router on-demand revalidation so it completes before function execution gets suspended

src/run/handlers/cache.cts
```diff
@@ -244,19 +244,33 @@ export class NetlifyCacheHandler implements CacheHandler {
...
+    const requestContext = getRequestContext()
+    if (requestContext) {
+      requestContext.trackBackgroundWork(revalidateTagPromise)
+    }
+
+    return revalidateTagPromise
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private async doRevalidateTag(tagOrTags: string | string[], ...args: any) {
```
pieh, line 273: App Router `revalidateTag` and `revalidatePath` don't return a promise so user can't even await on the revalidation work in their code. Instead we will track it internally (should not affect timing of the response, just that we should keep execution running and not suspend before this is completed)

src/run/handlers/cache.cts
```diff
@@ -275,7 +289,7 @@ export class NetlifyCacheHandler implements CacheHandler {
       }),
     )
 
-    purgeCache({ tags }).catch((error) => {
+    await purgeCache({ tags }).catch((error) => {
```
pieh, line 292: this doesn't change much, because `.revalidateTag` overall was not awaited on and all of this is ~background work that won't block response

src/run/revalidate.ts
```diff
@@ -12,7 +13,13 @@ export const nextResponseProxy = (res: ServerResponse, requestContext: RequestCo
...
         return async function newRevalidate(...args: any[]) {
           requestContext.didPagesRouterOnDemandRevalidate = true
-          return originalValue?.apply(target, args)
+
+          const result = originalValue?.apply(target, args)
+          if (result && isPromise(result)) {
+            requestContext.trackBackgroundWork(result)
+          }
+
+          return result
```
pieh, line 22: User's CAN await pages router `res.revalidate()`, so if user is doing that, this is not changing anything, but it's quite easy to miss `await` in user code (especially that the promise would resolve to `void` anyway, so user would not be doing anything with `res.revalidate()` return value (other than hopefully `await` it).

This just make sure that if user doesn't `await` it, we will (not blocking a response, just tracking this as background work)

src/run/revalidate.ts
```diff
@@ -1,4 +1,5 @@
 import type { ServerResponse } from 'node:http'
+import { isPromise } from 'node:util/types'
```
serhalp, line 2: TIL

serhalp (approved):
LGTM, good catch

src/run/revalidate.ts
```diff
@@ -12,7 +13,13 @@ export const nextResponseProxy = (res: ServerResponse, requestContext: RequestCo
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         return async function newRevalidate(...args: any[]) {
```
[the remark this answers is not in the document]
pieh: Not sure why, but at least with code as is we ourselves don't use `args` directly and only forward them to actual `res.revalidate` method so in this case (at least this specific line, as there are some other `any` in this module) we don't necessarily care about types, but there is no reason not to add proper types here I think. I'll take a look and see if I can get rid of all those `any` in this module.

tests/fixtures/page-router/pages/api/revalidate.js
```diff
@@ -1,6 +1,9 @@
 export default async function handler(req, res) {
   try {
-    await res.revalidate('/static/revalidate-manual')
+    // res.revalidate returns a promise that can be awaited to wait for the revalidation to complete
+    // if user doesn't await it, we still want to ensure the revalidation is completed, so we internally track
+    // this as "background work" to ensure it completes before function suspends execution
+    res.revalidate('/static/revalidate-manual')
```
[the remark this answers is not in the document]
pieh: I think it makes sense to have 2 cases here so in case those e2e tests fail it's easier to tell wether there is overall problem with revalidation or specifically with not awaited revalidation. Adding variants here would be much smoother after https://github.com/netlify/next-runtime/pull/2495 is in as it does add ~`test.each` kind of setup to on-demand revalidation tests where it's super easy to add another case, so I'll get this one in first and then revert this change and instead add another API endpoint that doesn't await

src/run/revalidate.ts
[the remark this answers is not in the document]
pieh: https://github.com/netlify/next-runtime/pull/2490/commits/b475810e3d8498ef4fdbcf5fabf48250c78aabd2 got rid of `any` usage. Added type guard there is pretty rudimentary, but I didn't really see a way to make it better/narrower (it only check `key` - which we already were doing + started checking if value is at least a function)

serhalp (dismissed):

tests/fixtures/page-router/pages/api/revalidate.js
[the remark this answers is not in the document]
pieh: https://github.com/netlify/next-runtime/pull/2490/commits/dd43ce8491f6642cdbb09d4a0d4bccaf29235ef5 added separate case for not awaiting `res.revalidate` on top of existing ones

orinokai (approved):
Well caught! Certainly looks like this will make on-demand revalidation more robust.

PR #2495, opened 2024-06-18
fix: set netlify-cache-tag for not prerendered content

pieh:
## Description

We are currently relying on `tags-manifest.json` to set `netlify-cache-tag` response header. Tags Manifest is produced at build time from prerendered content so it won't contain any not prerendered pages which will result in this header missing for not prerendered pages and therefore make it impossible to purge CDN cache (it's still possible to invalidate Next.js cache, but if there were CDN nodes that already cached content - we aren't able to purge that right now). This might result in "random response" being served to user depending on which CDN handled the request (it might be fresh response or it might be old response that was cached)

## Tests

Added failing E2E test cases in first 2 commits for both App and Pages Router

https://linear.app/netlify/issue/FRP-566/nextjs-routes-revalidating-when-not-required#comment-1bf42469


src/build/content/server.ts
```diff
@@ -311,47 +311,6 @@ export const copyNextDependencies = async (ctx: PluginContext): Promise<void> =>
   })
 }
 
-export const writeTagsManifest = async (ctx: PluginContext): Promise<void> => {
-  const manifest = await ctx.getPrerenderManifest()
```
serhalp, line 315: Looks like we can [delete this](https://github.com/netlify/next-runtime/blob/4b6de32222c22c1d5dda9809d5842e54a45a37f8/tests/utils/helpers.ts#L15-L27) too

serhalp: ... Oh, I don't think this is related to your changes, but still dead code

src/run/handlers/cache.cts
```diff
@@ -273,6 +312,10 @@ export class NetlifyCacheHandler implements CacheHandler {
 
       const value = this.transformToStorableObject(data, context)
 
+      // if previous CacheHandler.get call returned null (page was either never rendered on was on-demand revalidated)
```
serhalp, line 315:
```suggestion
      // if previous CacheHandler.get call returned null (page was either never rendered or was on-demand revalidated)
```
nit

src/run/handlers/cache.cts
```diff
@@ -109,6 +109,44 @@ export class NetlifyCacheHandler implements CacheHandler {
...
+
+    // Bail if we already have cache tags - `captureCacheTags()` is called on both `CacheHandler.get` and `CacheHandler.set`
+    // that's because `CacheHandler.get` might not have a cache value (cache miss or on-demand revalidation) in which case
+    // response is generated in blocking way and we need to capture cache tags from the cache value we are setting.
+    // If both `CacheHandler.get` and `CacheHandler.set` are called in the same request, we want to use cache tags from
+    // first `CacheHandler.get` and not from following `CacheHandler.set` as this is pattern for Stale-while-revalidate behavior
+    // and stale response is served while new one is generated.
+    if (requestContext.responseCacheTags) {
+      return
+    }
```
serhalp, line 131: I'm trying to think of something less leaky but I can't 😞

serhalp (dismissed):
LGTM, nice work! Thanks for the explanations and comments.

src/build/content/server.ts
pieh: Yeah, not really related to this change, but I did remove it now

PR #2498, opened 2024-06-18
chore: output playwright report in readable form in github actions output

pieh:
## Description

This is optimistic addition so we can have easy way to lookup complete report with status of tests on github action run page without having to download html or json artifacts or going through each shard results

## Tests

1. TODO


pieh:
This kind of worked, but it also capture all the other spam, so I'll close this as printed output is not very readable still

PR #2502, opened 2024-06-20
feat: add central cache directive

eduardoboucas:
## Description

Adds the central cache directive to cache control headers.

### DIFF
--- a/src/run/handlers/server.ts
+++ b/src/run/handlers/server.ts
@@ -48,6 +48,9 @@ const disableFaultyTransferEncodingHandling = (res: ComputeJsOutgoingMessage) =>
 // TODO: remove once https://github.com/netlify/serverless-functions-api/pull/219
 // is released and public types are updated
 interface FutureContext extends Context {
+  flags: {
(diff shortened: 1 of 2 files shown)


src/run/handlers/server.ts
```diff
@@ -48,6 +48,9 @@ const disableFaultyTransferEncodingHandling = (res: ComputeJsOutgoingMessage) =>
 // TODO: remove once https://github.com/netlify/serverless-functions-api/pull/219
 // is released and public types are updated
 interface FutureContext extends Context {
+  flags: {
```
eduardoboucas, line 51: This is not meant to be a customer-facing property, so we never added it to the `Context` types. I've added it here for now, but we should consider having a `InternalContext` or similar that extends the main context with any internal properties.

orinokai: Makes sense. Might be worth adding the reason for putting this here to the comment above, until `InternalContext` is implemented

eladroz:
@eduardoboucas I guess this can be closed now that #2510 is merged?

PR #2510, opened 2024-06-25
feat: use Netlify Durable Cache

serhalp:
## Description

👁️‍🗨️  **This will be easier to review commit by commit:**

- 287177af194b16003af43df9b309c19d0ea13d76 test: add missing unit test coverage for route handler headers
- a8d8fcaf4bae0b896b0f4e595e95e17b8e897e4c feat: specify `durable` cache-control directive if feature flag is enabled for request
- 4bd24ffe7e2152e29ae7887e21e9385e39ac6c69 feat: specify durable cache-control directive

This is gated behind a feature flag for now.

I can't link to any public docs yet, but by the time you're reading this you should be able to find a section on "Durable caching" at https://docs.netlify.com.

### Documentation

This functionality is now documented at https://docs.netlify.com/platform/caching/#durable-directive.

## Tests

This PR adds exhaustive unit tests and an e2e integration test deployed to a site with the feature flags enabled. I opted not to test through to the platform behavior here since there's no real way to guarantee you're actually hitting separate CDN nodes on subsequent requests and I'd rather not write tests that don't provide confidence in the behavior under test.

You can also manually deploy to that site and manually verify the expected behavior, which I've done.

You can also use this demo site I created: https://github.com/serhalp/next.js-netlify-durable-cache-demo.

FRP-749

Supersedes https://github.com/netlify/next-runtime/pull/2502.

### DIFF
--- a/src/run/handlers/server.ts
+++ b/src/run/handlers/server.ts
@@ -112,7 +112,9 @@ export default async (request: Request, context: FutureContext) => {
 
     await adjustDateHeader({ headers: response.headers, request, span, tracer, requestContext })
 
-    setCacheControlHeaders(response.headers, request, requestContext)
+    const useDurableCache =
+      context.flags.get('serverless_functions_nextjs_durable_cache_disable') !== true
+    setCacheControlHeaders(response.headers, request, requestContext, useDurableCache)
     setCacheTagsHeaders(response.headers, requestContext)
     setVaryHeaders(response.headers, request, nextConfig)
     setCacheStatusHeader(response.headers)
--- a/tests/utils/fixture.ts
+++ b/tests/utils/fixture.ts
@@ -322,6 +322,9 @@ export async function uploadBlobs(ctx: FixtureTestContext, blobsDir: string) {
   )
 }
 
+const DEFAULT_FLAGS = {
+  serverless_functions_nextjs_durable_cache_disable: true,
+}
 /**
  * Execute the function with the provided parameters
  * @param ctx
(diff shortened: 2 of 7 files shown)


src/run/handlers/server.ts
```diff
@@ -112,7 +112,10 @@ export default async (request: Request, context: FutureContext) => {
 
     await adjustDateHeader({ headers: response.headers, request, span, tracer, requestContext })
 
-    setCacheControlHeaders(response.headers, request, requestContext)
+    const useDurableCache = context.flags.get('serverless_functions_nextjs_durable_cache') as
+      | boolean
+      | undefined
```
serhalp, line 117: The return value here is typed as `unknown` because flags (a bit of a misnomer; they're "variables") can be booleans, strings, and numbers.

It seems a bit extreme to `assert` the type here, which would throw at runtime, so I just went with a type assertion.

tests/e2e/simple-app.test.ts
```diff
@@ -12,7 +12,9 @@ test('Renders the Home page correctly', async ({ page, simple }) => {
 
   await expect(page).toHaveTitle('Simple Next App')
 
-  expect(headers['cache-status']).toBe('"Next.js"; hit\n"Netlify Edge"; fwd=miss')
+  expect(headers['cache-status']).toBe(
+    '"Next.js"; hit\n"Netlify Durable"; fwd=miss\n"Netlify Edge"; fwd=miss',
+  )
```
serhalp, line 17: duplicated from https://github.com/netlify/next-runtime/pull/2527 to unblock myself

tests/utils/fixture.ts
```diff
@@ -322,6 +322,9 @@ export async function uploadBlobs(ctx: FixtureTestContext, blobsDir: string) {
   )
 }
 
+const DEFAULT_FLAGS = {
+  serverless_functions_nextjs_durable_cache_disable: true,
```
serhalp, line 326: Since I implemented the flag logic to treat a nullish flag as `false` (i.e. don't disable Durable Caching, so that we can actually archive the feature flag someday), dozens of integration tests (which were implicitly mocking runtime `context.flags` to be empty) started failing. This change fixes that, by opting those out of the flag for now.

pieh (approved):

PR #2512, opened 2024-06-26
feat: fail build when netlify form detected 

orinokai:
## Description

Netlify Forms are not supported in v5 and we have currently excluded affected customers from the rollout. This PR will fail the build when a Netlify Form is detected, meaning we can start to migrate customers away from Forms.

The PR simply uses a regex to check for the existence of `data-netlify` within a `<form>` tag in the prerendered content.

UPDATE: this verification is now passive and only logs a warning so that we can monitor and reach out to customers before beginning the migration

## Tests

Integration tests added

Closes FRP-696


src/build/content/prerendered.ts
```diff
@@ -169,6 +170,11 @@ export const copyPrerenderedContent = async (ctx: PluginContext): Promise<void>
                   throw new Error(`Unrecognized content: ${route}`)
               }
 
+              // Netlify Forms are not support and require a workaround
+              if (value.kind === 'PAGE') {
```
pieh, line 174: does it also need to check `APP_PAGE`? (next@canary use that kind for app router page cache)

orinokai: Ah, good catch 🙏 fixed in https://github.com/netlify/next-runtime/pull/2512/commits/faed5fafe6b564fdcf97300e97e971b4d1a59bad

tests/integration/netlify-forms.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  vi.stubEnv('SITE_ID', ctx.siteID)
+  vi.stubEnv('DEPLOY_ID', ctx.deployID)
+  vi.stubEnv('NETLIFY_PURGE_API_TOKEN', 'fake-token')
+  // hide debug logs in tests
+  // vi.spyOn(console, 'debug').mockImplementation(() => {})
+
+  await startMockBlobStore(ctx)
+})
+
+it.skip<FixtureTestContext>('test', async (ctx) => {
```
pieh, line 23: is it worth checking if `console.warn` was called? this test (and fixture) right now doesn't do anything (I guess until warning is swapped to build failing)?

orinokai: i wondered this, but since the dry run is a temporary solution i didn't think it was worth it

pieh: Also fine with pushing this in - but maybe we could name the test already?
```suggestion
// test skipped until we actually start failing builds - right now we are just showing a warning
it.skip<FixtureTestContext>('should fail build when netlify forms are used', async (ctx) => {
```

pieh (approved):

PR #2523, opened 2024-07-02
feat: check for netlify forms workaround

orinokai:
## Description

A [previous PR](https://github.com/netlify/next-runtime/pull/2512) introduced a warning for customers using Netlify Forms. This PR adds a check to skip the warning if the [forms workaround](https://docs.netlify.com/frameworks/next-js/overview/#limitations) is detected in use.

## Tests

Integration tests added

Closes FRP-696


src/build/content/prerendered.ts
```diff
@@ -172,7 +172,7 @@ export const copyPrerenderedContent = async (ctx: PluginContext): Promise<void>
 
               // Netlify Forms are not support and require a workaround
               if (value.kind === 'PAGE' || value.kind === 'APP_PAGE') {
-                verifyNoNetlifyForms(ctx, value.html)
+                verifyNetlifyForms(ctx, value.html)
```
orinokai, line 175: I made these verifications all positive because it was getting confusing knowing when to warn

src/build/verification.ts
```diff
@@ -88,11 +90,33 @@ export async function verifyNoAdvancedAPIRoutes(ctx: PluginContext) {
...
 
-export function verifyNoNetlifyForms(ctx: PluginContext, html: string) {
-  if (!warnings.has('netlifyForms') && /<form[^>]*?\s(netlify|data-netlify)[=>\s]/.test(html)) {
+const formDetectionRegex = /<form[^>]*?\s(netlify|data-netlify)[=>\s]/
+
+export async function verifyNetlifyFormsWorkaround(ctx: PluginContext) {
+  const srcDir = ctx.resolveFromSiteDir('public')
+  const paths = await glob(join(srcDir, '**/*.html'))
+  try {
+    const hasWorkaround = await paths.some(async (path): Promise<boolean> => {
```
pieh, line 99: I'm pretty sure this doesn't work as expected - `some` predicate will always be truthy as it always return a promise so `hasWorkaround` would always be `true`?

pieh: Can you add html file with this snippet (so missing `netlify` attribute) to previous fixture (i.e. `tests/fixtures/netlify-forms/public/not-netlify-form.html`)?:

```html
<form name="contact"></form>
```

in my local testing adding that file make the test we have for actually showing warning no longer do that, because `hasWorkaround` will be truthy as long there are any `.html` files in `public` dir

src/build/verification.ts
```diff
@@ -1,14 +1,16 @@
...
+import { glob } from 'fast-glob'
 import { satisfies } from 'semver'
 
 import { ApiRouteType, getAPIRoutesConfigs } from './advanced-api-routes.js'
 import type { PluginContext } from './plugin-context.js'
 
 const SUPPORTED_NEXT_VERSIONS = '>=13.5.0'
 
-const warnings = new Set<string>()
+const verifications = new Set<string>()
```
pieh, line 13: NIT: should we put this on `PluginContext` object instead of module scope maybe? this forces the use of `resetModules` which does work, but feels wrong to use for code we own if we could avoid that?

src/build/verification.ts
orinokai: well caught! fixed in [62faedc](https://github.com/netlify/next-runtime/pull/2523/commits/62faedc8ae7a6eae2542b859f175585830f264b2) and test modifications made as suggested

src/build/verification.ts
orinokai: i attempted to do this in various ways but ended up down a rabbit hole, because it didn't seem right to store the verifications state in the plugin context (and mutate it directly from the verifications file) as it doesn't have anything to do with the rest of the stuff in the plugins context. I'd like to handle verifications differently (maybe creating a class that can store its own state?) but that's for a follow up

pieh: That's ok, this is not a blocker to me

pieh (approved):

PR #2536, opened 2024-07-08
fix: only clear stale functions with build through CLI

pieh:
## Description

Alternative to https://github.com/netlify/next-runtime/pull/2511

This doesn't fully solve the problem as it can still happen with CLI (it does reduce reach of the problem), but feels much safer than previous PR. This relies on fact that builds on platform don't restore functions generated by previous builds therefore there is no need to clear those dirs. It also moves dir clearing to earlier stage so it's less likely that other integrations produced any functions yet (in case of using CLI)

## Tests

1. TODO

### DIFF
--- a/src/build/functions/edge.ts
+++ b/src/build/functions/edge.ts
@@ -162,9 +162,11 @@ const buildHandlerDefinition = (
   }))
 }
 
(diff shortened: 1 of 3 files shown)


serhalp (approved):
LGTM, seems ok for now

PR #2540, opened 2024-07-09
ci(e2e-report): support generating report site for version other than latest

serhalp:
## Description

This replaces the hardcoded Next.js "latest" version with an optional selector input.

While I was at it, I clarified some confusing aspects of this setup. It's tripped up multiple people already.

### Documentation

N/A

## Tests

When you make changes to a manual workflow setup, you can't see it reflected in the UI until you merge. However, I was able to use the GitHub CLI to manually trigger workflows against this branch's version of the workflow and it worked:

```sh
# run against default (latest)
gh workflow run -r ci/read-correct-e2e-results-file e2e-report.yml
# run against specific version (canary)
gh workflow run -r ci/read-correct-e2e-results-file e2e-report.yml -f version=canary
```

This resulted in these runs, respectively:
- https://github.com/netlify/next-runtime/actions/runs/9895424607
- https://github.com/netlify/next-runtime/actions/runs/9895445211

N/A

### DIFF
--- a/.github/workflows/e2e-report.yml
+++ b/.github/workflows/e2e-report.yml
@@ -7,6 +7,14 @@ on:
       use-branch:
         description: 'Enable if you want to test data from your selected branch instead of the scheduled test runs from Main'
         type: boolean
+      version:
+        description: 'Version of Next.js (most recent test run must have included this version)'
+        type: choice
+        options:
+          - 'latest'
+          - 'canary'
+          - '13.5.1'
+        default: 'latest'
 
 env:
   NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-report.yml
```diff
@@ -27,12 +35,15 @@ jobs:
...
           echo "runId=$E2E_RUN_ID" >> $GITHUB_OUTPUT
-      - name: Download latest e2e results
+      - name: Download e2e results
         if: ${{ steps.get-run-id.outputs.runId }}
         run: |
-          echo "Downloading latest test results from run https://github.com/netlify/next-runtime/actions/runs/${{ steps.get-run-id.outputs.runId }}"
-          rm e2e-report/data/test-results.json
-          gh run download ${{ steps.get-run-id.outputs.runId }} -n "latest-test-results.json" -D e2e-report/data/ --repo $GITHUB_REPOSITORY
+          version="${{ inputs.version }}"
+          version=${version:-latest}
```
serhalp, line 42: we need a second layer of defaulting here because the scheduled run won't have the workflow_dispatch default populated

eladroz: True, but we can have it manual only just so we can do first release, and then fix this.
I don't think we're ready to automatically deploy on a schedule yet, in terms of test stability. Would be great to have a deploy preview automatically though (so then it's much easier to evaluate before manual release)

serhalp: To clarify for anyone reading this: this job already runs on a schedule and only ever creates deploy previews. We publish deploys to production manually in the Netlify UI.

pieh (approved):

PR #2547, opened 2024-07-16
Michalpiechowiak/frp 765 migrate next runtime to use frameworks api

pieh:
## Description

work in progress on migrating next runtime to use frameworks API. It still will support older method to support cases of CLI usage (which user might not updated)

## Tests

1. TODO

### DIFF
--- a/src/build/functions/edge.ts
+++ b/src/build/functions/edge.ts
@@ -1,17 +1,17 @@
 import { cp, mkdir, readFile, rm, writeFile } from 'node:fs/promises'
 import { dirname, join } from 'node:path'
 
-import type { Manifest, ManifestFunction } from '@netlify/edge-functions'
+import type { IntegrationsConfig } from '@netlify/edge-functions'
 import { glob } from 'fast-glob'
 import type { EdgeFunctionDefinition as NextDefinition } from 'next/dist/build/webpack/plugins/middleware-plugin.js'
 import { pathToRegexp } from 'path-to-regexp'
 
 import { EDGE_HANDLER_NAME, PluginContext } from '../plugin-context.js'
 
-const writeEdgeManifest = async (ctx: PluginContext, manifest: Manifest) => {
-  await mkdir(ctx.edgeFunctionsDir, { recursive: true })
-  await writeFile(join(ctx.edgeFunctionsDir, 'manifest.json'), JSON.stringify(manifest, null, 2))
-}
+// const writeEdgeManifest = async (ctx: PluginContext, manifest: Manifest) => {
+//   await mkdir(ctx.edgeFunctionsDir, { recursive: true })
+//   await writeFile(join(ctx.edgeFunctionsDir, 'manifest.json'), JSON.stringify(manifest, null, 2))
+// }
 
 const copyRuntime = async (ctx: PluginContext, handlerDirectory: string): Promise<void> => {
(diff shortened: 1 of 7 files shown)


tests/utils/fixture.ts
```diff
@@ -298,28 +298,54 @@ export async function runPlugin(
...
+        const metadata = await readFile(metaFile, 'utf-8')
+          .then((meta) => JSON.parse(meta))
+          .catch(() => ({}))
+        await ctx.blobStore.set(key, await readFile(join(pluginContext.blobDir, key), 'utf-8'), {
+          metadata,
+        })
+      }),
+    )
+  }
 }
```
pieh, line 349: This entire function doesn't seem like should be implemented here (as in in next-runtime repo) as it does reimplement file-based blobs uploads (something which ideally could be imported from ~netlify/build and just used) - now it "has to" support different implementations (when moving to regional blobs it was only directory change, but frameworks api also has updated directory structure).

This is not the biggest deal because it's quite straight forward - but maybe something to keep in mind in case ~integration kind of tests would be introduced in other repos

src/build/functions/edge.ts
```diff
@@ -174,10 +184,12 @@ export const createEdgeHandlers = async (ctx: PluginContext) => {
...
-  const netlifyManifest: Manifest = {
-    version: 1,
-    functions: netlifyDefinitions,
+  if (!ctx.useFrameworksAPI) {
+    const netlifyDefinitions = nextDefinitions.flatMap((def) => buildHandlerDefinition(ctx, def))
+    const netlifyManifest: Manifest = {
+      version: 1,
+      functions: netlifyDefinitions,
+    }
+    await writeEdgeManifest(ctx, netlifyManifest)
```
pieh, line 193: this doesn't feel great to use conditional check like that in 2 separate places that relate to ~same thing (wether ~config is declared in manifest.json file or inline in function) - it feels brittle and also it feels hard to read

there is some simple abstraction that can be added to maybe increase readability - like introduce `shouldGenerateManifest` and `shouldUseInlineConfig` and following those make it easier to understand what happens when, but for now I wanted to do least code changes possible to maybe see if someone has nice idea how to best deal with those conditional code paths this isn't over the top overengineered but is still readable and easy to follow?

tests/utils/fixture.ts
```diff
@@ -184,7 +184,7 @@ export async function runPluginStep(
       // EDGE_FUNCTIONS_DIST: '.netlify/edge-functions-dist/',
       // CACHE_DIR: '.netlify/cache',
       // IS_LOCAL: true,
-      // NETLIFY_BUILD_VERSION: '29.23.4',
+      NETLIFY_BUILD_VERSION: '29.50.5',
```
pieh, line 187: This makes integration tests use Frameworks API code paths - because integration test setup rely on `PluginContext` already when looking for serverless and edge functions to bundle and invoke them, this almost doesn't need any additional changes other then adjustment to simulating file based blob uploads.

Is there value in trying to run integration tests vs multiple netlify/build version? I don't think it does given that we implement lot of integration/test setup ourselves, but doesn't hurt to ask for other opinions here.

I did add one ~kitchen-sink e2e test to run on ~older CLI implicitly testing blobs, serverless and edge functions as a way to at least have some ~smoke test to have some defense against regressions there in future, but it's pretty minimal

PR #2548, opened 2024-07-16
fix: ensure middleware matchers evaluate default locale

orinokai:
## Description

Middleware matching is performed at the edge and also in the edge function because there are some patterns we are unable to evaluate at the edge.

When i18n is enabled, the regex generated by Next matches paths with or without a locale in the first segment of the URL path. However, it does not match the specific list of locales and simply matches _any string_ in the first segment.

This is a problem because `/test` will also match any page at the next level, e.g. `/products/test` because the matcher assumes `products` is a locale.

This PR splits URL normalizing and localizing into separate functions, for clarity and type-safety (`buildNextRequest` now returns a partial `NextRequest` object), and returns a `localizedURL` parameter to the middleware matcher so that the above case is avoided by always evaluating the match using the prefixed locale or the default, e.g. `/en/products/test`.

## Tests

1. Tests have been updated to catch this case

Fixes FRP-761

### DIFF
--- a/edge-runtime/lib/response.ts
+++ b/edge-runtime/lib/response.ts
@@ -4,7 +4,7 @@ import { HTMLRewriter } from '../vendor/deno.land/x/html_rewriter@v0.1.0-pre.17/
 import { updateModifiedHeaders } from './headers.ts'
 import type { StructuredLogger } from './logging.ts'
 import { addMiddlewareHeaders, isMiddlewareRequest, isMiddlewareResponse } from './middleware.ts'
-import { RequestData } from './next-request.ts'
+import { NetlifyNextContext } from './next-request.ts'
 import {
   addBasePath,
   normalizeDataUrl,
@@ -20,20 +20,18 @@ export interface FetchEventResult {
 
 interface BuildResponseOptions {
   context: Context
-  logger: StructuredLogger
+  nextContext?: NetlifyNextContext
   request: Request
   result: FetchEventResult
-  nextConfig?: RequestData['nextConfig']
-  requestLocale?: string
+  logger: StructuredLogger
 }
 
 export const buildResponse = async ({
   context,
-  logger,
+  nextContext,
   request,
   result,
-  nextConfig,
-  requestLocale,
+  logger,
 }: BuildResponseOptions): Promise<Response | void> => {
   logger
     .withFields({ is_nextresponse_next: result.response.headers.has('x-middleware-next') })
@@ -185,9 +183,9 @@ export const buildResponse = async ({
     }
 
     // respect trailing slash rules to prevent 308s
-    rewriteUrl.pathname = normalizeTrailingSlash(rewriteUrl.pathname, nextConfig?.trailingSlash)
+    rewriteUrl.pathname = normalizeTrailingSlash(rewriteUrl.pathname, nextContext?.trailingSlash)
 
-    const target = normalizeLocalizedTarget({ target: rewriteUrl.toString(), request, nextConfig })
+    const target = normalizeLocalizedTarget({ target: rewriteUrl.toString(), request, nextContext })
     if (target === request.url) {
       logger.withFields({ rewrite_url: rewrite }).debug('Rewrite url is same as original url')
       return
@@ -198,8 +196,8 @@ export const buildResponse = async ({
   }
 
   // If we are redirecting a request that had a locale in the URL, we need to add it back in
-  if (redirect && requestLocale) {
-    redirect = normalizeLocalizedTarget({ target: redirect, request, nextConfig, requestLocale })
+  if (redirect && nextContext?.detectedLocale) {
+    redirect = normalizeLocalizedTarget({ target: redirect, request, nextContext })
     if (redirect === request.url) {
       logger.withFields({ rewrite_url: rewrite }).debug('Rewrite url is same as original url')
       return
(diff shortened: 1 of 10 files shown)


src/build/functions/edge.ts
```diff
@@ -65,10 +65,7 @@ const writeHandlerFile = async (ctx: PluginContext, { matchers, name }: NextDefi
 
   // Writing a file with the matchers that should trigger this function. We'll
   // read this file from the function at runtime.
-  await writeFile(
-    join(handlerRuntimeDirectory, 'matchers.json'),
-    JSON.stringify(augmentMatchers(matchers, ctx)),
-  )
+  await writeFile(join(handlerRuntimeDirectory, 'matchers.json'), JSON.stringify(matchers))
```
orinokai, line 68: We no longer need to augment the middleware matchers that run inside the edge function as we handle this directly now

tests/integration/edge-handler.test.ts
```diff
@@ -261,27 +261,27 @@ describe("aborts middleware execution when the matcher conditions don't match th
...
       const response = await invokeEdgeFunction(ctx, {
         functions: ['___netlify-edge-handler-middleware'],
         origin,
         url: path,
       })
-      expect(
-        response.headers.has('x-hello-from-middleware-res'),
-        `does not match ${path}`,
-      ).toBeFalsy()
+      expect(response.headers.has('x-hello-from-middleware-res'), `does match ${path}`).toBeFalsy()
```
orinokai, line 284: These `expect` messages were the wrong way around, which was complicating the debugging somewhat!

tests/integration/edge-handler.test.ts
```diff
@@ -261,27 +261,27 @@ describe("aborts middleware execution when the matcher conditions don't match th
 
     ctx.cleanup?.push(() => origin.stop())
 
-    for (const path of ['/hello', '/en/hello', '/nl-NL/hello', '/nl-NL/about']) {
+    for (const path of ['/hello', '/en/hello', '/es/hello', '/nl/about']) {
```
orinokai, line 264: `/hello` should be matched with any or no locale, whereas the `/nl/about` matcher was specified with `locale: false` (an undocumented Next.js feature 🙄) so should only be matched with that specific locale

tests/integration/edge-handler.test.ts
```diff
@@ -261,27 +261,27 @@ describe("aborts middleware execution when the matcher conditions don't match th
...
+      expect(
+        response.headers.has('x-hello-from-middleware-res'),
+        `does not match ${path}`,
+      ).toBeTruthy()
       expect(await response.text()).toBe('Hello from origin!')
       expect(response.status).toBe(200)
     }
 
-    for (const path of ['/hello/invalid', '/about', '/en/about']) {
+    for (const path of ['/hello/invalid', '/invalid/hello', '/about', '/en/about', '/es/about']) {
```
orinokai, line 278: `/hello` should not be matched when prefixed with a non-locale (this is the main part of the fix) and `/about` should not be matched when prefixed with anything other than the specific `nl` locale

tests/integration/edge-handler.test.ts
orinokai: I believe that `nl-NL` was previously asserted on here because the [Next.js docs](https://nextjs.org/docs/pages/building-your-application/routing/internationalization#getting-started) are confusing and seem to indicate that `nl-NL` should fallback to `nl`, but that's only for the initial i18n redirect on the index page - the matchers shouldn't do this (or at least they don't when I run Next.js locally). The test appeared to pass, but that was in fact only due to the bug we're trying to fix here.

pieh: But our i18n config - https://github.com/netlify/next-runtime/blob/a7620fdd9009f25fb7820dd43828a30aa22ff434/tests/fixtures/middleware-conditions/next.config.js#L5 doesn't match with Vercel's fixture - https://github.com/vercel/next.js/blob/canary/test/e2e/middleware-custom-matchers-i18n/app/next.config.js#L3 so probably this whole test was ~borked?

tests/integration/edge-handler.test.ts
pieh: would writing message with "should" instead of "does (not)" make things less confusing? so in this case "middleware should not match ${path}" or something like that? then it doesn't matter wether it passes or not - intent is clear on what should happen

edge-runtime/lib/util.ts
```diff
@@ -29,6 +29,19 @@ export const addBasePath = (path: string, basePath?: string) => {
   return path
 }
 
+// add locale prefix if not present, allowing for locale fallbacks
+export const addLocale = (path: string, locale?: string) => {
+  if (
+    locale &&
+    path.toLowerCase() !== `/${locale.toLowerCase()}` &&
+    !path.toLowerCase().startsWith(`/${locale.toLowerCase()}/`) &&
+    !path.toLowerCase().startsWith(`/${locale.toLowerCase()}-`)
```
pieh, line 38: this line feels like something to handle `nl-NL` case? but then in other parts of the PR we are dropping usage of such prefixes in tests so I'm a bit confused here wether we are attempting to handle those cases or not?

edge-runtime/lib/response.ts
```diff
@@ -187,26 +181,12 @@ export const buildResponse = async ({
     // respect trailing slash rules to prevent 308s
     rewriteUrl.pathname = normalizeTrailingSlash(rewriteUrl.pathname, nextConfig?.trailingSlash)
 
-    const target = normalizeLocalizedTarget({ target: rewriteUrl.toString(), request, nextConfig })
-    if (target === request.url) {
-      logger.withFields({ rewrite_url: rewrite }).debug('Rewrite url is same as original url')
-      return
-    }
```
pieh, line 194: is this check not needed anymore?

tests/integration/edge-handler.test.ts
orinokai: updated in https://github.com/netlify/next-runtime/pull/2548/commits/fb8d9ea93cfdd41fa3fc2e5dcb9d787da818f8cf

edge-runtime/lib/response.ts
```diff
@@ -198,8 +198,8 @@ export const buildResponse = async ({
   }
 
   // If we are redirecting a request that had a locale in the URL, we need to add it back in
-  if (redirect && requestLocale) {
-    redirect = normalizeLocalizedTarget({ target: redirect, request, nextConfig, requestLocale })
+  if (redirect && locale) {
+    redirect = normalizeLocalizedTarget({ target: redirect, request, nextConfig })
```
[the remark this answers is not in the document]
orinokai: it was intentional, yes - see the comment below about removing `normalizeLocalizedTarget`

edge-runtime/lib/util.ts
orinokai: Thanks - have removed that part of the condition and added a couple of others that were transplanted from `normalizeLocalizedTarget`

edge-runtime/lib/response.ts
orinokai: I did some digging after your comment above about the `normalizeLocalizedTarget` params because I realised that I don't think we actually need `normalizeLocalizedTarget` at all. All it does is force the locale to be whatever was specified in the target path or the current locale. However, that's the not the behaviour we want - Next.js doesn't default to the current locale if you don't specify a locale in the rewrite/redirect path, it just redirects to whatever the target URL is.

orinokai: It seems to be the behaviour we need (from my manual testing) and all our tests pass. I'm going to run the full e2e test suite on it and double check.

pieh: I was thinking about https://github.com/netlify/next-runtime-minimal/pull/193 and the check was added to avoid looping apparently? Do we handle this in other way now / looping doesn't happen anymore?

orinokai: Ah, I see what you mean. No it's fine because we already check that condition on [Line 135](https://github.com/netlify/next-runtime/pull/2548/files/a7620fdd9009f25fb7820dd43828a30aa22ff434..da8215f0d9d05aef90526e957ced6447f8567c0c#diff-4f9f8de63d5504fcd1a98cf04a47b85e6759a68279a524e2713058bf4564a99eR135). The second check was because previously we may have mutated the target URL in `normalizeLocalizedTarget`, but we no longer use that.

orinokai:
Superseded by a scaled down version https://github.com/netlify/next-runtime/pull/2555

PR #2549, opened 2024-07-18
fix: apply type: module only to runtime modules

pieh:
## Description

We are setting entire request handler function dir as `type: 'module'` right now, which result in `ERR_REQUIRE_ESM` kind of errors when `.js` file which is actually CJS is being required.

## Tests

Integration and e2e test added using same fixture - both fail. Integration test output https://github.com/netlify/next-runtime/actions/runs/9992611474/job/27618021931?pr=2549#step:12:151 showing `ERR_REQUIRE_ESM` error (e2e only see 500 status, as error is hidden in function logs).

Second commit makes them pass.

### DIFF
--- a/src/build/content/server.ts
+++ b/src/build/content/server.ts
@@ -260,9 +260,9 @@ export const copyNextDependencies = async (ctx: PluginContext): Promise<void> =>
   await tracer.withActiveSpan('copyNextDependencies', async () => {
     const entries = await readdir(ctx.standaloneDir)
     const promises: Promise<void>[] = entries.map(async (entry) => {
-      // copy all except the package.json and distDir (.next) folder as this is handled in a separate function
+      // copy all except the distDir (.next) folder as this is handled in a separate function
       // this will include the node_modules folder as well
-      if (entry === 'package.json' || entry === ctx.nextDistDir) {
+      if (entry === ctx.nextDistDir) {
         return
       }
       const src = join(ctx.standaloneDir, entry)
--- a/src/build/functions/server.ts
+++ b/src/build/functions/server.ts
@@ -53,11 +53,20 @@ const copyHandlerDependencies = async (ctx: PluginContext) => {
       )
     }
 
+    // We need to create a package.json file with type: module to make sure that the runtime modules
+    // are handled correctly as ESM modules
+    promises.push(
+      writeFile(
+        join(ctx.serverHandlerRuntimeModulesDir, 'package.json'),
+        JSON.stringify({ type: 'module' }),
+      ),
+    )
+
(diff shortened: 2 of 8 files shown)


src/build/functions/server.ts
```diff
@@ -87,7 +87,7 @@ const writeHandlerManifest = async (ctx: PluginContext) => {
 
 const writePackageMetadata = async (ctx: PluginContext) => {
   await writeFile(
-    join(ctx.serverHandlerRootDir, 'package.json'),
+    join(ctx.serverHandlerRootDir, '.netlify', 'package.json'),
     JSON.stringify({ type: 'module' }),
   )
```
pieh, line 92: I'm not quite sure if this is a correct fix (or rather if this is full fix) - we definitely shouldn't be setting `type: 'module'` for entire handler directory in every case (which is what cause the problem right now). 

We definitely should either set it for our runtime modules we copy to handler (which is what this PR does right now) OR make runtime emit `.mjs` files and not `.js` ones to `dist`.

But part that I'm not sure about is wether `type: 'module'` should be set in root of handler in case user use this mode? I'll run some experiments here, but in case someone is familiar - I'm happy for any heads up on that.

pieh: So I did update it since this comment - Next.js does output their own `package.json` in root of standalone which we were ignoring when copying from it. With my changes we copy it over (this is really just copy of site's package.json so it inherit site's `type` essentially) + create package.json just for runtime modules (possibly this could be avoided if we emitted `.mjs` files for runtime modules, but my initial attempt at that resulted in problems so I didn't pursue this too much)

src/build/content/server.ts
```diff
@@ -260,9 +260,9 @@ export const copyNextDependencies = async (ctx: PluginContext): Promise<void> =>
   await tracer.withActiveSpan('copyNextDependencies', async () => {
     const entries = await readdir(ctx.standaloneDir)
     const promises: Promise<void>[] = entries.map(async (entry) => {
-      // copy all except the package.json and distDir (.next) folder as this is handled in a separate function
+      // copy all except the distDir (.next) folder as this is handled in a separate function
       // this will include the node_modules folder as well
-      if (entry === 'package.json' || entry === ctx.nextDistDir) {
+      if (entry === ctx.nextDistDir) {
```
pieh, line 265: we are no longer writing top level package.json (we write it just for runtime modules, so it no longer is top-level), hence this copies top-level package.json provided by Next.js (which is really just site's package.json) so it will "inherit" `type` from user's site

orinokai (approved):
This looks great 🚢

PR #2553, opened 2024-07-23
test: adjust setup for next@canary

pieh:
## Description

Ref: https://github.com/netlify/next-runtime/actions/runs/10050411516/job/27778277122?pr=2552#step:12:211 - attempting to fix tests setup for next@canary

## Tests

1. TODO

### DIFF
--- a/tests/utils/fixture.ts
+++ b/tests/utils/fixture.ts
@@ -54,6 +54,20 @@ async function installDependencies(cwd: string) {
 export const getFixtureSourceDirectory = (fixture: string) =>
   fileURLToPath(new URL(`../fixtures/${fixture}`, import.meta.url))
 
+// https://github.com/vercel/next.js/pull/67539 added more imports to "globals" modules which does have a side effect at import time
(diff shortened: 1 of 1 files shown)


orinokai (approved):
This looks good, but I wonder whether we should retain all our Next.js patching in one place because we already some patching here https://github.com/netlify/next-runtime/blob/rs/frp-761-imiddleware-locale/.github/workflows/test-e2e.yml#L150-L154

pieh:
> This looks good, but I wonder whether we should retain all our Next.js patching in one place because we already do some patching here https://github.com/netlify/next-runtime/blob/main/.github/workflows/test-e2e.yml#L150-L154

This is not for Vercel tests and we do some stuff in this module already to help in our integration test setup which is few lines below hack added in this PR ( https://github.com/netlify/next-runtime/pull/2553/files#diff-2e2db2b91b140546c5f0cdeb8770dbc305ba4a6a525667353b1fe33f2f6b59c6R76-R92 )

PR #2572, opened 2024-08-22
fix: update cache handler to accommodate changes in next@canary

pieh:
## Description

Adjusting for recent changes in `next@canary`:
 - (integration test setup only): resetting some global that Next starting to set to signify that global fetch was already patched (otherwise running tests for different fixture instances would not use fetch cache correctly as second fixture would skip fetch patching) - see:
   - https://github.com/vercel/next.js/pull/68193/files#diff-4c54e369ddb9a2db1eed95fe1d678f94c8e82c540204475d42c78e49bf4f223aR40
   - https://github.com/vercel/next.js/pull/68193/files#diff-4c54e369ddb9a2db1eed95fe1d678f94c8e82c540204475d42c78e49bf4f223aR813
 - (build and runtime): handling cache `kind` changes (`PAGE` -> `PAGES` and `ROUTE` -> `APP_ROUTE`) - see: https://github.com/vercel/next.js/pull/68602 (not the most immediately visible, but https://github.com/vercel/next.js/pull/68602/files#diff-cdee3ad3788ca17ac8dbaa342fe7e7ac4b1362cbc10b7bf22f7568302c995584R40-R47 is current enum for it)
 - (runtime): force setting `NODE_ENV=production` env var - https://github.com/vercel/next.js/pull/68193/files#diff-37243d614f1f5d3f7ea50bbf2af263f6b1a9a4f70e84427977781e07b02f57f1R49 this import indirectly requires `react` and at this point nothing sets `NODE_ENV` yet which results in loading development react version (which is later used forever due to being stored in `require.cache`). This later causes some conflicts as actual rendering will use production react version. Do note that `NODE_ENV` setting is being done in standalone output entry point that Next.js produces - https://github.com/vercel/next.js/blob/c3c38bd191ae0343cb1a449b00970ede0913ea8d/packages/next/src/build/utils.ts#L2221 but we aren't actually using this entrypoint, so setting it ourselves brings as closer to Next.js overall while also fixing dev/prod react problem)

## Tests

1. TODO


tests/prepare.mjs
```diff
@@ -44,14 +44,17 @@ const promises = fixtures.map((fixture) =>
...
     const { packageManager } = JSON.parse(await readFile(join(cwd, 'package.json'), 'utf8'))
     if (packageManager?.startsWith('pnpm')) {
       // We disable frozen-lockfile because we may have changed the version of Next.js
       cmd = `pnpm install --frozen-lockfile=false ${
         process.env.DEBUG || NEXT_VERSION !== 'latest' ? '' : '--reporter silent'
       }`
+    } else {
+      // npm is the default
+      cmd = `npm install --no-audit --progress=false --prefer-offline --legacy-peer-deps`
+      await rm(join(cwd, 'package-lock.json'), { force: true })
```
pieh, line 57: This is for local deving. Doesn't impact our CI runs, because lock files for fixtures are not commited

tests/utils/next-version-helpers.mjs
```diff
@@ -97,8 +96,17 @@ export async function setNextVersionInFixture(
           return
         }
         packageJson.dependencies.next = version
+
+        const { stdout } = await execaCommand(
+          `npm info next@${resolvedVersion} peerDependencies --json`,
+          { cwd },
+        )
+
+        const nextPeerDependencies = JSON.parse(stdout)
```
pieh, line 105: `next@canary` started to bump react(-dom) peer dependencies quite frequently, so instead of hardcoding react version - we just check what react version given next version needs

src/shared/cache-types.cts
```diff
@@ -1,58 +1,133 @@
...
-  | Exclude<
-      IncrementalCacheValue,
-      CachedRouteValue | CachedPageValue | IncrementalCachedAppPageValue
-    >
-  | NetlifyCachedRouteValue
-  | NetlifyCachedPageValue
-  | NetlifyCachedAppPageValue
+export type CachedFetchValueForMultipleVersions = Omit<CachedFetchValue, 'kind'> & {
+  kind: 'FETCH'
+}
```
pieh, line 71: This one is really just to satisfy TS. Next now uses `const enum` with string values and you can't assign string literal to it (even tho the value matches).

For other similar cases like this - i.e. `kind: 'PAGE' | 'PAGES'` or `kind: 'ROUTE' | 'APP_ROUTE'` - Next actually changed the string value from `PAGE` to `PAGES` and `ROUTE` to `APP_ROUTE` and we want to support both.

For our `CacheHandler` implementation we do receive appropriate kind used by Next.js in `.set` variant as it's Next.js that produce cache value to store.

For `.get` case we check what we stored in handle things appropriately for kinds before and after `kind` change and we don't need to make any decision which should be used there.

We do have to figure out which `kind` to actually use when seeding blobs from prerendered content, so that's the only case that need to decide which `kind` to use (and it's done based on Next.js version)

src/shared/cache-types.cts
```diff
@@ -1,58 +1,133 @@
...
+ * Used for storing in blobs and reading from blobs
+ */
 export type NetlifyCacheHandlerValue = MapCachedRouteValueToNetlify<CacheHandlerValue>
+
+/**
+ * Used for storing in blobs and reading from blobs
+ */
+export type NetlifyIncrementalCacheValue = NetlifyCacheHandlerValue['value']
+
+// type IncrementalCacheValueToMultipleVersions<T> = T extends CachedRouteValue ?
```
serhalp, line 93:
```suggestion
```

src/shared/cache-types.cts
```diff
@@ -1,58 +1,133 @@
...
+    ? IncrementalCachedPageValueForMultipleVersions
+    : T extends IncrementalCachedAppPageValue
+      ? IncrementalCachedAppPageValueForMultipleVersions
+      : T extends CachedFetchValue
+        ? CachedFetchValueForMultipleVersions
+        : T extends CacheHandlerValue
+          ? {
+              [K in keyof T]: IncrementalCacheValueToMultipleVersions<T[K]>
+            }
+          : T
```
serhalp, line 106: 🙈🙈🙈

serhalp (approved):
LGTM, heroic effort dealing with this madness!

tests/utils/fixture.ts
```diff
@@ -89,6 +89,7 @@ export const createFixture = async (fixture: string, ctx: FixtureTestContext) =>
   ) {
     // @ts-ignore fetch doesn't have _nextOriginalFetch property in types
     globalThis.fetch = globalThis._nextOriginalFetch || globalThis.fetch._nextOriginalFetch
+    delete globalThis[Symbol.for('next-patch')]
```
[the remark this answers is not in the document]
pieh: Yeah, I probably can incorporate first bullet point from PR's description which does link to specific code lines as code comment there

src/shared/cache-types.cts
pieh: Yeah, this whole thing is quite a mess. It doesn't help that there are 2 type variants generally created there:
 1. (new)mapping types from currently used `next` version we have for types to not that version specific but instead have kind of union that "work" for all the next versions we support (which is what this inline comment highlights)
 2. Mapping those types to how we actually store data in blobs (i.e. instead of using `Buffer` we use `string` for storage or adding extra information about `revalidate` that is used to figure out cacheable route handler response maxage).

For 1. I did try having multiple `next` version in deps and import types from there and have some kind of union on them, but setup for it was pretty messy:
 - I could not use npm alias and have multiple next versions there because their react(-dom) peerDeps could not be satisfied at the same time)
 - Using unions there was still not enough because next@canary now uses const enum so I would have to do messy things either way.

So I don't think we should treat those types necessarily as safeguard that we are using them correctly for Next purposes (unfortunately :( ).

The option where this could maybe work is if we would have actual 2 (or more) variants of `CacheHandler` implementations, but because those would be largely the same we would either have almost copies of them or maybe some abstraction that would be more difficult to read :/

tests/utils/fixture.ts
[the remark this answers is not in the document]
pieh: Added comment in https://github.com/netlify/next-runtime/pull/2572/commits/d7b3492c5c5f3bd3851ea9a7053e32ccf00badaa

orinokai (approved):
LGTM and 👏🏼 for dealing with this all. It's also made me think we should discuss our options regarding long-term support of different Next versions. I feel like there might be trade-offs we can make that could avoid incrementally adding complexity like this. Will start a doc...

tests/utils/next-version-helpers.mjs
```diff
@@ -97,8 +96,17 @@ export async function setNextVersionInFixture(
           return
         }
         packageJson.dependencies.next = version
+
+        const { stdout } = await execaCommand(
+          `npm info next@${resolvedVersion} peerDependencies --json`,
```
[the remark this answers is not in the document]
serhalp: Yes it's about the need to do this at all, not about the solution you went with! :)

PR #2573, opened 2024-08-22
fix: ignore middleware rewrite when also redirecting

serhalp:
## Description

The (undocumented, untested) Next.js behaviour appears to be for a middleware redirect to take precedence over a middleware rewrite when both are present at once.

I was going to swap the order of evaluation of rewrites and redirects in `lib/response` but then decided to go with a simpler "pretend there was no rewrite if there's a redirect" approach. I'm open to feedback here.

### Documentation

N/A

## Tests

Added an integration test that was failing before this change. You can also create [a simple repro site like this](https://github.com/serhalp/next-repro-x-middleware-rewrite) and see that it behaves differently with `next dev` than via Netlify.

FRP-1221


serhalp:
(All the failures are for next@canary. We're working on fixing that, but it isn't related to this PR.)

pieh (approved):

serhalp:
Published as `﻿@netlify/plugin-nextjs@5.7.0-canary.0`.

PR #2583, opened 2024-09-06
fix: generate correct import path when 'src' directory is used and middleware imports wasm module

pieh:
## Description

Fixes cases of bundling middleware edge function when `src` directory is used and middleware is using WASM

## Tests

Adding test case failing without any logic changes (new fixture - `wasm-src` is almost copy of `wasm`, except for putting middleware, app and pages under `src` directory).

https://github.com/netlify/next-runtime/actions/runs/10734892318/job/29771032475?pr=2583#step:12:208 shows failure without any logic changes in runtime

Fixes https://linear.app/netlify/issue/FRP-1279/failed-to-bundle-nextjs-middleware


serhalp (approved):
Good catch!

PR #2587, opened 2024-09-12
test(e2e): fix failures due to latest and canary releases

serhalp:
## Description

[v14.2.10](https://github.com/vercel/next.js/releases/tag/v14.2.10) and [v15.0.0-canary.147](https://github.com/vercel/next.js/releases/tag/v15.0.0-canary.147) included https://github.com/vercel/next.js/pull/69802 which added `private` back into `no-cache,no-store` `cache-control` headers in some cases.

There are still failures after this fix but they are caused by two (three?) additional unrelated issues.

### Documentation

N/A

## Tests

N/A

N/A

### DIFF
--- a/tests/e2e/cli-before-regional-blobs-support.test.ts
+++ b/tests/e2e/cli-before-regional-blobs-support.test.ts
@@ -22,12 +22,10 @@ test('should serve 404 page when requesting non existing page (no matching route
   // https://github.com/vercel/next.js/pull/69802 made changes to returned cache-control header,
   // after that (14.2.10 and canary.147) 404 pages would have `private` directive, before that it
   // would not
-  const shouldHavePrivateDirective = nextVersionSatisfies(
-    '>=14.2.10 <15.0.0 || >=15.0.0-canary.147',
-  )
+  const shouldHavePrivateDirective = nextVersionSatisfies('^14.2.10 || >=15.0.0-canary.147')
   expect(headers['netlify-cdn-cache-control']).toBe(
     (shouldHavePrivateDirective ? 'private, ' : '') +
-      'private, no-cache, no-store, max-age=0, must-revalidate, durable',
+      'no-cache, no-store, max-age=0, must-revalidate, durable',
   )
   expect(headers['cache-control']).toBe(
     (shouldHavePrivateDirective ? 'private,' : '') + 'no-cache,no-store,max-age=0,must-revalidate',
--- a/tests/e2e/cli-before-regional-blobs-support.test.ts
+++ b/tests/e2e/cli-before-regional-blobs-support.test.ts
@@ -1,5 +1,6 @@
 import { expect } from '@playwright/test'
 import { test } from '../utils/playwright-helpers.js'
+import { nextVersionSatisfies } from '../utils/next-version-helpers.mjs'
 
 test('should serve 404 page when requesting non existing page (no matching route) if site is deployed with CLI not supporting regional blobs', async ({
   page,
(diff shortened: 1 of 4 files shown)


tests/e2e/cli-before-regional-blobs-support.test.ts
```diff
@@ -18,8 +19,17 @@ test('should serve 404 page when requesting non existing page (no matching route
...
+  // https://github.com/vercel/next.js/pull/69802 made changes to returned cache-control header,
+  // after that (14.2.10 and canary.147) 404 pages would have `private` directive, before that it
+  // would not
+  const shouldHavePrivateDirective = nextVersionSatisfies(
+    '>=14.2.10 <15.0.0 || >=15.0.0-canary.147',
+  )
   expect(headers['netlify-cdn-cache-control']).toBe(
-    'no-cache, no-store, max-age=0, must-revalidate, durable',
+    (shouldHavePrivateDirective ? 'private, ' : '') +
+      'private, no-cache, no-store, max-age=0, must-revalidate, durable',
```
pieh, line 30: We conditionally adding `private`, before that check
```suggestion
      'no-cache, no-store, max-age=0, must-revalidate, durable',
```

tests/e2e/simple-app.test.ts
```diff
@@ -220,8 +220,12 @@ test('requesting a non existing page route that needs to be fetched from the blo
   expect(await page.textContent('h1')).toBe('404 Not Found')
 
   // https://github.com/vercel/next.js/pull/66674 made changes to returned cache-control header,
-  // before that 404 page would have `private` directive, after that it would not
-  const shouldHavePrivateDirective = !nextVersionSatisfies('>=14.2.4 <15.0.0 || >=15.0.0-canary.24')
+  // before that 404 page would have `private` directive, after that (14.2.4 and canary.24) it
+  // would not ... and then https://github.com/vercel/next.js/pull/69802 changed it back again
+  // (14.2.10 and canary.147)
+  const shouldHavePrivateDirective = nextVersionSatisfies(
+    '<14.2.4 || >=14.2.10 < 15 || <15.0.0-canary.24 || >= 15.0.0-canary.147',
```
pieh, line 227: This condition is different than one used in other places - because this one has `<15.0.0-canary.24` without lower limit, it will actually match `14.2.5` for example (tho we won't be really hitting those edge cases with the version we test setup as all currently tested version would have `private` directive (`13.5.1`, latest which is now `14.2.11` and canary which is now `15.0.0-canary.157`)

This would work (and then probably same condition should be used across the board? maybe we centralize it and add `shouldHavePrivateDirective` to same helper module that already has `nextVersionSatisfies` helper?)
```suggestion
    '<14.2.4 || >=14.2.10 <15.0.0-canary.24 || >= 15.0.0-canary.147',
```
Or we use the condition used elsewhere with understanding that those earlier canary versions (`>=15.0.0-canary.0 <15.0.0-canary.24`) generally should have `private` directive but we skip those because we don't really test those anymore and keeping condition cleaner and easier to reason with

serhalp: Interesting. I've read those [semver docs](https://www.npmjs.com/package/semver) a hundred times and I still keep finding things I've misinterpreted. I thought prerelease range specifiers could never match stable releases (and prereleases with a different prefix).

That clearly wasn't right:
```
$ npm i -g semver
$ semver -r '<15.0.0-canary.24' '14.2.5' '15.2.5' '15.0.0-canary.20' '15.0.0-canary.30'
14.2.5
15.0.0-canary.20
```

serhalp: Fascinating:
```
$ semver -p -r '<15.0.0' 15.0.0-canary.150
15.0.0-canary.150
```

Maybe this was obvious to you, but I think this behaved kind of like JS's comparison operators with values like `NaN`:
```
> 3 < NaN
false
> 3 > NaN
false

Comment (pieh) [reply tests/e2e/simple-app.test.ts:227]: It's a bit more nuanced:
```
semver -r '<15.0.0-canary.24' '14.2.5' '14.2.5-canary.5' '15.2.5'  '15.0.0-canary.20' '15.0.0-canary.30'
14.2.5
15.0.0-canary.20
```

The `14.2.5-canary.5` wouldn't match but `14.2.5` would ... unless you would use `includePrerelease`:
```
semver --include-prerelease -r '<15.0.0-canary.24' '14.2.5' '14.2.5-canary.5' '15.2.5'  '15.0.0-canary.20' '15.0.0-canary.30'
14.2.5-canary.5
14.2.5
15.0.0-canary.20
```

serhalp: > maybe we centralize it and add shouldHavePrivateDirective to same helper module that already has nextVersionSatisfies helper?

I had thought about that, but strangely we'd need two different versions of it (some assertions are affected by the change a few months ago, some are affected by both changes) and I couldn't come up with a reasonable name for these functions I decided this was a case of "a poor abstraction is worse than no abstraction".

serhalp: > It's a bit more nuanced [...]

Somehow I knew about that nuance but not the more fundamental one above!

pieh (approved):

PR #2636, opened 2024-10-01
fix: handle middleware redirects to default locale and same path

pieh:
## Description

We've been capturing original request locale and forcing it on redirects if redirect didn't contain locale. This was originally introduced in https://github.com/netlify/next-runtime-minimal/pull/276 . Since that PR we also had couple of iteration of i18n handling and the previous handling seems no longer needed and in fact it actually does break redirect functionality when `nextUrl.locale = '<default_locale>'` is being used - in which case the `location` header for redirect would result in path that doesn't contain any locale (due to https://github.com/vercel/next.js/blob/eecf90c09ae2e39965191427776713c6b636e50f/packages/next/src/shared/lib/router/utils/add-locale.ts#L15-L17 ) and would fall into handling of adding original locale back.

## Tests

Initially failing test case added

https://linear.app/netlify/issue/FRB-1363/nextjs-i18n-middleware-redirect-not-working


edge-runtime/lib/response.ts
```diff
@@ -197,11 +195,10 @@ export const buildResponse = async ({
...
   }
 
-  // If we are redirecting a request that had a locale in the URL, we need to add it back in
-  if (redirect && requestLocale) {
-    redirect = normalizeLocalizedTarget({ target: redirect, request, nextConfig, requestLocale })
+  if (redirect) {
+    redirect = normalizeLocalizedTarget({ target: redirect, request, nextConfig })
     if (redirect === request.url) {
-      logger.withFields({ rewrite_url: rewrite }).debug('Rewrite url is same as original url')
+      logger.withFields({ redirect_url: redirect }).debug('Redirect url is same as original url')
```
pieh, line 201: this seems like was previously c&p-ed from rewrite handling so log fields/msg was off

PR #2645, opened 2024-10-04
fix: handle non ASCII characters in cache-tag headers

lukasholzer:
## Description

Fixes https://linear.app/netlify/issue/FRB-1352/handle-non-ascii-characters-in-cache-tag-headers

Handles non ASCII characters in cache-tag headers by URI encoding them (does not encode / or valid characters just the non ASCII ones)

## Tests

Added e2e test case for cacheable page router pages with non-ascii paths

1. TODO

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -25,6 +25,7 @@ export default defineConfig({
     extraHTTPHeaders: {
       /* Add debug logging for netlify cache headers */
       'x-nf-debug-logging': '1',
+      'x-next-debug-logging': '1',
     },
   },
   timeout: 10 * 60 * 1000,
--- a/src/run/handlers/cache.cts
+++ b/src/run/handlers/cache.cts
@@ -147,7 +147,9 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
       ) {
         // pages router doesn't have cache tags headers in PAGE cache value
         // so we need to generate appropriate cache tags for it
-        const cacheTags = [`_N_T_${key === '/index' ? '/' : key}`]
+        // encode here to deal with non ASCII characters in the key
+
+        const cacheTags = [`_N_T_${key === '/index' ? '/' : encodeURI(key)}`]
         requestContext.responseCacheTags = cacheTags
       }
     }
@@ -341,7 +343,8 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
       if (data?.kind === 'PAGE' || data?.kind === 'PAGES') {
         const requestContext = getRequestContext()
         if (requestContext?.didPagesRouterOnDemandRevalidate) {
-          const tag = `_N_T_${key === '/index' ? '/' : key}`
+          // encode here to deal with non ASCII characters in the key
+          const tag = `_N_T_${key === '/index' ? '/' : encodeURI(key)}`
           getLogger().debug(`Purging CDN cache for: [${tag}]`)
           requestContext.trackBackgroundWork(
             purgeCache({ tags: [tag] }).catch((error) => {
--- a/src/run/handlers/cache.cts
+++ b/src/run/handlers/cache.cts
@@ -139,7 +139,7 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
       cacheValue.kind === 'APP_ROUTE'
     ) {
       if (cacheValue.headers?.[NEXT_CACHE_TAGS_HEADER]) {
-        const cacheTags = (cacheValue.headers[NEXT_CACHE_TAGS_HEADER] as string).split(',')
+        const cacheTags = (cacheValue.headers[NEXT_CACHE_TAGS_HEADER] as string).split(/,|%2c/gi)
         requestContext.responseCacheTags = cacheTags
       } else if (
         (cacheValue.kind === 'PAGE' || cacheValue.kind === 'PAGES') &&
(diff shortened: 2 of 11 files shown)


orinokai (commented):
I wonder whether we might need to expand the encoding? AFAIK, `encodeURL` will leave reserved characters unencoded, which means a URL like `http://foo.com/resource,123` would emit a cache tag with a comma in it. Should we use `encodeURIComponent` instead?

pieh:
> I wonder whether we might need to expand the encoding? AFAIK, `encodeURL` will leave reserved characters unencoded, which means a URL like `http://foo.com/resource,123` would emit a cache tag with a comma in it. Should we use `encodeURIComponent` instead?

I don't think it's that simple - I was running some tests here and it seems to me that comma case need to be handled on different level than current change - current change only applies to page router, because app router already actually handles it fine (I did add test cases in last commit here for it)

The comma problem however applies right now to both Pages and App router in similar way, so just swapping currently added `encodeURI` with `encodeURIComponent` would be insufficient to fully support it.

Also `encodeURIComponent` would replace `/` with encoded character so this would be bigger departure from what we currently generate vs using `encodeURI` - maybe we would only want to target comma additionally here to limit changes? Or we treat those completely as implementation detail that can change at any time, then using that would be fine (tho debugging might be a bit more annoying with so many extra characters being encoded in header we produce)

pieh:
I was looking into comma case, and this is getting quite not straight forward with our usage of `split(',')` in places like:
https://github.com/netlify/next-runtime/blob/0b74e135b213de07e9b31bf27f517becd125c212/src/run/handlers/cache.cts#L142
https://github.com/netlify/next-runtime/blob/0b74e135b213de07e9b31bf27f517becd125c212/src/run/handlers/cache.cts#L422

it basically looks like we can't really encode this specifically because we can't distinguish between between comma being a separator and being actual part of one of the items.

It does feel like we kind of have to apply same handling and actually produce 2 (or more) cache tag if actual tag contains comma and instead make sure the way we interact with tag manifest blobs and purge cdn API also need to split on it - this might mean accidentally purging more content than desired but because we don't handle entire pipeline at least for app router - we just can't make it work well in this scenario in which case it would make sense to arrive and similar handling for both app and pages router?

pieh:
I was doing much more tests with comma and this getting super confusing and annoying ...

cache tags for app router are generated differently when prerendered and when rendering on demand later ...

in prerendered comma does get encoded:
```json5
{
  path: '/product/事前レンダリングされていない,test'
  responseCacheTags: [
    '_N_T_/layout',
    '_N_T_/product/layout',
    '_N_T_/product/[slug]/layout',
    '_N_T_/product/[slug]/page',
    // notice '%2Ctest' which is ',test'
    '_N_T_/product/%E4%BA%8B%E5%89%8D%E3%83%AC%E3%83%B3%E3%83%80%E3%83%AA%E3%83%B3%E3%82%B0%2Ctest' 
  ]
}
```
while non-prerendered gets this:
```json5
{
  path: '/product/事前レンダリングされていない,test',
  responseCacheTags: [
    '_N_T_/layout',
    '_N_T_/product/layout',
    '_N_T_/product/[slug]/layout',
    '_N_T_/product/[slug]/page',
    // in here comma was not encoded so our splitting on ',' made 'test' into separate cache tag
    '_N_T_/product/%E4%BA%8B%E5%89%8D%E3%83%AC%E3%83%B3%E3%83%80%E3%83%AA%E3%83%B3%E3%82%B0%E3%81%95%E3%82%8C%E3%81%A6%E3%81%84%E3%81%AA%E3%81%84',
    'test'
  ]
}
```

I'm thinking that maybe we should split on both `,` and `%2c` to get ~consistent handling - even if it is not the exactly correct one, because it will always generate 2 cache tags if cache tag contain comma (encoded or not) and then some unrelated purge _might_ purge those too, but because we are not getting encoded comma consistently - we can't really handle this in any other way?

orinokai:
> I'm thinking that maybe we should split on both , and %2c to get ~consistent handling

Maybe we could make the splitting a bit smarter and continue to split on comma, but only if the comma is followed by `_N_T_`?

pieh:
> Maybe we could make the splitting a bit smarter and continue to split on comma, but only if the comma is followed by `_N_T_`?

fetch cache tags are not prepended with `_N_T_` and instead we get them as they are set by users - otherwise that would be awesome idea to be able to accurately distinguish between entries (that they all must start with `_N_T_`

orinokai:
> fetch cache tags are not prepended with _N_T_

true, but the fetch cache tags are always prepended in the `cacheValue.headers[NEXT_CACHE_TAGS_HEADER]` string, so we can start splitting by `, _N_T_` after the first `_N_T_`

pieh:
Ughhh. There is another layer here with comma in fetch cache tags ... sigh
```
next: { tags: ["tagstart,tagend"] },
```
and
```
next: { tags: ["tagstart","tagend"] },
```
Will produce same thing (both prerendered and generated on demand) - this would still work with current code in this PR (albeit it would always treat both of above as the last one in practice), but smart splitting idea (which I hoped would result in more exact purges) wouldn't help here - how should:
```
revalidateTag('tagstart')
revalidateTag('tagstart,tagend')
```
those work when you can't distinguish them? the only thing that does work is splitting on comma again - but then how do you handle
```
revalidatePath('/foo,baz')
```
?

because we don't know which one (`revalidateTag` or `revalidatePath`) user called and both result in calling `CacheHandler.revalidateTag` and we would need different behavior based on wether it was path or a tag 

Now I'm even more in favor of current approach that just works with both of those cases (it might purge more than needed if there is overlap with different cache tag after splitting on both comma and encoded comma, but this feels better than not being able to invalidate something at all)

orinokai:
I think you're right - the current solution is the best because the impact of any problems is limited - the worst scenario I can think of is that a URL has a comma in it and, after splitting, the first part matches a layout, so a whole bunch of unintended pages get purged, but that's an edge case and why would you use a URL scheme like that anyway?!

orinokai (approved):
lgtm, as discussed

Issue #2627, opened 2024-09-27
[v15 canary] Dynamic routes with `fallback: true` are failing with cache error


orinokai:
Non-prerendered paths on dynamic routes no longer work when the `fallback: true` option is used. The response is a 500 and the error is as follows: `cache entry required but not generated`.

The following test:
https://github.com/vercel/next.js/blob/v15.0.0-canary.171/test/e2e/prerender-crawler.test.ts#L61-L69

...fails for the following fixture:
https://github.com/vercel/next.js/blob/v15.0.0-canary.171/test/e2e/prerender-crawler.test.ts#L10-L51

We have tests in place for dynamic routes, so it's likely this is solely related to `fallback: true`, which should render a fallback component while the page is generated in the background.

The error appears to have been introduced in canary 114, but this was a large update that could have obscured a previous issue.

```
FAIL test/e2e/prerender-crawler.test.ts (78.216 s)
  Prerender crawler handling
    ✓ should return prerendered page for correctly (901 ms)
    ✕ should return fallback for non-crawler correctly (370 ms)
    ✓ should block for crawler correctly (3283 ms)

  ● Prerender crawler handling › should return fallback for non-crawler correctly

    FetchError: Invalid response body while trying to fetch https://66d298ca3b06cc9fe8162eff--next-e2e-tests.netlify.app/blog/test-1: incorrect header check

      at Gunzip.<anonymous> (../node_modules/.pnpm/node-fetch@2.6.7_encoding@0.1.13/node_modules/node-fetch/lib/index.js:400:12)
```

### Data

_The following is parsed automatically by the Next.js repo e2e test report generator._

test: Prerender crawler handling should return prerendered page for correctly 
reason: Non-prerendered dynamic routes with fallback: true are failing with cache error

PR #2649, opened 2024-10-08
fix: create cache entries for fallback pages to support next@canary

pieh:
## Description

https://github.com/vercel/next.js/pull/68603/files#diff-6f4291cc2bfc5073fdca12a014011769e840ee68583db1468acef075f037015aL2826 this change moved from serving fallback directly from fs to using cache-handler

We were not creating cache entries for fallbacks which was resulting in `Error: invariant: cache entry required but not generated` errors as Next now expected those fallbacks to exist.

This adds handling of fallback pages at build time and create cache entries for them which fixes above error.

Additionally - before that change in Next.js when fs was used to serve fallback - we were wrongly applying permament caching header for it (because of handling we have for fully static pages that is also using fs reads). This adds more nuance to that logic to not apply permament caching headers for fallback html reads

## Tests

Test cases added to e2e

https://linear.app/netlify/issue/FRB-1360/[v15-canary]-dynamic-routes-with-fallback-true-are-failing-with-cache
fixes https://github.com/netlify/next-runtime/issues/2627


src/run/next.cts
```diff
@@ -96,13 +117,17 @@ export async function getMockedRequestHandlers(...args: Parameters<typeof getReq
       } catch (error) {
         // only try to get .html files from the blob store
         if (typeof path === 'string' && path.endsWith('.html')) {
+          const fsBlobsManifest = await getFSBlobsManifest()
+
           const store = getRegionalBlobStore()
-          const relPath = relative(resolve('.next/server/pages'), path)
+          const relPath = relative(resolve(join(fsBlobsManifest.outputRoot, '/server/pages')), path)
```
pieh, line 123: I think this was not actually working correctly if user had non-default `distDir` (aka not `.next`) and also likely not working in monorepos

src/run/headers.ts
```diff
@@ -272,7 +272,10 @@ export const setCacheControlHeaders = (
 }
 
 export const setCacheTagsHeaders = (headers: Headers, requestContext: RequestContext) => {
-  if (requestContext.responseCacheTags) {
+  if (
+    requestContext.responseCacheTags &&
+    (headers.has('cache-control') || headers.has('netlify-cdn-cache-control'))
```
pieh, line 277: this is to prevent setting tags on uncacheable responses when fallback in next@canary is served

src/run/next.cts
orinokai: this makes sense, however the `outputRoot` seems to have nothing to do with blob manifest so I'm wondering if there should be a different way to specify this value?

orinokai (commented):
@serhalp and I have just been through the PR and generally it seems to do what it needs to, but we were wondering about the added complexity of having the fallback paths read from to the blob manifest file, rather than just having this check the blob store?

src/build/content/prerendered.ts
```diff
@@ -195,7 +207,41 @@ export const copyPrerenderedContent = async (ctx: PluginContext): Promise<void>
...
+                const key = routeToFilePath(posixJoin(locale, route))
+                const value = await buildPagesCacheValue(
+                  join(ctx.publishDir, 'server/pages', key),
+                  shouldUseEnumKind,
+                  true, // there is no corresponding json file for fallback, so we are skipping it for this entry
+                )
+                // Netlify Forms are not support and require a workaround
+                if (value.kind === 'PAGE' || value.kind === 'PAGES' || value.kind === 'APP_PAGE') {
+                  verifyNetlifyForms(ctx, value.html)
+                }
```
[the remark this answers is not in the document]
pieh: Yeah, I'll drop that completely form here - we actually already are doing this right now because of https://github.com/netlify/next-runtime/blob/5f5ec0651726abf1027319707beaa1580a0c5dd8/src/build/content/static.ts#L35 so this was just doing same check again on same content

This was just copied from above code heh

pieh:
> but we were wondering about the added complexity of having the fallback paths read from to the blob manifest file, rather than just having this check the blob store?

How about I change things so the `.html` blobs we create (not cache entries as those are unrelated) in https://github.com/netlify/next-runtime/blob/5f5ec0651726abf1027319707beaa1580a0c5dd8/src/build/content/static.ts#L18-L43 instead of being just `html` string be:

```ts
type HtmlBlob {
  html: string
  isFallback: boolean
}
```

Then we get away of the need to create the manifest that is currently created for the most part (except for `outputRoot` stuff that I think I will fully drop from this PR and keep hardcoded `.next` there and treat this part as out-of-scope for this PR) as we will correlate things at build time instead of at runtime.

The main reason (which is not great one heh) that I produced blobs fallback manifest, because we were handling `prerender-manifest.json` in different place than handling static html blobs and I could handle generating manifest at the same time as creating cache entries (for newer next versions) in single go - but to do above I will have to decouple it and handle prerender-manifest in 2 places separately (which is fine, I still probably can have some shared code there, just call it from multiple places)

src/run/next.cts
pieh: given https://github.com/netlify/next-runtime/pull/2649#issuecomment-2416419879 - I'll drop this change and keep hardcoded `.next` here as out of scope for this PR - I still think this is an issue, but I will not try to fix it in same PR that introduces creation of blob cache entries for fallbacks

orinokai:
> I will have to decouple it and handle prerender-manifest in 2 places separately (which is fine, I still probably can have some shared code there, just call it from multiple places)

Great, this sounds like a good solution. Thanks @pieh.

src/run/handlers/request-context.cts
```diff
@@ -11,7 +11,7 @@ export type RequestContext = {
   responseCacheGetLastModified?: number
   responseCacheKey?: string
   responseCacheTags?: string[]
-  usedFsRead?: boolean
+  usedFsReadForNonFallback?: boolean
```
orinokai, line 14: not blocking, but consider calling this something like `usedFsReadForStaticRoute`

orinokai (approved):
This is awesome, LGTM (one minor suggestion, but take it or leave it)

src/run/handlers/request-context.cts
pieh: I think I'll merge this as-is to get this over with. 

Minor side comment is that it hopefully now is actually functionally equivalent ( `usedFsReadForNonFallback` <=> `usedFsReadForStaticRoute`) but there's a chance there is another case here that we will be trying to handle in the future :D In this case I can at least say that `usedFsReadForNonFallback` name is actually correct because it literally check if html is fallback or not, but I am actually not 100% sure that all non-fallbacks are actually "StaticRoutes" heh

PR #2666, opened 2024-10-14
fix: adjust cache-control handling for next@15.0.0-canary.187

pieh:
## Tests

1. TODO

https://linear.app/netlify/issue/FRB-1385/next1500-canary187-produces-invalid-cache-control-somtimes


src/run/headers.ts
```diff
@@ -54,7 +54,10 @@ const generateNetlifyVaryValues = ({
 }
 
 const getHeaderValueArray = (header: string): string[] => {
-  return header.split(',').map((value) => value.trim())
+  return header
+    .split(',')
+    .map((value) => value.trim())
+    .filter(Boolean)
```
pieh, line 60: https://github.com/vercel/next.js/pull/71159/files#diff-ae41ee59d88093c9c0551e12f654b1b99e4211b1c9fa6dc1471febb0361004e7R21 this can now produce invalid value like `s-maxage=31536000, ` so we remove ~falsy/empty directives, so at least on our end we produce valid value.

Hopefully they fix their stuff in which this added `.filter` would do nothing (so it also won't hurt to have it always)

orinokai (approved):
this looks fine to me - the tests could be simplified by only checking for the `max-age` and `durable` directives without having the next version check (since even if the `stale-while-revalidate` header was missing in the previous version it's kind of redundant anyway), but it doesn't hurt to be explicit so lgtm

PR #2671, opened 2024-10-16
ci: fix next.js repo tests workflow for 13.5.1

serhalp:
The latest Ubuntu (24.04) doesn't work out of the box with older versions of playwright.

next@latest and next@canary have been updated: https://github.com/vercel/next.js/pull/71312.

13.5.1 has not been, so let's keep using 22.04 just for that


orinokai (approved):
That looks like it will do the job 👍🏼

PR #2673, opened 2024-10-16
ci: fix e2e report workflow missing `netlify` CLI

serhalp:
Ubuntu 24.04 (now ubuntu-latest) has removed the `netlify` CLI :(

We were installing it manually in some CI workflows, but not the e2e report one, which started failing.

This fixes that by pulling in the easy parts of https://github.com/netlify/next-runtime/pull/2484, i.e. using netlify-cli as a dev dep where possible. See inline comment for further work needed.


.github/workflows/e2e-report.yml
```diff
@@ -56,4 +56,4 @@ jobs:
       - name: Deploy e2e page
         if: success()
         run: |
-          ntl deploy --build --cwd e2e-report
+          npx netlify deploy --build --cwd e2e-report
```
serhalp, line 59: All the `npx` changes are because when binaries are installed locally, GitHub doesn't automatically add `./node_modules/.bin` to the `PATH`. Just using `npx` is the right approach I think.

.github/workflows/run-tests.yml
```diff
@@ -204,14 +204,13 @@ jobs:
...
-      - name: 'Netlify Login'
+      - name: 'Prepare Netlify CLI'
         env:
           NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
         run: |
-          # TODO(serhalp) Unpin version once
-          # https://netlify.slack.com/archives/C07686YAY13/p1718139257978049 is resolved.
-          npm i -g netlify-cli@17.26.0
-          netlify login
+          # Control netlify-cli as a regular dev dep but expose it globally for test fixtures to use
```
lukasholzer, line 211: NIT: you could pull this out into a reusable action in github actions.


Something like `.github/actions/netlify`

```

name: 'Install Netlify CLI'
description: some description
runs:
  using: 'composite'
  steps:
    -  run: |
          # Control netlify-cli as a regular dev dep but expose it globally for test fixtures to use
          npm install -g "netlify-cli@$(npm list --json --depth=0 netlify-cli | jq -r ".dependencies[\"netlify-cli\"].version")"
          npx netlify login
```

lukasholzer (approved):
LGTM

PR #2698, opened 2024-10-24
test: skip reporting CanaryOnly failures for stable version tests

pieh:
## Description

This is grab-back of various improvements for running Vercel tests:
 - we are now able to collect build logs which mean we should no longer NEED to skip those tests to prevent failures that are only related to CLI and not actual functionality not working - if the test is testing just the output - it makes sense to skip it to avoid wasting time on test that is not actually testing anything related to Netlify (I actually do skip one of those here, despite the test now passing)
 - we are able to recognize if test is using Canary only feature ( https://github.com/vercel/next.js/issues/71307 ) and skip processing results if that was cause for build/deploy failure as this test is just not applicable to stable Next versions
 - lower shards we use after move to not hog all the available runners
 - minor adjustment to `run-local-test` so that it doesn't require you locally checkout repo to actually be in `opennextjs-netlify` and instead just use `pwd` to locate helper files we copy or use  

## Tests

N/A

### DIFF
--- a/tests/netlify-deploy.ts
+++ b/tests/netlify-deploy.ts
@@ -146,7 +146,7 @@ export class NextDeployInstance extends NextInstance {
       this._url = url
       this._parsedUrl = new URL(this._url)
       this._deployId = deployID
-      this._cliOutput = deployRes.stdout + deployRes.stdout
+      this._cliOutput = deployRes.stdout + deployRes.stderr
       require('console').log(`Deployment URL: ${this._url}`)
 
       const [buildLogsUrl] =
--- a/tests/netlify-deploy.ts
+++ b/tests/netlify-deploy.ts
@@ -6,15 +6,6 @@ import { tmpdir } from 'node:os'
 import path from 'path'
 import { NextInstance } from './base'
 
-type NetlifyDeployResponse = {
-  name: string
-  site_id: string
-  site_name: string
-  deploy_id: string
-  deploy_url: string
-  logs: string
-}
-
 async function packNextRuntimeImpl() {
   const runtimePackDir = await fs.mkdtemp(path.join(tmpdir(), 'opennextjs-netlify-pack'))
 
@@ -133,7 +124,7 @@ export class NextDeployInstance extends NextInstance {
 
     const deployRes = await execa(
       'npx',
-      ['netlify', 'deploy', '--build', '--json', '--message', deployTitle ?? ''],
+      ['netlify', 'deploy', '--build', '--message', deployTitle ?? ''],
       {
         cwd: this.testDir,
         reject: false,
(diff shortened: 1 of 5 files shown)


tests/netlify-deploy.ts
```diff
@@ -133,7 +124,7 @@ export class NextDeployInstance extends NextInstance {
 
     const deployRes = await execa(
       'npx',
-      ['netlify', 'deploy', '--build', '--json', '--message', deployTitle ?? ''],
```
serhalp, line 136: there's a weird bug in `ntl deploy --json` that makes debugging failures annoying anyway so I like this change :)

(I should write it up but I don't know how to reproduce it 😞)

tests/netlify-deploy.ts
```diff
@@ -142,17 +133,29 @@ export class NextDeployInstance extends NextInstance {
...
+        throw new Error('Could not extract the URL from the build logs')
+      }
+      const [deployID] = new URL(url).host.split('--')
+      this._url = url
       this._parsedUrl = new URL(this._url)
-      this._deployId = data.deploy_id
-      require('console').log(`Deployment URL: ${data.deploy_url}`)
-      require('console').log(`Logs: ${data.logs}`)
+      this._deployId = deployID
+      this._cliOutput = deployRes.stdout + deployRes.stdout
```
serhalp, line 149: 🙃 is this a typo? did you mean stdout + stderr?

tools/deno/junit2json.ts
```diff
@@ -117,6 +117,13 @@ function junitToJson(xmlData: { testsuites: JUnitTestSuites }): Array<TestSuite>
       if (skippedTestsForFile?.some(({ name }) => name === testCase['@name'])) {
         continue
       }
+
+      // skip reporting on tests that even fail to deploy because they rely on experiments not available
+      // in currently tested version
+      if (testCase.failure?.includes('CanaryOnlyError')) {
+        continue
+      }
```
serhalp, line 125: Doing this here is a little funky because we report test results in 3(?) different places and this is just one of them. So now we'll have a mismatch between the Travis CI summary thing, the actual raw test output, and the e2e report built from this JSON (previously we also had a fourth which was the Slack notification).

Maybe this is fine for now, but I wonder if there's a more direct way to configure the test runner the same way Vercel's CI does? I would imagine it's some slight env var change we need to make or something 🤷🏼.

serhalp (approved):
Awesome, thanks for making these improvements!

tests/netlify-deploy.ts
pieh: I think we are just not getting any output at all on failures with `--json` flag. At least that's was my experience here and why I wanted to change this stuff in the first place to see actual logs on deploy failures

tools/deno/junit2json.ts
pieh: https://github.com/vercel/next.js/blob/a8de8730d75625e3bd068787121561cf6ab5eaac/packages/next/src/server/config.ts#L244 setting this env var might prevent build failure and allow actual usage of those features on stable versions (so the deploys wouldn't be failing but features themselves very well might not work correctly yet for us), but I wasn't sure what can of worms I could open when playing with this env var so went with "skipping"

Overall - as we are producing report for users to see and we are generating it for stable version - I think it's fair to skip or ignore tests for features that can't be used on stable versions (without getting into "test mode") - the change here might be not enough overall - but I think direction is correct - skip/ignore them instead of trying to run them?

Those tests would be displayed when we run tests against canary because we wouldn't be hitting CanaryOnly error then

tests/netlify-deploy.ts
pieh: https://github.com/opennextjs/opennextjs-netlify/pull/2698/commits/7faa8763172a6aad97225872879e6b4d9112f4fb

tools/deno/junit2json.ts
pieh: In any case - this is something we can change, but we need something here and I think this is good enough at least for now until we figure out exactly how we should handle this?

tests/netlify-deploy.ts
serhalp: yeah, I think that's right... maybe it's by design?

serhalp (approved):

PR #2699, opened 2024-10-25
ci: add deploy cleanup step for Vercel tests

pieh:
## Tests

1. TODO


.github/workflows/e2e-report.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         id: get-run-id
         run: |
           if [ "${{ inputs.use-branch }}" == "true" ]; then
-            E2E_RUN_ID=$(gh run list -w test-e2e.yml -e workflow_dispatch -s success -b $GITHUB_REF_NAME --json databaseId --jq ".[0].databaseId" --repo $GITHUB_REPOSITORY)
+            E2E_RUN_ID=$(gh run list -w test-e2e.yml -s success -b $GITHUB_REF_NAME --json databaseId --jq ".[0].databaseId" --repo $GITHUB_REPOSITORY)
```
pieh, line 34: not really related to deleting deploys, but this allow to generate test results report from runs on pull request (when label is applied) and not just on `workflow_dispatch`

.github/workflows/test-e2e.yml
```diff
@@ -28,7 +28,7 @@ env:
   DATADOG_TRACE_NEXTJS_TEST: true
   DATADOG_API_KEY: foo
   TEST_CONCURRENCY: 2
-  NEXT_E2E_TEST_TIMEOUT: 300000
+  NEXT_E2E_TEST_TIMEOUT: 600000
```
pieh, line 31: Similarly as above - not related to deleting deploys, but some of the test suites seemed to be on the verge of timeouts and would sometimes go over so this _might_ reduce some of the flakiness (not very confident in that part due to troubles with reproducing problems locally or even when running individual tests in github actions and not locally, but I think it's worth a try)

tests/netlify-deploy.ts
```diff
@@ -169,6 +174,32 @@ export class NextDeployInstance extends NextInstance {
...
+        `{ "deploy_id": "${this._deployId}" }`,
+      ])
+
+      if (deleteResponse.exitCode !== 0) {
+        require('console').error(
+          `Failed to delete deploy ${deleteResponse.stdout} ${deleteResponse.stderr} (${deleteResponse.exitCode})`,
+        )
+      } else {
+        require('console').log(`Successfully deleted deploy with deploy_id ${this._deployId}`)
+        this._shouldDeleteDeploy = false
```
mrstork, line 198: learning - why do we need to set this to false in the end? will it keep attempting and looping if we don't?

mrstork (approved):

tests/netlify-deploy.ts
pieh: It should not, this is just in case - I was looking at base class we extend - it does have some checks that look if something was deleted / destroyed already - there they throw:

https://github.com/vercel/next.js/blob/2e28c965279de90ce4bfca673196c27dd6117027/test/lib/next-modes/base.ts#L425-L428

but I didn't want to play with this as I don't fully know the contract that Vercel tests have with those modules - so this just felt safe way to just avoid wasting time on calling API to delete deploy that was already deleted in case this is called multiple times (`deleteDeploy` api call is not immediate - it takes couple seconds to finish so just no point in calling it again)

PR #2707, opened 2024-11-22
fix: use version of htmlrewriter which does not make use of asyncify,  which looks to have a potential memory leak under high load

JakeChampion:
## Description

We noticed the memory issue with Netlify's CSP plugin which used the same htmlrewriter library. We've built a new htmlrewriter library which uses the latest version of lol-html and removes the ability to use async-handlers, which is what required asyncify to be included.

### Documentation

This is intended to be a pure refactor and have no user visible changes, as such no documentation updates should be required

## Tests

1. `deno test -A edge-runtime/`

<https://deno.land/x/htmlrewriter@v1.0.0>

image


edge-runtime/lib/response.ts
```diff
@@ -1,5 +1,5 @@
 import type { Context } from '@netlify/edge-functions'
-import { HTMLRewriter } from '../vendor/deno.land/x/html_rewriter@v0.1.0-pre.17/index.ts'
+import { HTMLRewriter, init, type TextChunk } from 'https://deno.land/x/htmlrewriter@v1.0.0/src/index.ts'
```
serhalp, line 2: @JakeChampion I'm not super familiar with this setup, but it looks like all our deno imports use a relative path and should be included [in this file](https://github.com/opennextjs/opennextjs-netlify/blob/767b812969b8c099a2f60d87e5d0866594ecea18/edge-runtime/vendor.ts#L15) 🤔

pieh: Should be covered by https://github.com/opennextjs/opennextjs-netlify/pull/2707/commits/4a359c69c79f388cb936d18f4e9d572b588e2061

PR #2709, opened 2024-11-25
Use pull_request_target for triggering the lint work

JakeChampion:
This should allow the workflow to correctly run on forks

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,7 +1,7 @@
 name: Lint
 
 on:
-  pull_request:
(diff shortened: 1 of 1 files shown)


serhalp:
Doesn't seem like this would work, if I'm reading this right? 🤔 https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target

jessehouwing:
Don't use this, the pull_request_target is a very hard to use trigger.

PR #2717, opened 2024-12-06
feat: support after()

pieh:
## Description

This provides `waitUntil` implementation that Next.js will use for `after()` feature. If `context.waitUntil` is defined (it's conditionally available now depending on rollout) we will use just that. If it is not defined - we will use our TransformStream trick with keeping response open while awaiting `after()` (and other background work) to finish.

## Tests

Test added as second commit - fails without making any changes and work with code changes.
The test use similar setup as Next.js e2e `after` test - https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/next-after-app-deploy/index.test.ts and uses 2 pages:
- `/after/check` - cacheable page that renders timestamp
- `/after/trigger` which is SSR that uses `after()` to invalidate `/after/check` page

First we hit the `/after/check` and will get immediately stale content and regenerate it in background getting fresh timestamp
Then we hit that page twice again making sure we get stable timestamp in both hits
Then we hit `/after/trigger` which should invalidate `/after/check` if `after()` is allowed to fully execute
Then we hit `/after/check` again and compare timestamp to timestamps we saw before invalidating it - timestamp now should be greater than before

Closes https://github.com/opennextjs/opennextjs-netlify/issues/2695


src/run/handlers/server.ts
```diff
@@ -128,19 +124,19 @@ export default async (request: Request, context: FutureContext) => {
...
-        if (!context.waitUntil) {
-          // if waitUntil is not available, we have to keep response stream open until background promises are resolved
-          // to ensure that all background work executes
-          await requestContext.backgroundWorkPromise
-        }
+
+        // Next.js relies on `close` event emitted by response to trigger running callback variant of `next/after`
+        // however @fastly/http-compute-js never actually emits that event - so we have to emit it ourselves,
+        // otherwise Next would never run the callback variant of `next/after`
+        res.emit('close')
```
pieh, line 135: This is to support callback variant of `after()`:
```
after(async () => {
  // this will not start executing immediately - instead this function will be stored to and start executing after response was ~sent
  // make sense to not compete for any resources while working on response and delay executing this function to after that
})
```

This is not needed to support promise variant like so:
```
after(new Promise(resolve => {
  // this starts executing immediately and will ensure it will finish as long 
  // as function timeout is not reached
})
```

tests/smoke/deploy.test.ts
```diff
@@ -71,9 +71,13 @@ describe('version check', () => {
       )
     },
   )
-  test('yarn monorepo multiple next versions site is compatible', { retry: 0 }, async () => {
-    await smokeTest(selfCleaningFixtureFactories.yarnMonorepoMultipleNextVersionsSiteCompatible)
-  })
+  test(
+    'yarn monorepo multiple next versions site is compatible',
+    { retry: 0, timeout: 1_000 * 60 * 5 },
```
pieh, line 76: this change just bumps/sets timeout - we have 3 extensions being installed which we don't even use because site is in Netlify-testing account and that's pretty slow and this test case in particular was already slow before, but those 2 things combined meant that it was always timing out now

tests/fixtures/simple/next.config.js
```diff
@@ -8,6 +8,7 @@ const nextConfig = {
     outputFileTracingIncludes: {
       '/': ['public/**'],
     },
+    after: true,
```
pieh, line 11: TODO: this test need to be moved to separate / new fixture because `experimental.after` and `(unstable_)after` import only works for next@>=15 and this test is used also with older versions

The test setup and test itself won't really change other than that and can be already reviewed

src/run/handlers/request-context.cts
```diff
@@ -1,11 +1,18 @@
 import { AsyncLocalStorage } from 'node:async_hooks'
 
+import type { Context } from '@netlify/functions'
 import { LogLevel, systemLogger } from '@netlify/functions/internal'
 
 import type { NetlifyCachedRouteValue } from '../../shared/cache-types.cjs'
 
 type SystemLogger = typeof systemLogger
 
+// TODO: remove once https://github.com/netlify/serverless-functions-api/pull/219
```
mrstork, line 10: https://github.com/netlify/serverless-functions-api/pull/270

pieh: While the PR link is outdated (I just moved this code from other file and didn't write this part now) - the types for `Context` still omit `waitUntil` so we still need to have `FutureContext` ( https://github.com/netlify/serverless-functions-api/blob/a6e469c32d472d55f5f2cce0e0683d147354df70/src/context.ts#L14 is what `Context` export from `@netlify/functions` is)

So I think we should have just:
```
// TODO: remove once public types are updated
```

mrstork: If we can find a link to include to the list of public types, that would also help

src/run/handlers/server.ts
```diff
@@ -128,19 +124,19 @@ export default async (request: Request, context: FutureContext) => {
...
+
+        // if waitUntil is not available, we have to keep response stream open until background promises are resolved
+        // to ensure that all background work executes
+        await requestContext.backgroundWorkPromise
```
mrstork, line 139: This used to be wrapped in `!context.waitUntil`, is that still needed?
If not, could we update the comment? `if waitUntil is not available...`

tests/fixtures/simple/next.config.js
mrstork: - [ ] move to separate test that only uses `next >= 15`

src/run/handlers/server.ts
pieh: I'll update the comment.

When `context.waitUntil` is available we don't track background work and instead let `context.waitUntil` handle that. So this ends up being no-op (`await Promise.allSettled([])`) when `context.waitUntil` is available

tests/fixtures/simple/next.config.js
pieh: Done in https://github.com/opennextjs/opennextjs-netlify/pull/2717/commits/819be98917f25a3d7a6c9d726dbc38d438dbc2bb

src/run/handlers/request-context.cts
pieh: Updated comment in https://github.com/opennextjs/opennextjs-netlify/pull/2717/commits/80de44f1ff0e32f0c79c5321d040e614a5e84107

src/run/handlers/server.ts
pieh: Some clarification in https://github.com/opennextjs/opennextjs-netlify/pull/2717/commits/c938bb6aefd1f4f94c6c0d0f748e0b1bddda0239

orinokai (approved):
Looks good to me. Feels nicely abstracted in the `createRequestContext` function 👍🏼

mrstork (approved):

PR #2721, opened 2024-12-09
fix: use version of htmlrewriter which does not make use of asyncify, which looks to have a potential memory leak under high load

pieh:
## Description

This was based on https://github.com/opennextjs/opennextjs-netlify/pull/2707 (new PR, primarily due to problems with running tests against forks).

Need to use bit of hacks and indirections due to:
 - `deno vendor` not playing well with wasm files ( https://github.com/denoland/deno/issues/14123), so I added manual pull for this file after `deno vendor` in our build script
 - just vendoring even with above result in `fetch` attempt against `file://` url which fails:
 ```
TypeError: NetworkError when attempting to fetch resource.
    at async mainFetch (ext:deno_fetch/26_fetch.js:277:12)
    at async fetch (ext:deno_fetch/26_fetch.js:504:7)
    at async __wbg_init (file:///var/folders/qh/jgsvdzxn6y9bfw4w27bd4wtw0000gn/T/opennextjs-netlify-uvRxF2/.netlify/edge-functions-bundled/source/root/.netlify/edge-functions/___netlify-edge-handler-middleware/edge-runtime/vendor/deno.land/x/htmlrewriter@v1.0.0/pkg/htmlrewriter.js:1212:49)
    at async file:///var/folders/qh/jgsvdzxn6y9bfw4w27bd4wtw0000gn/T/opennextjs-netlify-uvRxF2/.netlify/edge-functions-bundled/source/root/.netlify/edge-functions/___netlify-edge-handler-middleware/edge-runtime/lib/response.ts:5:1
```
 - to workaround above we pass wasm module to `init` function, but Netlify bundling also seems to skip `.wasm` files so I inline wasm module using same method we currently inline wasm modules used by user in middleware

## Tests

Any test that make use of middleware already tests this implicitly

Fixes https://linear.app/netlify/issue/FRB-1523/nextjs-runtime-uses-the-same-html-rewriter-as-the-csp-plugin-that

### DIFF
--- a/tools/build.js
+++ b/tools/build.js
@@ -109,6 +109,7 @@ async function vendorDeno() {
       const destination = join(vendorDest, url.hostname, url.pathname)
 
       const res = await fetch(url)
+      if (!res.ok) throw new Error('Failed to fetch .wasm file to vendor', { cause: err })
       const fileStream = createWriteStream(destination, { flags: 'wx' })
       await finished(Readable.fromWeb(res.body).pipe(fileStream))
     }),

### CODE REVIEW

File: src/build/functions/edge.ts [hunk 1]
```diff
@@ -85,13 +85,27 @@ const writeHandlerFile = async (ctx: PluginContext, { matchers, name }: NextDefi
...
     `
+    import { decode as _base64Decode } from './edge-runtime/vendor/deno.land/std@0.175.0/encoding/base64.ts';
+    import { init as htmlRewriterInit } from './edge-runtime/vendor/deno.land/x/htmlrewriter@v1.0.0/src/index.ts'
     import {handleMiddleware} from './edge-runtime/middleware.ts';
     import handler from './server/${name}.js';
+
+    await htmlRewriterInit({ module_or_path: _base64Decode(${JSON.stringify(
+      htmlRewriterWasm.toString('base64'),
+    )}).buffer });
+
```
pieh, line 108: bundling wasm file as-is doesn't seem to work, so instead we inline it using same method we do for wasm modules used by users in middleware https://github.com/opennextjs/opennextjs-netlify/blob/6b56128b90ce9d7d8bc7a6a125fb755f6fe160ac/src/build/functions/edge.ts#L125-L133 

and pass module to `init` function of htmlrewriter (argumentless init doesn't work after vendoring)

serhalp (dismissed):

tools/build.js
```diff
@@ -94,6 +97,22 @@ async function vendorDeno() {
   console.log(`📦 Vendoring Deno modules into '${vendorDest}'...`)
 
   await execaCommand(`deno vendor ${vendorSource} --output=${vendorDest} --force`)
+
+  // htmlrewriter contains wasm files and those don't currently work great with vendoring
+  // see https://github.com/denoland/deno/issues/14123
+  // to workaround this we copy the wasm files manually
```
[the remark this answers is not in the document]
pieh: Deno landed it, but in 2.1 - https://deno.com/blog/v2.1#first-class-wasm-support , so support is not guaranteed :( otherwise we could migrate to imports in https://github.com/netlify/htmlrewriter/blob/edb477af3d08359c8f25edd8d301df69ffcc8e4b/pkg/htmlrewriter.js#L1184 to make use of it

src/build/functions/edge.ts
pieh: https://gist.github.com/pieh/b205a685e518ef62afaab392e61e064f this is standalone script using this method that can be tested.

You can also test that passing crap to `init` first arg like:
```
await htmlRewriterInit({
  module_or_path: 'wat',
})
```

would result in errors:

```
error: Uncaught (in promise) TypeError: Invalid URL: 'wat'
    module_or_path = fetch(module_or_path);
                     ^
    at getSerialization (ext:deno_url/00_url.js:98:11)
    at new URL (ext:deno_url/00_url.js:405:27)
    at new Request (ext:deno_fetch/23_request.js:329:25)
    at ext:deno_fetch/26_fetch.js:320:27
    at new Promise (<anonymous>)
    at fetch (ext:deno_fetch/26_fetch.js:316:18)
    at __wbg_init (https://deno.land/x/htmlrewriter@v1.0.0/pkg/htmlrewriter.js:1207:22)
    at file:///Users/misiek/dev/pgs-next-runtime/edge-runtime/test.ts:11:7
```

so the argument has effect. 

Commenting out `init` call also result in errors:

```
error: Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'htmlrewriter_new')
    const ret = wasm.htmlrewriter_new(
                     ^
    at new HTMLRewriter (https://deno.land/x/htmlrewriter@v1.0.0/pkg/htmlrewriter.js:781:22)
    at Object.start (https://deno.land/x/htmlrewriter@v1.0.0/src/index.ts:53:20)
    at Module.invokeCallbackFunction (ext:deno_webidl/00_webidl.js:981:16)
    at new TransformStream (ext:deno_web/06_streams.js:6214:16)
    at HTMLRewriter.transform (https://deno.land/x/htmlrewriter@v1.0.0/src/index.ts:48:29)
    at rewriter (file:///Users/misiek/dev/pgs-next-runtime/edge-runtime/test.ts:39:6)
    at file:///Users/misiek/dev/pgs-next-runtime/edge-runtime/test.ts:43:27
```

youvalv (approved):

src/build/functions/edge.ts
```diff
@@ -85,13 +85,27 @@ const writeHandlerFile = async (ctx: PluginContext, { matchers, name }: NextDefi
...
     join(handlerDirectory, `${handlerName}.js`),
     `
+    import { decode as _base64Decode } from './edge-runtime/vendor/deno.land/std@0.175.0/encoding/base64.ts';
+    import { init as htmlRewriterInit } from './edge-runtime/vendor/deno.land/x/htmlrewriter@v1.0.0/src/index.ts'
     import {handleMiddleware} from './edge-runtime/middleware.ts';
     import handler from './server/${name}.js';
+
+    await htmlRewriterInit({ module_or_path: _base64Decode(${JSON.stringify(
+      htmlRewriterWasm.toString('base64'),
+    )}).buffer });
```
JakeChampion, line 107: we could use a uint8array instead of a base64 string here, it would save on memory usage by 33%

I did this in a similar package I made specifically for CSP support - I could do the same for the htmlrewriter package if we want:
```js
await Deno.writeTextFile(
    "./pkg/embedded-wasm.ts",
    `export const wasmBinary = Uint8Array.from(${
      JSON.stringify(
        Array.from(
          await Deno.readFile(
            `./pkg/${wasmFile}`,
          ),
        ),
      )
    });`,
  );
 ```

https://github.com/netlify/csp_nonce_html_transformer/blob/main/scripts/build.ts#L13C3-L25

PR #2722, opened 2024-12-11
test: update test setup for stable after()

pieh:
## Description

Our tests failed last run with underlying reason being:
```
Dec 11, 09:38:45 AM: 45433016 ERROR   ⨯ TypeError: (0 , o.unstable_after) is not a function
    at a (.next/server/app/after/trigger/page.js:1:2753) {
  digest: '1467464559'
}
```

`unstable_after` import is no longer available, so this updates test setup to use stable `after` (and bumps version constraint)

## Tests

1. TODO


tests/e2e/after.test.ts
```diff
@@ -25,7 +28,9 @@ test('next/after callback is executed and finishes', async ({ page, after }) =>
 
   expect(pageInfo2.timestamp, 'Check page should be cached').toBe(pageInfo1.timestamp)
 
-  await page.goto(`${after.url}/after/trigger`)
+  const response = await page.goto(`${after.url}/after/trigger`)
+
+  expect(response?.status(), 'Trigger should return 200').toBe(200)
```
pieh, line 33: In `main` this results in 50x status (due to no longer existing export) - the test failed which is good, but assertion errors didn't show much details other than things not working

This added assertion is just to have more context about potential future failures

tests/utils/next-version-helpers.mjs
```diff
@@ -85,7 +85,7 @@ export async function setNextVersionInFixture(
         if (
           operation === 'update' &&
           versionConstraint &&
-          !satisfies(checkVersion, versionConstraint) &&
+          !satisfies(checkVersion, versionConstraint, { includePrerelease: true }) &&
```
pieh, line 88: not needed for this PR, but I noticed that `after` fixture building was skipped in canary test run which should not happen (we don't have integration test using that fixture, so it wasn't impacting test results - this is just fixing this for the future)

mrstork (approved):

PR #2723, opened 2024-12-11
ci: parallelise e2e via workers

orinokai:
## Description

Testing alternative parallelisation


serhalp:
@orinokai make sure to change from `feat:` to `chore:` or `ci:` if we merge this. We don't need a release or release notes for this one.

PR #2727, opened 2024-12-18
fix: make `revalidateTags` no-op when list of tags is empty

eduardoboucas:
## Description

Makes the `revalidateTags` call a no-op when the array of tags is empty.

## Tests

Added integration test with scenario resulting in Next.js calling `CacheHandler.revalidateTag` with empty array of tags and ensuring purge API endpoint is not being used

https://linear.app/netlify/issue/FRB-1536/dont-call-purgecache-when-tags-is-empty-array


tests/integration/simple-app.test.ts
```diff
@@ -22,6 +24,8 @@ import {
   startMockBlobStore,
 } from '../utils/helpers.js'
 import { nextVersionSatisfies } from '../utils/next-version-helpers.mjs'
+import { purgeCache } from '@netlify/functions'
+import { afterEach } from 'node:test'
```
serhalp, line 28: 🤔 shouldn't this be from `vitest` like all the others?

tests/integration/simple-app.test.ts
```diff
@@ -210,6 +243,39 @@ test<FixtureTestContext>('cacheable route handler is cached on cdn (revalidate=f
...
+  // set the NETLIFY_PURGE_API_TOKEN to get pass token check and allow fetch call to be made
+  vi.stubEnv('NETLIFY_PURGE_API_TOKEN', 'mock')
+
+  const page1 = await invokeFunction(ctx, {
+    url: '/unstable_cache',
+  })
+  const data1 = load(page1.body)('pre').text()
+
+  // allow for cache entry to get stale
+  await new Promise((res) => setTimeout(res, 2000))
```
serhalp, line 259: is mocking and advancing the clock here problematic?

serhalp (approved):
LGTM

tests/integration/simple-app.test.ts
pieh: oooh, yes

tests/integration/simple-app.test.ts
pieh: we use above in plenty of other places that check behaviors related to stale cache items so I just the same method here

I'm not sure if attempting to mock time would work

tests/integration/simple-app.test.ts
pieh: https://github.com/opennextjs/opennextjs-netlify/pull/2727/commits/4a02e345b378a22fec355ced93a1e9ffa9886a87

pieh (approved):

PR #2734, opened 2024-12-27
fix: adjust middleware json data rewrite to work with recent next@canary

pieh:
## Description

Adjusts middleware handling to use json data urls and not route paths with `__nextDataReq` query param (that no longer works with recent next@canary after https://github.com/vercel/next.js/pull/74100) for middleware json data rewrites

--- 

Initial previous attempt description below (this felt the quickest "hot fix" but it seemed really wrong to do so):

~This is attempting to address changes made in https://github.com/vercel/next.js/pull/74100 by keeping previous query param "working".~

~This feels like wrong approach to me, but I'm not quite sure how this should be fixed properly. In our middleware handling we are relying on this search param:~ https://github.com/opennextjs/opennextjs-netlify/blob/c3e328c67aee611f83567ebf95c86abcfdd52d93/edge-runtime/lib/response.ts#L182-L186 

~Maybe instead we should transform pathname into json data path instead of keeping it as cannonical route path with added query param, but comment in existing code gives me a pause - I'm not sure if it's about our own handling or it's about Next handling in general.~

~This feels like quickest (and quite hacky ...) way to keep middleware rewrites for data routes working, so maybe this could be considered stop-gap solution that will do for now, until more investigation can happen.~

## Tests

Added e2e test (we only had integration tests really and those were not checking all parts of handling request - one test was checking origin - literally if using `__nextDataReq` search param results in data response (I did remove this one as it would no longer work with `next@canary` and because we are not relying on this behavior anymore) and the other ones were testing just middleware part - I did update those)

https://linear.app/netlify/issue/FRB-1545/investigate-integration-test-failures-due-to-recent-pr-changes

### DIFF
--- a/edge-runtime/lib/response.ts
+++ b/edge-runtime/lib/response.ts
@@ -14,6 +14,8 @@ import {
   normalizeLocalePath,
   normalizeTrailingSlash,
   relativizeURL,
+  removeBasePath,
+  rewriteDataPath,
 } from './util.ts'
 
 export interface FetchEventResult {
@@ -180,13 +182,15 @@ export const buildResponse = async ({
     }
 
     if (isDataReq) {
-      // The rewrite target is a data request, but a middleware rewrite target is always for the page route,
-      // so we need to tell the server this is a data request. Setting the `x-nextjs-data` header is not enough. 🤷
-      rewriteUrl.searchParams.set('__nextDataReq', '1')
+      rewriteUrl.pathname = rewriteDataPath({
+        dataUrl: new URL(request.url).pathname,
+        newRoute: removeBasePath(rewriteUrl.pathname, nextConfig?.basePath),
+        basePath: nextConfig?.basePath,
+      })
+    } else {
+      // respect trailing slash rules to prevent 308s
+      rewriteUrl.pathname = normalizeTrailingSlash(rewriteUrl.pathname, nextConfig?.trailingSlash)
     }
-
-    // respect trailing slash rules to prevent 308s
-    rewriteUrl.pathname = normalizeTrailingSlash(rewriteUrl.pathname, nextConfig?.trailingSlash)
 
     const target = normalizeLocalizedTarget({ target: rewriteUrl.toString(), request, nextConfig })
     if (target === request.url) {
--- a/tests/e2e/edge-middleware.test.ts
+++ b/tests/e2e/edge-middleware.test.ts
@@ -54,14 +54,11 @@ test('it should render OpenGraph image meta tag correctly', async ({ page, middl
 })
 
 test('json data rewrite works', async ({ middlewarePages }) => {
-  const response = await fetch(
-    `${middlewarePages.url}/_next/data/build-id/sha.json?__nextDataReq=1`,
-    {
-      headers: {
-        'x-nextjs-data': '1',
-      },
+  const response = await fetch(`${middlewarePages.url}/_next/data/build-id/sha.json`, {
+    headers: {
+      'x-nextjs-data': '1',
     },
-  )
+  })
 
   expect(response.ok).toBe(true)
   const body = await response.text()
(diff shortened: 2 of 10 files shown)


edge-runtime/lib/response.ts
```diff
@@ -14,6 +14,8 @@ import {
   normalizeLocalePath,
   normalizeTrailingSlash,
   relativizeURL,
+  removeBasePath,
+  rewriteDataPath,
```
pieh, line 18: Fun fact - we had this function already but we never used it

edge-runtime/lib/response.ts
```diff
@@ -180,14 +182,16 @@ export const buildResponse = async ({
     }
 
     if (isDataReq) {
```
pieh, line 184: As far as I can tell - this will only be truthy for pages router json requests
https://github.com/opennextjs/opennextjs-netlify/blob/a1d859dfa6868a668a6b84e117467a18c748439b/edge-runtime/lib/response.ts#L126-L127

App router RSC requests don't seem to have that header set so this _should_ be safe to do

tests/e2e/edge-middleware.test.ts
```diff
@@ -52,3 +52,20 @@ test('it should render OpenGraph image meta tag correctly', async ({ page, middl
...
+
+  expect(response.ok).toBe(true)
+  const body = await response.text()
+
+  expect(body).toMatch(/^{"pageProps":/)
+
+  const data = JSON.parse(body)
+
+  expect(data.pageProps.message).toBeDefined()
+})
```
pieh, line 71: Using already existing setup in one of our fixtures https://github.com/opennextjs/opennextjs-netlify/blob/c3e328c67aee611f83567ebf95c86abcfdd52d93/tests/fixtures/middleware-pages/middleware.js#L37-L40 for middleware rewrite

and https://github.com/opennextjs/opennextjs-netlify/blob/c3e328c67aee611f83567ebf95c86abcfdd52d93/tests/fixtures/middleware-pages/pages/shallow.js#L37-L43 for asserting body

edge-runtime/lib/response.ts
mrstork: It looks like next uses `getRequestMeta(params.req, 'isNextDataReq')` for this? There's a chance we could do the same.
https://github.com/vercel/next.js/blob/canary/packages/next/src/server/request-meta.ts#L183-L205

mrstork (approved):

edge-runtime/lib/response.ts
pieh: This was my initial attempt - https://github.com/opennextjs/opennextjs-netlify/pull/2734/commits/f5bb340270f656069efb84fe2abc3b7a91f37844 but to make it work, something in middleware handling (edge function) would need to signal to origin (lambda) that this meta data should be set - in that commit I just made the previous query param work again (if it was present I was setting that meta data), but this felt wrong given that Next.js explicitly removed support for that. I'm not exactly sure how it works on Vercel - I'm clearly missing something here, but I don't think attempting to set those meta tags directly is the right way to go - maybe figuring out how that meta data is meant to be set would be the best solution here - I'll try to look if I can figure that out

pieh: Ok, so the part of code that is setting the meta data on requests (~https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/router-utils/resolve-routes.ts) ... but it only happens when it is `_next/data` request to begin with, so because we were using canonical path before we never hit this code path and we just relied on fact that we could "externally" force it with query param.

So current fix (at least in terms of the approach of using data url instead of trying to add the internal meta data on request ourselves) seems like generally correct approach to me

mrstork: Thanks for looking into it even more deeply!

PR #2738, opened 2025-01-14
chore: use uint8array for wasm module instead of base64 decoding

JakeChampion:
This reduces the amount of memory being used by the function, base64 will use four thirds of the original size, whereas using uint8array will use the same size as the original size


JakeChampion:
https://github.com/opennextjs/opennextjs-netlify/pull/2739

PR #2739, opened 2025-01-14
fix: use uint8array for htmlrewriter wasm module instead of base64

JakeChampion:
This reduces the amount of memory being used by the function, base64 will use four thirds of the original size, whereas using uint8array will use the same size as the original size

### DIFF
--- a/src/build/functions/edge.ts
+++ b/src/build/functions/edge.ts
@@ -97,14 +97,13 @@ const writeHandlerFile = async (ctx: PluginContext, { matchers, name }: NextDefi
   await writeFile(
     join(handlerDirectory, `${handlerName}.js`),
     `
-    import { decode as _base64Decode } from './edge-runtime/vendor/deno.land/std@0.175.0/encoding/base64.ts';
     import { init as htmlRewriterInit } from './edge-runtime/vendor/deno.land/x/htmlrewriter@v1.0.0/src/index.ts'
-    import {handleMiddleware} from './edge-runtime/middleware.ts';
+    import { handleMiddleware } from './edge-runtime/middleware.ts';
     import handler from './server/${name}.js';
 
-    await htmlRewriterInit({ module_or_path: _base64Decode(${JSON.stringify(
-      htmlRewriterWasm.toString('base64'),
-    )}).buffer });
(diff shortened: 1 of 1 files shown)


pieh:
Not exactly sure why, but https://github.com/opennextjs/opennextjs-netlify/actions/runs/12771972335/job/35600421283#step:13:183 fails with
```
CompileError: WebAssembly.instantiate(): BufferSource argument is empty

       at ../../../../../../../tmp/opennextjs-netlify-sfuoSQ/apps/page-router/.netlify/edge-functions/___netlify-edge-handler-src-middleware/edge-runtime/vendor/deno.land/x/htmlrewriter@v1.0.0/pkg/htmlrewriter.js:997

       995 |     return await WebAssembly.instantiate(bytes, imports);
       996 |   } else {
    >  997 |     const instance = await WebAssembly.instantiate(module, imports);
           |                                        ^
       998 |
       999 |     if (instance instanceof WebAssembly.Instance) {
      1000 |       return { instance, module };

        at __wbg_load (file:///tmp/opennextjs-netlify-sfuoSQ/apps/page-router/.netlify/edge-functions/___netlify-edge-handler-src-middleware/edge-runtime/vendor/deno.land/x/htmlrewriter@v1.0.0/pkg/htmlrewriter.js:997:40)
        at __wbg_init (file:///tmp/opennextjs-netlify-sfuoSQ/apps/page-router/.netlify/edge-functions/___netlify-edge-handler-src-middleware/edge-runtime/vendor/deno.land/x/htmlrewriter@v1.0.0/pkg/htmlrewriter.js:1212:38)
        at async file:///tmp/opennextjs-netlify-sfuoSQ/apps/page-router/.netlify/edge-functions/___netlify-edge-handler-src-middleware/___netlify-edge-handler-src-middleware.js:6:5
```

It also similarly seems to fail in integration tests
 - https://github.com/opennextjs/opennextjs-netlify/actions/runs/12771972335/job/35600418860
 - https://github.com/opennextjs/opennextjs-netlify/actions/runs/12771972335/job/35600423495

Does this error ring any bells?

JakeChampion:
> @JakeChampion keep in mind `chore:` won't result in a release

changed to fix 👍

pieh (approved):

PR #2743, opened 2025-01-20
feat: add logging to understand impact of trailing slash 308s

orinokai:
## Description

We are considering caching 308 trailing slash redirects to reduce function invocations.

This PR will allow us to see the impact of this change.

### DIFF
--- a/src/run/headers.ts
+++ b/src/run/headers.ts
@@ -3,7 +3,7 @@ import type { NextConfigComplete } from 'next/dist/server/config-shared.js'
 
 import { encodeBlobKey } from '../shared/blobkey.js'
 
-import type { RequestContext } from './handlers/request-context.cjs'
+import { getLogger, RequestContext } from './handlers/request-context.cjs'
 import type { RuntimeTracer } from './handlers/tracer.cjs'
 import { getRegionalBlobStore } from './regional-blob-store.cjs'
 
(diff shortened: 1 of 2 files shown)


src/run/headers.ts
```diff
@@ -234,6 +235,13 @@ export const setCacheControlHeaders = (
     return
   }
 
+  // temporary diagnostic to evaluate number of trailing slash redirects
+  if (status === 308 && request.url.endsWith('/') !== nextConfig.trailingSlash) {
```
pieh, line 239: Should we perhaps also check if `Location` header and `request.url` pathnames are almost the same and only diff is that one of those would have trailing slash to limit potential false positives? 

Conceptually there could other redirects that would fall into this condition that are not related to trailing slash redirects (tho not sure about exact details)

orinokai: Thanks, I thought about that, but then I couldn't think of any other redirects that would fall into this condition. It's impossible to configure user redirects that don't match the trailing slash settings and I couldn't think of any others. Did you have any specific in mind?

pieh: Maybe redirects to external domains/targets could have this characteristic possibly next skipping trailing slash handling if target is external, but as I mentioned - I don't have any concrete scenario in mind, so we can skip this if this is already something you thought about

PR #2749, opened 2025-01-27
test: fix test setup

pieh:
## Description

Our tests are currently failing - see last scheduled test run - https://github.com/opennextjs/opennextjs-netlify/actions/runs/13279231786

This PR contains couple of adjustments:
1. Upgrade corepack ( currently we are hitting https://github.com/nodejs/corepack/issues/612 because corepack version shipped with used Node.js version doesn't contain the fix - only Node@23 ships with corepack version that has fix for it)
2. Skip building fixtures that are not used in our integration tests in our `prepare` script - this is both making our CI runs a bit faster, but also it skips `middleware-og` fixture in paricular that currently fails to build on Windows on Node 18.20.6 - this is not next runtime issue - this is pure Next.js issue that is result of some Node.js changes that were done as security fix that unfortunately also broke Next's OpenGraph handling on Windows on that Node version. The same issue was kind of tackled in some capacity before in https://github.com/opennextjs/opennextjs-netlify/pull/2752#discussion_r1936041720 that has a bit of context on the problem - build time issue (for different fixture) was not visible then because cached fixtures were used (that were created on previous Node.js patch version before the change that broke OpenGraph handling in Next)
3. If prepare script fails to build a fixture - it now exits with non-zero code. We were not checking if our prepare script actually build test fixtures and as a result we not only attempt to run tests on broken fixtures - see https://github.com/opennextjs/opennextjs-netlify/actions/runs/12944298814/job/36105004178#step:11:291 failure that later result in test failures in same run, but also possibly caching that (if one of test shards don't fail any tests - which can happen for a shard that ... runs only unit tests - like this one https://github.com/opennextjs/opennextjs-netlify/actions/runs/12944298814/job/36105016872 )
   
   There is still some root problem to solve on why npm installation fails on windows while it doesn't on linux (maybe some npm cache related problem?), so this is just attempting to short circuit CI when any of fixtures fail to build - avoid running tests and more importantly - avoid caching them (for unit test shards case), as subsequent test runs will have confusing errors about not exising "publish dir" and test run not even building fixtures and just pulling from cache

## Tests

For point 3:

I did intentionally break 2 fixtures with second commit:
 - first one by adding not existing package to package.json to test just deps installation case
 - second one by making syntax error so Next.js fails `next build` (which we do run as `postinstall` script in our fixtures)

Results can be seen in https://github.com/opennextjs/opennextjs-netlify/actions/runs/12996052794 with all integration and unit tests shards failing, not even proceeding to running tests and just skipping fixture caching

I later reverted intentional breakage


.github/workflows/run-tests.yml
```diff
@@ -127,8 +129,17 @@ jobs:
...
+        if: runner.os == 'Windows'
+        # On Windows by default PATH prefers corepack bundled with Node.js
+        # This prepends npm global to PATH to ensure that npm installed global corepack is used instead
+        run: |
+          echo "$(npm config get prefix)" >> "$GITHUB_PATH"
+        shell: bash
       - name: setup pnpm/yarn
-        run: corepack enable
+        run: |
+          npm install -g corepack --force
```
mrstork, line 141: Learning - why `corepack@latest` above, and then `--force` in this step?

tests/prepare.mjs
```diff
@@ -18,12 +18,31 @@ const NEXT_VERSION = process.env.NEXT_VERSION ?? 'latest'
...
+  'after',
+  'cli-before-regional-blobs-support',
+  'dist-dir',
+  // There is also a bug on Windows on Node.js 18.20.6, that cause build failures on this fixture
+  // see https://github.com/opennextjs/opennextjs-netlify/actions/runs/13268839161/job/37043172448?pr=2749#step:12:78
+  'middleware-og',
+  'middleware-single-matcher',
+  'nx-integrated',
+  'turborepo',
+  // We do have a skipped test using this fixture (due to integration test setup not working for monorepos)
```
mrstork, line 32: Same here, perhaps a linear issue link?

.github/workflows/run-tests.yml
pieh: The `@latest` selector in the `e2e` is probably not needed - just result of iterating and trying to find working solution. I will remove it to keep things in sync.

The `--force` here is needed, because Windows one otherwise would fail to install - without it I was getting following error:
```
npm error code EEXIST
npm error path C:\npm\prefix\yarn
npm error EEXIST: file already exists
npm error File exists: C:\npm\prefix\yarn
npm error Remove the existing file and try again, or run npm
npm error with --force to overwrite files recklessly.
```

And we only needed it in jobs that run integration tests because only those use Windows runners. E2E ones are linux only hence don't need it.

tests/prepare.mjs
```diff
@@ -18,12 +18,31 @@ const NEXT_VERSION = process.env.NEXT_VERSION ?? 'latest'
 const fixturesDir = fileURLToPath(new URL(`./fixtures`, import.meta.url))
 const fixtureFilter = argv[2] ?? ''
 
+// E2E tests run next builds, so we don't need to prepare those ahead of time for integration tests
+const e2eOnlyFixtures = new Set([
+  'after',
+  'cli-before-regional-blobs-support',
+  'dist-dir',
+  // There is also a bug on Windows on Node.js 18.20.6, that cause build failures on this fixture
+  // see https://github.com/opennextjs/opennextjs-netlify/actions/runs/13268839161/job/37043172448?pr=2749#step:12:78
```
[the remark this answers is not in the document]
pieh: There is currently no Next.js issue to link to as of now and I didn't find time to create proper Next.js issue for it myself (and I would need to prep repro and make sure it works on my personal windows machine, so that's not something I really have time to do during work hours :S). Without that issue in Next.js there is not much of a point creating internal linear issue, because it wouldn't say anything more than this comment is saying now.

Being a bit sneaky I could just remove this whole comment as it's still true that we don't actually use this fixture in integration tests and let someone possibly encounter that problem without warning, so adding this comment was just trying to balance providing at least some information here while acknowledging that I don't have bandwidth to create proper tracking ticket in Next.js repository.

tests/prepare.mjs
pieh: Would removing the skipped test in this PR (the test was skipped a year ago - https://github.com/opennextjs/opennextjs-netlify/commit/219588ea7ae021eb1da267db5685d3cc51e37941#diff-924485d81a4d8552389a0b2697f69ee3508807e55c08077e48455ea7e44be5d2R39 work? 

As comment around reason for skipping mention - we actually do have e2e tests using this fixture proving the fixture is working https://github.com/opennextjs/opennextjs-netlify/blob/4bcb34fc7fce435c29376bb90895bc4e827b2ec6/tests/integration/turborepo.test.ts#L27

The test was skipped over a year ago https://github.com/opennextjs/opennextjs-netlify/commit/219588ea7ae021eb1da267db5685d3cc51e37941#diff-924485d81a4d8552389a0b2697f69ee3508807e55c08077e48455ea7e44be5d2R36-R39 so this doesn't seem like actual priority that we would work on anytime soon.

mrstork (approved):

PR #2753, opened 2025-01-29
fix(perf): exclude entire `/_next/static/` dir from generated function

serhalp:
## Description

See inline comment. This only matters because of browsers requesting assets from previous deploys or bots crawling for garbage. This will avoid unnecessary function invocations in those cases, improving performance and reducing customer bills in some cases.

### Documentation

N/A

## Tests

Added e2e coverage (requires the full Netlify platform to be exercised)

### DIFF
--- a/src/build/templates/handler-monorepo.tmpl.js
+++ b/src/build/templates/handler-monorepo.tmpl.js
@@ -53,4 +53,9 @@ export default async function (req, context) {
 export const config = {
   path: '/*',
   preferStatic: true,
+  excludedPath: [
+    // We use `preferStatic: true` so we already won't run this on *existing* static assets,
+    // but by excluding this entire path we also avoid invoking the function just to 404.
+    '/_next/static/*',
+  ],
 }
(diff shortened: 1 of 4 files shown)


src/build/templates/handler-monorepo.tmpl.js
```diff
@@ -53,4 +53,9 @@ export default async function (req, context) {
 export const config = {
   path: '/*',
   preferStatic: true,
+  excludedPath: [
+    // We use `preferStatic: true` so we already won't run this on *existing* static assets,
+    // but by excluding this entire path we also avoid invoking the function just to 404.
+    '/_next/static/*',
```
pieh, line 59: This might need to be prefixed with [`basePath`](https://nextjs.org/docs/app/api-reference/config/next-config-js/basePath)

Checking how fixture with this https://github.com/opennextjs/opennextjs-netlify/blob/f4b59b67e0b97af83d22750bc9828d2820278305/tests/fixtures/page-router-base-path-i18n/next.config.js#L8 behave with `next build && next start` shows requests being made for `http://localhost:3000/base/path/_next/static/chunks/webpack-039418b95632e39f.js` which currently wouldn't be excluded.

PR #2764, opened 2025-02-13
fix: apply caching headers to pages router 404 with getStaticProps

pieh:
## Tests

1. TODO


tests/integration/page-router.test.ts
```diff
@@ -126,6 +126,10 @@ test<FixtureTestContext>('Should serve correct locale-aware custom 404 pages', a
     load(responseImplicitDefaultLocale.body)('[data-testid="locale"]').text(),
     'Served 404 page content should use default locale if locale is not explicitly used in pathname (after basePath)',
   ).toBe('en')
+  expect(
+    responseImplicitDefaultLocale.headers['netlify-cdn-cache-control'],
+    'Response for not existing route if locale is not explicitly used in pathname (after basePath) should have 404 status',
```
pieh, line 131: ops this is not right msg - just copied some `expect()` line from before and forgot to adjust it (and same for other assertions here)

src/run/headers.ts
```diff
@@ -242,11 +250,25 @@ export const setCacheControlHeaders = (
...
+    if (
+      typeof requestContext.pageHandlerRevalidate !== 'undefined' &&
+      ['GET', 'HEAD'].includes(request.method) &&
+      !headers.has('cdn-cache-control') &&
+      !headers.has('netlify-cdn-cache-control')
+    ) {
+      // handle CDN Cache Control on 404 Page responses
+      setCacheControlFromRequestContext(headers, requestContext.pageHandlerRevalidate)
+      return
+    }
```
orinokai, line 271: Just making sure I understand correctly - this will only add cache control headers to 404 pages with a revalidate value. What about 404 pages that use getStaticProps but don't specify revalidate?

pieh: The method (I extracted from existing handling for App Router Route Handlers) supports `false | number` for `revalidate`:
https://github.com/opennextjs/opennextjs-netlify/blob/330dca0ceee048d05dbed4958e3c33e1008c3ae0/src/run/headers.ts#L220

`false` meaning that `revalidate` is not specified and then it would set ~permanent / 1 year caching header

pieh: The test adjustment I did (minus currently wrong "message") - https://github.com/opennextjs/opennextjs-netlify/blob/330dca0ceee048d05dbed4958e3c33e1008c3ae0/tests/integration/page-router.test.ts#L117-L132 is for the case of `revalidate` not being specified and test asserting 1 year caching headers - this is used test fixture for it https://github.com/opennextjs/opennextjs-netlify/blob/330dca0ceee048d05dbed4958e3c33e1008c3ae0/tests/fixtures/page-router-base-path-i18n/pages/404.js#L9-L16

We currently don't have test case for 404 with `getStaticProps` + `revalidate` being actually defined - I did some quick manual test and it _seemed_ to be working (but this was real quick test and with rushing it maybe it's not). We should add a test case for it

src/run/headers.ts
```diff
@@ -208,6 +209,19 @@ export const adjustDateHeader = async ({
...
 
+function setCacheControlFromRequestContext(
+  headers: Headers,
+  revalidate: NetlifyCachedRouteValue['revalidate'],
+) {
+  const cdnCacheControl =
+    // if we are serving already stale response, instruct edge to not attempt to cache that response
+    headers.get('x-nextjs-cache') === 'STALE'
+      ? 'public, max-age=0, must-revalidate, durable'
+      : `s-maxage=${revalidate || 31536000}, stale-while-revalidate=31536000, durable`
```
orinokai, line 220: I think it should be okay to use `${revalidate || 31536000}` because we need to account for `false` and `undefined` (see comment below), and even if `revalidate` was `0` then we should be using 1 year.

src/run/headers.ts
```diff
@@ -242,14 +250,27 @@ export const setCacheControlHeaders = (
...
+      headers.set('cache-control', 'public, max-age=0, must-revalidate')
+      headers.set('netlify-cdn-cache-control', `max-age=31536000, durable`)
+      return
+    }
+
+    if (
+      ['GET', 'HEAD'].includes(request.method) &&
+      !headers.has('cdn-cache-control') &&
+      !headers.has('netlify-cdn-cache-control')
+    ) {
```
orinokai, line 267: In addition to removing `!cacheControl || cacheControl.includes('no-store')` from the condition, I found I had to also remove `typeof requestContext.pageHandlerRevalidate !== 'undefined'` because 404s without `getStaticProps` do not go through the cache handler so `pageHandlerRevalidate` is `undefined` in those cases. Was it there as a precaution, or was there a specific case you were guarding against? @pieh

tests/integration/page-router.test.ts
```diff
@@ -153,3 +153,124 @@ test<FixtureTestContext>('Should serve correct locale-aware custom 404 pages', a
     'Served 404 page content should use non-default locale if non-default locale is explicitly used in pathname (after basePath)',
   ).toBe('fr')
 })
+
+describe.only('404 caching', () => {
+  describe('404 without getStaticProps', () => {
```
orinokai, line 158: I found it was necessary to keep these tests because handling is different for 404 without `getStaticProps`, however I did remove cases for 'matching dynamic path without revalidate' because behaviour should always be the same as 'matching dynamic path with revalidate' and cache handler should be called in both cases

src/run/headers.ts
pieh: > Was it there as a precaution, or was there a specific case you were guarding against? @pieh

Precaution mainly (same like previous `cache-control`), but I think the change you made make sense overall - it does widen scenarios of 404 caching where it possibly could not be intentional, but at the same time it catch more cases of 404 not being cached today

src/run/headers.ts
```diff
@@ -242,14 +250,27 @@ export const setCacheControlHeaders = (
...
+      // temporary CDN Cache Control handling for bot probes on PHP files
+      // https://linear.app/netlify/issue/FRB-1344/prevent-excessive-ssr-invocations-due-to-404-routes
+      headers.set('cache-control', 'public, max-age=0, must-revalidate')
+      headers.set('netlify-cdn-cache-control', `max-age=31536000, durable`)
+      return
+    }
+
+    if (
+      process.env.CACHE_404_PAGE &&
+      request.url.endsWith('/404') &&
```
orinokai, line 265: I know this has the potential for false positives if a site had a path like, `/products/404`, this feature is behind an env var and only intended for a single customer, so I think it's not worth the effort to make this any stricter

PR #2765, opened 2025-02-19
feat: make CDN SWR background revalidation discard stale cache content in order to produce fresh responses

pieh:
## Description

We are currently having "double SWR":
 - 1st level on edge
 - 2nd level in serverless handler

This leads to us having to instruct edge not to cache stale responses served by Next serverless handler as otherwise it would lead to stale content lingering for longer than it should.

This is making use of new request headers and for background SWR revalidation invocation - we instruct cache handler to discard stale entries. Do note - that it might result in discarding `APP_PAGE` entry when it's stale, but not necessarily discarding `FETCH` entry (if it's not stale) - this is behavior we see on Vercel and it is intentional.

## Tests

Initially testing was done manually for various scenarios comparing those changes both against the way things work today and when deployed on Vercel - from those manual tests things seem to be working correctly:
 a) served content remain the same as with current setup (just it leads to less invocations and ability to serve content directly from edge caches without blocking on serverless responses)
 b) comparing against behavior when deployed on Vercel and reported status ([`x-vercel-cache`](https://vercel.com/docs/edge-network/caching#x-vercel-cache))

For our automated tests - added test case that should ensure that responses (minus very first one) are from edge or durable caches and that responses have expected content which should prove that background revalidation requests results in producing fresh responses that can be cached on the edge/in durable

resolves https://linear.app/netlify/issue/FRB-1639/investigate-disabling-double-swr-behavior-in-next-runtime


tests/e2e/page-router.test.ts
```diff
@@ -390,6 +390,63 @@ test.describe('Simple Page Router (no basePath, no i18n)', () => {
...
+    expect(response1?.headers()['netlify-cdn-cache-control']).toMatch(
+      /s-maxage=60, stale-while-revalidate=[0-9]+, durable/,
+    )
+
+    // ensure response was NOT produced before invocation
+    const date1 = (await page.textContent('[data-testid="date-now"]')) ?? ''
+    expect(date1.localeCompare(beforeFirstFetch)).toBeGreaterThan(0)
+
+    // allow page to get stale
+    await page.waitForTimeout(60_000)
```
pieh, line 416: :( need for Durable so reduce potential flakiness if different edge nodes would be used - tested page has 60s revalidate time because of that as well

tests/fixtures/page-router/netlify/functions/purge-cdn.ts
```diff
@@ -0,0 +1,41 @@
...
+      {
+        status: 500,
+      },
+    )
+  }
+}
+
+export const config: Config = {
+  path: '/api/purge-cdn',
+}
```
pieh, line 41: I had some troubles with this function being in Next.js api handler - in the end I don't rely on it for added test, but overall it doesn't really make sense to use Next.js API for it, if it can use just vanilla Netlify function, as it only purge cdn cache, and not any Next.js caches in blobs

src/run/handlers/request-context.cts
```diff
@@ -41,7 +45,20 @@ type RequestContextAsyncLocalStorage = AsyncLocalStorage<RequestContext>
 export function createRequestContext(request?: Request, context?: FutureContext): RequestContext {
   const backgroundWorkPromises: Promise<unknown>[] = []
 
+  const isDebugRequest =
+    request?.headers.has('x-nf-debug-logging') || request?.headers.has('x-next-debug-logging')
```
mrstork, line 49: Just comparing this to https://github.com/opennextjs/opennextjs-netlify/pull/2769/files
Should this be `debug-x-nf-logging`?

src/run/handlers/cache.cts
```diff
@@ -52,6 +52,27 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
...
 
+  private getTTL(blob: NetlifyCacheHandlerValue) {
+    if (
+      blob.value?.kind === 'FETCH' ||
+      blob.value?.kind === 'ROUTE' ||
+      blob.value?.kind === 'APP_ROUTE' ||
+      blob.value?.kind === 'PAGE' ||
+      blob.value?.kind === 'PAGES' ||
+      blob.value?.kind === 'APP_PAGE'
+    ) {
```
mrstork, line 63: Is this list exhaustive?

mrstork (dismissed):

src/run/handlers/cache.cts
pieh: This is all `kind`s that we do support (and not hit this case https://github.com/opennextjs/opennextjs-netlify/blob/f3e24b1d2e4674574eef4c628d58b2d2a41e0be9/src/run/handlers/cache.cts#L285-L286

But right now this is manually constructed and there is no guard rail ensuring this list get's updated and in sync in the future.

Will think about best way of ensuring that all supported `kind`s are in here (but I would do this probably as follow up in case I don't figure out quick/easy solution)

pieh: Ok, there is a way but it's rather not pretty - full `tsc` check results in pretty awful errors then that are really non-descript (if we miss kind) and the whole setup looks pretty iffy, so I would rather not go this road.

https://typescript-eslint.io/rules/switch-exhaustiveness-check/ would be great for this ... but it doesn't really work with version of eslint-typescript we have - we would need some major upgrades of this ... which would require major upgrades of bunch of other things. Those major upgrades also change default configs.

And lastly - we don't even control those versions ourselves - and shared config we use was last released ... over 2 years ago ( https://github.com/netlify/eslint-config-node/releases )

So, I won't be touching this in this PR and will keep it as is

src/run/handlers/cache.cts
```diff
@@ -52,6 +52,27 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
...
+      const { revalidate } = blob.value
+
+      if (typeof revalidate === 'number') {
+        const revalidateAfter = revalidate * 1_000 + blob.lastModified
+        return (revalidateAfter - Date.now()) / 1_000
+      }
+      return revalidate === false ? 'PERMANENT' : 'NOT SET'
```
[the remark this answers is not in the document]
pieh: Done in https://github.com/opennextjs/opennextjs-netlify/pull/2765/commits/5f084d49467d95194fa2e91d4923b8f4512afc58

src/run/handlers/server.ts
```diff
@@ -103,15 +108,31 @@ export default async (request: Request) => {
     const response = await toComputeResponse(resProxy)
 
     if (requestContext.responseCacheKey) {
-      span.setAttribute('responseCacheKey', requestContext.responseCacheKey)
+      topLevelSpan.setAttribute('responseCacheKey', requestContext.responseCacheKey)
     }
 
-    await adjustDateHeader({ headers: response.headers, request, span, tracer, requestContext })
+    const nextCache = response.headers.get('x-nextjs-cache')
+    const isServedFromCache = nextCache === 'HIT' || nextCache === 'STALE'
```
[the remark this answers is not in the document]
pieh: Done in https://github.com/opennextjs/opennextjs-netlify/pull/2765/commits/5f084d49467d95194fa2e91d4923b8f4512afc58

src/run/handlers/server.ts
```diff
@@ -138,6 +169,20 @@ export default async (request: Request) => {
...
+      ...expandHeaders('response.headers', response.headers),
+    })
+
+    // requestContext.logger
+    //   .withFields({
+    //     isBackgroundRevalidation: requestContext.isBackgroundRevalidation,
+    //     ...expandHeaders('request.headers', request.headers),
+    //     ...expandHeaders('response.headers', response.headers),
+    //   })
+    //   .debug('[NetlifyNextRuntime] Response cache headers')
```
mrstork, line 184: I typically don't like keeping comments like this around. My preference is to wrap this in a debug flag (if possible) and when set to true it logs the line, or remove the comment

mrstork (approved):

src/run/handlers/server.ts
pieh: Ops, those were not meant to be commited - I will remove them next time I have a chance

pieh: Ammended last commit without unintended commented out code - https://github.com/opennextjs/opennextjs-netlify/pull/2765/commits/45f3a0a51d32ad7a42ae0b4a21b81cef46cf35a8

mrstork (approved):

PR #2766, opened 2025-02-20
fix: don't cache POST when serving embedded static html

pieh:
## Description

This doesn't fix any known issue, just noticed we didn't check for method in this case so just for safety, we should


orinokai (approved):
Good catch!

PR #2769, opened 2025-02-26
fix: expect renamed debug headers

serhalp:
## Description

The debug response headers were renamed a few weeks ago

### Documentation

N/A - undocumented

## Tests

🤔

https://github.com/netlify/stargate/pull/2666


tests/utils/fixture.ts
```diff
@@ -473,17 +473,17 @@ export async function invokeEdgeFunction(
...
+      'debug-x-nf-blobs-info': Buffer.from(
         JSON.stringify({ url: `http://${ctx.blobStoreHost}`, token: BLOB_TOKEN }),
       ).toString('base64'),
-      'x-nf-passthrough': 'passthrough',
-      'x-nf-passthrough-host': passthroughHost,
-      'x-nf-passthrough-proto': 'http:',
+      'debug-x-nf-passthrough': 'passthrough',
+      'debug-x-nf-passthrough-host': passthroughHost,
+      'debug-x-nf-passthrough-proto': 'http:',
       'x-nf-request-id': v4(),
```
pieh, line 487: Those probably should not change as they are used to make `@netlify/edge-bundler` serve work. Looking at how CLI handles those:
- https://github.com/netlify/cli/blob/e5c6a136c94f9c786c4c39a7daecddda2070473c/src/lib/edge-functions/proxy.ts#L151-L190
- https://github.com/netlify/cli/blob/e5c6a136c94f9c786c4c39a7daecddda2070473c/src/lib/edge-functions/headers.ts#L3-L20

Those should not be prefixed as they are request headers that setup edge function invocations and not response headers

Also https://github.com/netlify/edge-functions-bootstrap/blob/e49ee4a71e417c5ba91759a24f5b6ae0bd3ecc2c/src/bootstrap/headers.ts#L1-L41 is probably where those headers are used to make things work and they are not prefixed there

PR #2774, opened 2025-03-17
fix: adjust bundling to not produce duplicate inlined internal modules

pieh:
## Description

We are producing duplicate inlining in built package if built in directory that is not named `opennextjs-netlify`. This adjusts build script to not rely on directory name and adds some validation.

This is still a bit bandaidy - ideally we don't do multiple passes and do single build emitting mix of ESM and CJS, but I couldn't find to make `esbuild` behave in this way, so to not overhaul entire setup and spend a lot of time on it - I just made some adjustments and really barebones verification/validation of built package.

## Tests

Adding some validation to build script

https://linear.app/netlify/issue/FRB-1681/fix-next-runtime-bundling-so-we-dont-duplicateinline-internal-modules

### DIFF
--- a/tools/build.js
+++ b/tools/build.js
@@ -147,7 +147,7 @@ async function ensureNoRegionalBlobsModuleDuplicates() {
   }
   if (!foundInExpectedModule) {
     throw new Error(
-      'Expected to find "fetchBeforeNextPatchedIt" variable in "run/regional-blob-store.cjs", but it was not found. This might indicate setup change that require bundling validation in "tools/build.js" to be adjusted.',
+      'Expected to find "fetchBeforeNextPatchedIt" variable in "run/regional-blob-store.cjs", but it was not found. This might indicate a setup change that requires the bundling validation in "tools/build.js" to be adjusted.',
     )
   }
   if (unexpectedModulesContainingFetchBeforeNextPatchedIt.length !== 0) {


tools/build.js
```diff
@@ -126,8 +129,38 @@ await Promise.all([
...
+  if (!foundInExpectedModule) {
+    throw new Error(
+      'Expected to find fetchBeforeNextPatchedIt in run/regional-blob-store.cjs, but it was not found. This might indicate setup change that require bundling validation to be adjusted.',
+    )
+  }
+  if (unexpectedModulesContainingFetchBeforeNextPatchedIt.length !== 0) {
+    throw new Error(
+      `Bundling produced unexpected duplicates of 'regional-blob-store' module in following built modules:\n${unexpectedModulesContainingFetchBeforeNextPatchedIt.map((filePath) => ` - ${filePath}`).join('\n')}`,
+    )
+  }
```
pieh, line 157: This produce error like this:
```
Error: Bundling produced unexpected duplicates of 'regional-blob-store' module in following built modules:
 - dist/esm-chunks/chunk-K4RDUZYO.js
 - dist/run/next.cjs
 - dist/run/handlers/cache.cjs
    at ensureNoRegionalBlobsModuleDuplicates (file:///Users/misiek/dev/pgs-next-runtime/tools/build.js:154:11)
    at async file:///Users/misiek/dev/pgs-next-runtime/tools/build.js:163:3
```
The main purpose is to prevent publishing mis-bundlend version of runtime from local machines

tools/build.js
```diff
@@ -39,7 +42,7 @@ async function bundle(entryPoints, format, watch) {
         name: 'mark-runtime-modules-as-external',
         setup(pluginBuild) {
           pluginBuild.onResolve({ filter: /^\..*\.c?js$/ }, (args) => {
-            if (args.importer.includes(join('opennextjs-netlify', 'src'))) {
+            if (args.importer.includes(join(repoDirectory, 'src'))) {
```
pieh, line 45: Main fix/adjustment - we were publishing runtime locally since move to Opennext organization and if locally repo clone directory was not named exactly `opennextjs-neltify`, then we would produce duplicates.

tests/e2e/simple-app.test.ts
```diff
@@ -12,8 +12,8 @@ test('Renders the Home page correctly', async ({ page, simple }) => {
 
   await expect(page).toHaveTitle('Simple Next App')
 
-  expect(headers['cache-status']).toMatch(/^"Next.js"; hit$/m)
-  expect(headers['cache-status']).toMatch(/^"Netlify Edge"; fwd=miss$/m)
+  expect(headers['cache-status'].replaceAll(', ', '\n')).toMatch(/^"Next.js"; hit$/m)
+  expect(headers['cache-status'].replaceAll(', ', '\n')).toMatch(/^"Netlify Edge"; fwd=miss$/m)
```
pieh, line 16: Not related to main portion of PR - seems like we might be getting comma separated cache-status values, so I'm massing it to have each statement as separate line to match regex assumption

tools/build.js
```diff
@@ -126,8 +129,38 @@ await Promise.all([
...
+      if (fileToTest.endsWith('run/regional-blob-store.cjs')) {
+        foundInExpectedModule = true
+      } else {
+        unexpectedModulesContainingFetchBeforeNextPatchedIt.push(fileToTest)
+      }
+    }
+  }
+  if (!foundInExpectedModule) {
+    throw new Error(
+      'Expected to find "fetchBeforeNextPatchedIt" variable in "run/regional-blob-store.cjs", but it was not found. This might indicate setup change that require bundling validation in "tools/build.js" to be adjusted.',
```
eduardoboucas, line 150: nit

```suggestion
      'Expected to find "fetchBeforeNextPatchedIt" variable in "run/regional-blob-store.cjs", but it was not found. This might indicate a setup change that requires the bundling validation in "tools/build.js" to be adjusted.',
```

eduardoboucas (approved):

eduardoboucas (approved):

PR #2775, opened 2025-03-17
fix: add more measures to prevent using data-cache for blob operations

pieh:
## Tests

1. TODO


src/run/regional-blob-store.cts
```diff
@@ -1,11 +1,62 @@
...
+  extendedGlobalThis[FETCH_BEFORE_NEXT_PATCHED_IT] = forceOptOutOfUsingDataCache(
+    attemptToGetOriginalFetch(fetch),
+  )
+}
+
+const fetchBeforeNextPatchedItFallback = forceOptOutOfUsingDataCache(
+  attemptToGetOriginalFetch(globalThis.fetch),
+)
+const getFetchBeforeNextPatchedIt = () =>
+  extendedGlobalThis[FETCH_BEFORE_NEXT_PATCHED_IT] ?? fetchBeforeNextPatchedItFallback
```
orinokai, line 54: Is this necessary? You're calling `setFetchBeforeNextPatchedIt` in the global scope of `server.ts`, so would the fallback ever be used?

orinokai (approved):
All good, just one non-blocking question

src/run/regional-blob-store.cts
pieh: It should not be needed when running full on server handler, no. 

I did want to make this module to not necessarily rely on setup (might be useful for unit tests, tho not sure if we actually have any that would would fall into using this module today)

pieh: This is also just using pattern of allowing few points of failure because of overall messiness with patched fetch - if for whatever reason we don't call `setFetchBeforeNextPatchedIt` before we execute `getFetchBeforeNextPatchedIt` - I do want to use rest of defensive measures (such as trying to grab `_originalFetch` first and ultimately adding `next.internal = true` param to `fetch` calls) and continue to have this fallback just fits the best here (IMO)

pieh: So I'll go ahead and merge as-is

PR #2777, opened 2025-03-19
perf: memoize blobs requests in the request scope

pieh:
## Description

If site is doing multiple fetches with data caching enabled or is calling method wrapped in `unstable_cache`  resulting in same cache key multiple times to render single page - we currently are repeatedly calling `blobStore.get` for each of the repeated use.

This PR looks to add memoization of those calls scoped to concrete request. We don't want to memoize them globally due to distributed nature of serverless with potentially many instances running at the same time, each holding different potentially state, but we should be able to safely memoize blobs call while handling same request. This also allow for reading it's own writes within scope of same request handling.

This PR also replace similar handling we already had for tag manifests with more generic one shared for all blobs operations.

## Tests

- Added fetch specific integration test asserting we only fetch same thing from blobs once per request
- Added memoized blob store unit tests
- We already did have tests specific to tag manifests that I will not change, but they should continue to pass ( for example https://github.com/opennextjs/opennextjs-netlify/blob/8675b3ecf40f0e8f79ad68737e9435d36ede3f3f/tests/integration/revalidate-path.test.ts#L77-L81 )

https://linear.app/netlify/issue/FRB-1629/intermittent-proxy-timeouts-on-runtime-v594


src/run/regional-blob-store.cts
```diff
@@ -1,8 +1,23 @@
...
+const NullValue = Symbol.for('null-value')
+type BlobLRUCache = LRUCache<string, BlobType | typeof NullValue | Promise<BlobType | null>>
 
 const FETCH_BEFORE_NEXT_PATCHED_IT = Symbol.for('nf-not-patched-fetch')
+const IN_MEMORY_CACHE_MAX_SIZE = Symbol.for('nf-in-memory-cache-max-size')
+const IN_MEMORY_LRU_CACHE = Symbol.for('nf-in-memory-lru-cache')
 const extendedGlobalThis = globalThis as typeof globalThis & {
   [FETCH_BEFORE_NEXT_PATCHED_IT]?: typeof globalThis.fetch
+  [IN_MEMORY_CACHE_MAX_SIZE]?: number
+  [IN_MEMORY_LRU_CACHE]?: BlobLRUCache | null
```
mrstork, line 20: Move `noOpInMemoryCache` here? Instead of `null`?

src/run/regional-blob-store.cts
```diff
@@ -53,10 +68,131 @@ const fetchBeforeNextPatchedItFallback = forceOptOutOfUsingDataCache(
...
+  if (typeof extendedGlobalThis[IN_MEMORY_LRU_CACHE] === 'undefined') {
+    const maxSize =
+      typeof extendedGlobalThis[IN_MEMORY_CACHE_MAX_SIZE] === 'number'
+        ? extendedGlobalThis[IN_MEMORY_CACHE_MAX_SIZE]
+        : DEFAULT_FALLBACK_MAX_SIZE
+
+    extendedGlobalThis[IN_MEMORY_LRU_CACHE] =
+      maxSize === 0
+        ? null // if user sets 0 in their config, we should honor that and not use in-memory cache
+        : new LRUCache<string, BlobType | typeof NullValue | Promise<BlobType | null>>({
```
mrstork, line 95: Is `LRUCache` a lot better than just using `{}` for example?

src/run/regional-blob-store.cts
```diff
@@ -53,10 +68,131 @@ const fetchBeforeNextPatchedItFallback = forceOptOutOfUsingDataCache(
...
+    },
+  }
+}
+
+const encodeBlobKey = async (key: string) => {
+  const { encodeBlobKey: encodeBlobKeyImpl } = await import('../shared/blobkey.js')
+  return await encodeBlobKeyImpl(key)
+}
+
+export const getMemoizedKeyValueStoreBackedByRegionalBlobStore = (
```
mrstork, line 151: This looks like a whole other feature, potentially should be in another PR?

src/run/handlers/cache.cts
```diff
@@ -492,7 +465,8 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
   }
 
   resetRequestCache() {
```
mrstork, line 467: Now that it's a no-op, are we free to remove this function or does next still call this?

src/run/regional-blob-store.cts
```diff
@@ -53,10 +68,131 @@ const fetchBeforeNextPatchedItFallback = forceOptOutOfUsingDataCache(
...
+const noOpInMemoryCache: RequestSpecificInMemoryCache = {
+  get(): undefined {
+    // no-op
+  },
+  set() {
+    // no-op
+  },
+}
+
+const getRequestSpecificInMemoryCache = (): RequestSpecificInMemoryCache => {
```
mrstork, line 120: This looks like a interface/wrapper around LRUCache, should we pop this in another file? Perhaps `in-memory-cache.cts`?

src/shared/cache-types.cts
```diff
@@ -154,3 +158,60 @@ export type CacheHandlerForMultipleVersions = BaseCacheHandlerForMultipleVersion
...
+  isFallback: boolean
+}
+
+export type BlobType = NetlifyCacheHandlerValue | TagManifest | HtmlBlob
+
+const isTagManifest = (value: BlobType): value is TagManifest => {
+  return false
+}
+
+const isHtmlBlob = (value: BlobType): value is HtmlBlob => {
```
mrstork, line 175: This looks like it belongs in another file as it is not a "cache type"

src/shared/cache-types.cts
```diff
@@ -154,3 +158,60 @@ export type CacheHandlerForMultipleVersions = BaseCacheHandlerForMultipleVersion
...
+  return false
+}
+
+export const estimateBlobSize = (valueToStore: BlobType | null | Promise<unknown>): number => {
```
mrstork, line 179: Same here, this is not a "cache type" so I think we should put this in a different file

src/shared/cache-types.cts
```diff
@@ -154,3 +158,60 @@ export type CacheHandlerForMultipleVersions = BaseCacheHandlerForMultipleVersion
...
+  // very approximate size calculation to avoid expensive exact size calculation
+  // inspired by https://github.com/vercel/next.js/blob/ed10f7ed0246fcc763194197eb9beebcbd063162/packages/next/src/server/lib/incremental-cache/file-system-cache.ts#L60-L79
+  if (valueToStore === null || isPromise(valueToStore) || isTagManifest(valueToStore)) {
+    return 25
+  }
+  if (isHtmlBlob(valueToStore)) {
```
mrstork, line 185: I might be reading the function incorrectly, but it seems like `isHtmlBlob` always returns false

src/run/handlers/tracer.cts
```diff
@@ -89,3 +89,16 @@ export function getTracer(): RuntimeTracer {
 
   return tracer
 }
+
+export function recordWarning(warning: Error, span?: Span) {
```
mrstork, line 93: `getCacheKeySpan` looked to be non-nullable, can we make `span` non-nullable here as well?

src/run/regional-blob-store.cts
```diff
@@ -1,8 +1,23 @@
...
+import { getRequestContext } from './handlers/request-context.cjs'
+import { getTracer } from './handlers/tracer.cjs'
+
+// lru-cache types don't like using `null` for values, so we use a symbol to represent it and do conversion
+// so it doesn't leak outside
+const NullValue = Symbol.for('null-value')
+type BlobLRUCache = LRUCache<string, BlobType | typeof NullValue | Promise<BlobType | null>>
 
 const FETCH_BEFORE_NEXT_PATCHED_IT = Symbol.for('nf-not-patched-fetch')
+const IN_MEMORY_CACHE_MAX_SIZE = Symbol.for('nf-in-memory-cache-max-size')
```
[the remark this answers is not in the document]
pieh: Customer wouldn't use this directly. Idea was to honor actual `next.config.js` setting https://nextjs.org/docs/app/api-reference/config/next-config-js/incrementalCacheHandlerPath (`cacheMaxMemorySize` one) that I do wire up here - https://github.com/opennextjs/opennextjs-netlify/blob/41bca906f838cc7fb1e45ea722db75826422b92e/src/run/config.ts#L53-L56

Note checking 2 settings because https://github.com/vercel/next.js/pull/57953 landed in 14.1.0 that moved experimental one into stable (and renamed it)

src/run/regional-blob-store.cts
pieh: `LRUCache` ensures some limits, so that lambda that stays warm for a long time doesn't suddenly OOM. The memory/size is not exact (hence naming of `estimate` for the size calculation) and should be thought more as "ballpark" values.

If we don't use `LRUCache` - we'd need to come up with some other mechanism to manage in-memory pruning, but I don't think it's worth doing given how battle-tested `lru-cache` package is (with 200 million downloads a wek).

Lastly - default implementation for self-hosted Next.js also uses LRUCache for its in-memory cache (size calculation was inspired by how it's used in that default implementation)

src/run/regional-blob-store.cts
```diff
@@ -53,10 +68,131 @@ const fetchBeforeNextPatchedItFallback = forceOptOutOfUsingDataCache(
...
+            maxSize,
+            sizeCalculation: (valueToStore) => {
+              return estimateBlobSize(valueToStore === NullValue ? null : valueToStore)
+            },
+          })
+  }
+  return extendedGlobalThis[IN_MEMORY_LRU_CACHE]
+}
+
+interface RequestSpecificInMemoryCache {
```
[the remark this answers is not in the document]
pieh: Yeah, naming is hard - I will use one of suggestions

src/run/regional-blob-store.cts
pieh: Good idea!

src/run/handlers/tracer.cts
pieh: https://github.com/opennextjs/opennextjs-netlify/blob/41bca906f838cc7fb1e45ea722db75826422b92e/src/shared/cache-types.cts#L210-L214 this call doesn't have access to `span` to pass it, so that's why I made it optional.

Generally preferably we do pass span explicitly instead of getting "active span", because it's more stable on which span things would be recorded, but it's better to record at all than skip recording because we don't have reference to a span.

Alternative could be to make above linked code get active span and then make `span` here non-nullable, but I did want to simplify usage of it, but would be fine making that change (less "magic")

src/shared/cache-types.cts
pieh: Ops, yes - this is actually missing - thanks for spotting. This would result in getting into fallback and produce warnings (tho things would work)

src/run/regional-blob-store.cts
mrstork: In general I will always lean towards single-focus PRs, but let's leave things as they are this time around as your explanation gives enough context for me to review here.

src/shared/cache-types.cts
pieh: Yeah, this is more of possible value that is being stored in blobs, but not exactly cache type. TagManifest is also same way, because it's also not really a cache type (it's not comming from next.js - it's our custom handling for tracking tag invalidations).

src/run/regional-blob-store.cts
mrstork: I realized these are types, but the intention of my comment still sort of stands. Can we try to have these be non-nullable if possible?

pieh: `noOpInMemoryCache` is not implementing same interface as `BlobLRUCache` so it couldn't be used here. To make it non-nullable I would need to create a shim class that implements `LRUCache` interface and this doesn't seem worth doing and maintaing to get rid of `null` case to me?

src/run/regional-blob-store.cts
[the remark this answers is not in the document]
pieh: Renamed to `RequestScopedInMemoryCache` in https://github.com/opennextjs/opennextjs-netlify/pull/2777/commits/42977edfb101276ccb4dcd30144000e4f5aac210

src/run/regional-blob-store.cts
pieh: I did a "refactor" in https://github.com/opennextjs/opennextjs-netlify/pull/2777/commits/cff9f73dfefce8feb047e3177251c9b48d6b8f4c that did few things:

1. I did create `run/storage` directory to not pollute `run` with "implementation details" of storage handling
2. There is "public" module `run/storage/storage.ts` that is meant to be used everywhere and other modules in `run/storage` should not be imported directly outside of `storage` directory to not leak abstractions/implementation details (added eslint rule for that)
3. `regional-blob-store.cts` is now exactly as it is in `main` branch - it was just moved (with adjusted import paths only) - this now is not meant to be used directly because it's considered implementation detail
4. created `request-scoped-in-memory-cache.cts` and moved details of in-memory handling there (this is also implementation detail not meant to be used directly outside of `run/storage`)

Hopefully this splits concerns reasonably well, making modules rather small and hopefully easier to digest and provide some boundaries between concerns with standard import/export contract

src/shared/cache-types.cts
pieh: Did some shuffling in https://github.com/opennextjs/opennextjs-netlify/pull/2777/commits/ca233efee8b35879bfeeaf491138f380bbe3a642 - i did create sibling module `blob-types` and moved those non Next.js CacheHandler related types and type guards there. Actually fixing type guards ( https://github.com/opennextjs/opennextjs-netlify/pull/2777#discussion_r2014445442 ) will be done in separate commit

The main thing with this move to me was question about `estimateBlobSize` and wether this should be in this new module or it should be in the `in-memory` module, but decided to move it to `in-memory` module because size is only concern/detail of `in-memory` cache and not shared anywhere else (at least not right now). Maybe it could have it's own module, but to me over-modularizing make things less readable when you have to jump too much between different modules 🤷

src/shared/cache-types.cts
pieh: See https://github.com/opennextjs/opennextjs-netlify/pull/2777#discussion_r2016264366

src/shared/cache-types.cts
pieh: Hopeffully fixed in https://github.com/opennextjs/opennextjs-netlify/pull/2777/commits/578fbc54fb33f29febea2bf770af713e817c19b9 with some added unit tests

src/run/handlers/tracer.cts
pieh: I did attempt the alternative but to me it moved to much tracing related logic and checks to module that doesn't already have reference to containing span I mentioned above and concern of finding active span seems much more in `tracer.cts` module than modules that might want to record warnings, so I'd prefer to keep the signature of this function as it is

src/run/storage/request-scoped-in-memory-cache.cts
```diff
@@ -0,0 +1,108 @@
...
+// lru-cache types don't like using `null` for values, so we use a symbol to represent it and do conversion
+// so it doesn't leak outside
+const NullValue = Symbol.for('null-value')
+type BlobLRUCache = LRUCache<string, BlobType | typeof NullValue | Promise<BlobType | null>>
+
+const IN_MEMORY_CACHE_MAX_SIZE = Symbol.for('nf-in-memory-cache-max-size')
+const IN_MEMORY_LRU_CACHE = Symbol.for('nf-in-memory-lru-cache')
+const extendedGlobalThis = globalThis as typeof globalThis & {
+  [IN_MEMORY_CACHE_MAX_SIZE]?: number
+  [IN_MEMORY_LRU_CACHE]?: BlobLRUCache | null
```
mrstork, line 18: Noting that we're using three types here `BlobLRUCache | null | undefined`

src/run/storage/request-scoped-in-memory-cache.cts
```diff
@@ -0,0 +1,108 @@
...
+            sizeCalculation: (valueToStore) => {
+              return estimateBlobSize(valueToStore === NullValue ? null : valueToStore)
+            },
+          })
+  }
+  return extendedGlobalThis[IN_MEMORY_LRU_CACHE]
+}
+
+interface RequestScopedInMemoryCache {
+  get(key: string): BlobType | null | Promise<BlobType | null> | undefined
```
mrstork, line 90: Noting here that we're using `undefined` and `null` as distinctly different values

src/run/storage/storage.cts
```diff
@@ -0,0 +1,68 @@
...
+  const store = getRegionalBlobStore(...args)
+  const tracer = getTracer()
+
+  return {
+    async get<T extends BlobType>(key: string, otelSpanTitle: string): Promise<T | null> {
+      const inMemoryCache = getRequestScopedInMemoryCache()
+
+      const memoizedValue = inMemoryCache.get(key)
+      if (typeof memoizedValue !== 'undefined') {
+        return memoizedValue as T | null | Promise<T | null>
```
mrstork, line 29: We convert `undefined` back to null here, so perhaps we could push this change upstream to simplify the interface

src/run/handlers/server.ts
```diff
@@ -13,8 +13,8 @@ import {
   setCacheTagsHeaders,
   setVaryHeaders,
 } from '../headers.js'
-import { setFetchBeforeNextPatchedIt } from '../regional-blob-store.cjs'
 import { nextResponseProxy } from '../revalidate.js'
+import { setFetchBeforeNextPatchedIt } from '../storage/storage.cjs'
```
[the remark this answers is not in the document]
pieh: Only alternative really would be to not call this function (and just remove this function altogether) and instead directly store fetch on `globalThis` to be consumed later

.eslintrc.cjs
```diff
@@ -72,6 +72,14 @@ module.exports = {
                 message: 'Please use `getTracer()` from `./handlers/tracer.cjs` instead',
               },
             ],
+            patterns: [
+              {
+                // only */storage/storage.cjs is allowed to be imported
+                // rest are implementation details that should not be used directly
+                group: ['*/storage/*', '!*/storage/storage.cjs'],
```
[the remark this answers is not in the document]
pieh: I personally prefer not to because my personal experience is that using `index` modules makes navigation around code base harder when you start to get multiple modules with same name in different directories, as well as just seeing multiple `index` modules opened in IDE tabs hellish so I do try to avoid it

similar example (just instead of `index` module, is Next's app router pattern to have `page.j/ts(x)`):
image

I can appreciate that `storage/storage.cjs` looks pretty weird, but I think it's better than added mental load of tracking multiple `index` modules. Additionally with IDEs auto-adding imports (like https://code.visualstudio.com/docs/languages/javascript#_auto-imports ) it's a bit "out of mind" as nowadays you usually don't even write your import statements manually

Some compromise could be to have index that just re-exports and doesn't actually have any logic inside, but then it just pollute directory tree IMO

src/run/handlers/cache.cts
```diff
@@ -476,7 +449,7 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
     await Promise.all(
       tags.map(async (tag) => {
         try {
-          await this.blobStore.setJSON(await this.encodeBlobKey(tag), data)
+          await this.cacheStore.set(tag, data, 'tagManifest.set')
```
mrstork, line 452: Just noting that this will add tracing (I believe that's a good thing)

src/run/next.cts
```diff
@@ -97,11 +92,9 @@ export async function getMockedRequestHandler(...args: Parameters<typeof getRequ
       } catch (error) {
         // only try to get .html files from the blob store
         if (typeof path === 'string' && path.endsWith('.html')) {
-          const store = getRegionalBlobStore()
+          const cacheStore = getMemoizedKeyValueStoreBackedByRegionalBlobStore()
           const relPath = relative(resolve('.next/server/pages'), path)
-          const file = (await store.get(await encodeBlobKey(relPath), {
-            type: 'json',
-          })) as HtmlBlob | null
+          const file = await cacheStore.get<HtmlBlob>(relPath, 'staticHtml.get')
```
mrstork, line 97: Again noting that this will add tracing (again I think it's a good thing)

src/run/next.cts
```diff
@@ -97,11 +92,9 @@ export async function getMockedRequestHandler(...args: Parameters<typeof getRequ
       } catch (error) {
         // only try to get .html files from the blob store
         if (typeof path === 'string' && path.endsWith('.html')) {
-          const store = getRegionalBlobStore()
+          const cacheStore = getMemoizedKeyValueStoreBackedByRegionalBlobStore()
```
mrstork, line 95: Thoughts on using `getStore<HtmlBlob>.get` instead of `getStore.get<HtmlBlob>` (not putting the full name)?

src/run/next.cts
mrstork, line 97: It's probably automatic, but I'm just calling it out in case - I see `blobStore.set` and `tagManifest.set` but not seeing `staticHtml.set` anywhere.

src/run/handlers/cache.cts
pieh: Yes, I did want to use this opportunity to capture all blobs operations for debugging purposes

src/run/next.cts
pieh: We upload those only at build time ( https://github.com/opennextjs/opennextjs-netlify/blob/b5754f40721dc8438bcb1aaf1df1a2727a726a36/src/build/content/static.ts#L42-L46 which will make use of [File-based blobs uploads](https://docs.netlify.com/blobs/overview/#file-based-uploads). As we only copy and let `@nelify/cli` or `buildbot` upload those blobs after build command - we don't create build-time traces for that.

We don't add/update those anymore once deployed

src/run/headers.ts
```diff
@@ -149,45 +146,29 @@ export const adjustDateHeader = async ({
...
-      severity: 'alert',
-      warning: true,
-    })
 
-    const blobStore = getRegionalBlobStore({ consistency: 'strong' })
-    const blobKey = await encodeBlobKey(key)
-
-    // TODO: use metadata for this
-    lastModified = await tracer.withActiveSpan(
+    const cacheStore = getMemoizedKeyValueStoreBackedByRegionalBlobStore({ consistency: 'strong' })
```
mrstork, line 154: Naming note - perhaps `getMemoizedRegionalBlobStore` or `getRegionalBlobService`?

src/run/storage/storage.cts
```diff
@@ -0,0 +1,68 @@
...
+        span.setAttributes({ key, blobKey })
+        const blob = (await store.get(blobKey, { type: 'json' })) as T | null
+        inMemoryCache.set(key, blob)
+        span.addEvent(blob ? 'Hit' : 'Miss')
+        return blob
+      })
+      inMemoryCache.set(key, getPromise)
+      return getPromise
+    },
+    async set(key: string, value: BlobType, otelSpanTitle: string) {
```
mrstork, line 43: Not sure if this works, but maybe this?
```
async set<T extends BlobType>(key: string, value: T, otelSpanTitle: string) {
```

src/run/storage/storage.cts
```diff
@@ -0,0 +1,68 @@
...
+
+import { getRegionalBlobStore } from './regional-blob-store.cjs'
+import { getRequestScopedInMemoryCache } from './request-scoped-in-memory-cache.cjs'
+
+const encodeBlobKey = async (key: string) => {
+  const { encodeBlobKey: encodeBlobKeyImpl } = await import('../../shared/blobkey.js')
+  return await encodeBlobKeyImpl(key)
+}
+
+export const getMemoizedKeyValueStoreBackedByRegionalBlobStore = (
```
mrstork, line 17: I'm hopeful that the 3 implementations of this don't diverge too much. If we find ourselves adding any if statements in here for each of the types, the responsible thing to do quickly becomes splitting off.

src/run/storage/request-scoped-in-memory-cache.cts
```diff
@@ -0,0 +1,109 @@
...
+
+import { LRUCache } from 'lru-cache'
+
+import { type BlobType, isHtmlBlob, isTagManifest } from '../../shared/blob-types.cjs'
+import { getRequestContext } from '../handlers/request-context.cjs'
+import { recordWarning } from '../handlers/tracer.cjs'
+
+// lru-cache types don't like using `null` for values, so we use a symbol to represent it and do conversion
+// so it doesn't leak outside
+const NullValue = Symbol.for('null-value')
```
mrstork, line 11: Is there a case today where we save `null` in the blob store?

src/run/storage/storage.cts
pieh: I only added generic to `.get` to save few characters as we currently any time we use `blobStore.get` we have to do something like this:
https://github.com/opennextjs/opennextjs-netlify/blob/b5754f40721dc8438bcb1aaf1df1a2727a726a36/src/run/handlers/cache.cts#L256

so it just avoid adding `| null` in all those places while preserving same type safety.

I don't think that there does need to be parity between `.get` and `.set` just for parity sake.

That said, adding generic to `.set` is fine and we could lock down to concrete types from available here https://github.com/opennextjs/opennextjs-netlify/blob/7daeeca6604ec8a874cabf4b1d71aabf776bd963/src/shared/blob-types.cts#L10 and it could technically increase type safety (so we don't ever try to set `TagManifest` when generally given code path should set `NetlifyCacheHandlerValue`), so if you feel up for it - you can go ahead and make that change.

src/run/storage/request-scoped-in-memory-cache.cts
mrstork: Answer - 404 pages in certain scenarios

src/run/storage/storage.cts
pieh: If we can move size calculation out of `storage` directory, we could add similar eslint rule like I did to not allow imports other than `BlobType` from the module that exports it in that directory to enforce storage not caring about details of individual types and only enforce that we use allowed union of types

mrstork (approved):

PR #2778, opened 2025-03-22
test: add test ensuring `GHSA-f82v-jwr5-mffw` doesn't affect Next.js on Netlify

pieh:
### Documentation

This is ensuring Next.js apps hosted on Netlify are not affected by https://github.com/advisories/GHSA-f82v-jwr5-mffw

Our custom edge middleware implementation doesn't make use of `x-middleware-subrequest` header. But just for good measure - let's add test ensuring that suppling this header to request against app using vulnerable Next.js version (15.2.2, so one patch before fix for self hosted)

### DIFF
--- a/tests/e2e/edge-middleware.test.ts
+++ b/tests/e2e/edge-middleware.test.ts
@@ -216,3 +216,19 @@ test.describe('Middleware with i18n and excluded paths', () => {
     })
   })
 })
+
+test("requests with x-middleware-subrequest don't skip middleware (GHSA-f82v-jwr5-mffw)", async ({
+  middlewareSubrequestVuln,
+}) => {
+  const response = await fetch(`${middlewareSubrequestVuln.url}`, {
+    headers: {
+      'x-middleware-subrequest': 'middleware:middleware:middleware:middleware:middleware',
+    },
+  })
+
+  // middleware was not skipped
+  expect(response.headers.get('x-test-used-middleware')).toBe('true')
+
+  // ensure we are testing version before the fix for self hosted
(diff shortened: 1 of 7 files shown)


tests/e2e/edge-middleware.test.ts
```diff
@@ -216,3 +216,19 @@ test.describe('Middleware with i18n and excluded paths', () => {
...
+      'x-middleware-subrequest': 'middleware:middleware:middleware:middleware:middleware',
+    },
+  })
+
+  // middleware was not skipped
+  expect(response.headers.get('x-test-used-middleware')).toBe('true')
+
+  // ensure we are testing version before the fix for self hosted
+  expect(response.headers.get('x-test-used-next-version')).toBe('15.2.2')
+})
```
pieh, line 234: This test is inspired by https://github.com/vercel/next.js/commit/52a078da3884efe6501613c7834a3d02a91676d2#diff-c5e08cd8ee165bce3a9fb81c8315d6e149b76f6f9b2e53f70bb2f39fb248d2aeR148-R157

Just assertions are hopefully more readable in their intention - goal is to ensure that middleware is not skipped when `x-middleware-subrequest` request header with 5+ hops is provided ( as then https://github.com/vercel/next.js/blob/cdb9a8c3b355593ab19ee6dec7c59e622bf59c81/packages/next/src/server/web/sandbox/sandbox.ts#L105-L114 this code path would be executed - at least when self hosting, but we don't even use this code path for our handling)

mrstork (approved):

PR #2786, opened 2025-03-26
Frb 1708/purge cache catch all routes

mrstork:
## Tests

1. TODO


src/run/handlers/cache.cts
```diff
@@ -121,11 +121,19 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
...
   private captureCacheTags(cacheValue: NetlifyIncrementalCacheValue | null, key: string) {
+    const requestContext = getRequestContext()
+
     if (!cacheValue) {
+      if (requestContext?.responseCacheTags) {
+        const cacheTags = [`_N_T_${key === '/index' ? '/' : encodeURI(key)}`]
+        requestContext.responseCacheTags = cacheTags
+
+        console.log('cache tags set:', cacheTags)
+      }
```
mrstork, line 132: TODO - We may not need this block

src/run/handlers/cache.cts
```diff
@@ -393,28 +401,28 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
...
-        const requestContext = getRequestContext()
-        if (requestContext?.didPagesRouterOnDemandRevalidate) {
-          // encode here to deal with non ASCII characters in the key
-          const tag = `_N_T_${key === '/index' ? '/' : encodeURI(key)}`
-          const tags = tag.split(/,|%2c/gi).filter(Boolean)
-
-          if (tags.length === 0) {
-            return
-          }
+      // if (data?.kind === 'PAGE' || data?.kind === 'PAGES') {
```
mrstork, line 404: TODO - can we remove this if entirely, or do we need to make it less restrictive so 404 pages can purge cache as well?

src/run/next.cts
```diff
@@ -99,7 +99,8 @@ export async function getMockedRequestHandler(...args: Parameters<typeof getRequ
             if (!file.isFallback) {
               const requestContext = getRequestContext()
               if (requestContext) {
-                requestContext.usedFsReadForNonFallback = true
```
mrstork, line 102: TODO - unrelated to this fix, remove these changes

src/run/headers.ts
```diff
@@ -281,13 +283,24 @@ export const setCacheControlHeaders = (
...
+
+  if (requestContext.responseCacheTags) {
     headers.set('netlify-cache-tag', requestContext.responseCacheTags.join(','))
+    return
   }
+
+  const key = new URL(request.url).pathname
+  const cacheTag = `_N_T_${key === '/index' ? '/' : encodeURI(key)}`
+  console.log('setCacheTagsHeaders', 'netlify-cache-tag', key)
+  headers.set('netlify-cache-tag', cacheTag)
```
pieh, line 303: deriving cache tag from request pathname is not reliable

- there could be rewrites ( https://nextjs.org/docs/app/api-reference/config/next-config-js/rewrites )
- json responses for Pages Router would be something like `_next/data/build_id/<route>.json`, so would not share cache tags with html routes
- there could be trailingSlash option enabled ( https://nextjs.org/docs/app/api-reference/config/next-config-js/trailingSlash )

(just to name a few that immediately comes to mind - there likely is much more)

Instead we should figure out how to capture those more reliably in cache handler, so logic here is not really changed (but we can keep simplified breaking up of condition with early bail so it's more readable overall)

src/run/headers.ts
```diff
@@ -273,6 +274,7 @@ export const setCacheControlHeaders = (
     ['GET', 'HEAD'].includes(request.method) &&
     !headers.has('cdn-cache-control') &&
     !headers.has('netlify-cdn-cache-control') &&
+    !new URL(request.url).pathname.startsWith('/api/') &&
```
pieh, line 277: we probably should not do this here and instead have something more proper for the API responses being potentially cached as it might impact not just pages router as this is trading one problem for another

It's fine to be there temporary while working on this PR, but should be removed before merging

src/run/handlers/cache.cts
```diff
@@ -393,7 +394,7 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
 
       await this.cacheStore.set(key, { lastModified, value }, 'blobStore.set')
 
-      if (data?.kind === 'PAGE' || data?.kind === 'PAGES') {
+      if (!data || data?.kind === 'PAGE' || data?.kind === 'PAGES') {
```
pieh, line 397: This might be too broad, we really should try to figure out exact key that is being invalidated as a result of calling `res.revalidate` and use that as condition instead of inspecting `data` at all.

We also can't just store argument passed to `res.revalidate` as-is, because Next.js might do some massaging to it so the argument might not match cache key later (somewhat similar to why we can't use request pathname to generate cache tags reliably).

This is quite tricky and I don't think we do have good tools today to figure this out correctly, but we will need to find some to do it

mrstork: Updated in 4cf7b2f400a31d1a067ca8e129fae04106cf3070

src/run/headers.ts
mrstork: Removed in ca7ca4e0fc556b2bf5919a42e42abb4a5d16d104

pieh (approved):

PR #2816, opened 2025-04-04
fix: add debug information around potential html/rsc response mismatches

pieh:
## Description

This is to help in debugging potential cases of unexpected rsc or html responses for app pages


orinokai (approved):
looks good!

PR #2819, opened 2025-04-07
fix: handle case of zero-length cacheable route handler responses

pieh:
## Description

Size calculcation for in-memory store in some cases was returning 0 (in particular known case is cacheable route handler with 0-length body).

This PR does few things:
 - ensure there is base blob size added to all known blob types so it should never return 0
 - add more checks to ensure fallback case also returns positive number
 - make crashes related to in-memory cache failures not fatal - this is meant to be optimization, but it should never cause handler to crash and lead to 500 errors. I did add warning reporting so we can spot those cases and adjust handling to get on happy path

## Tests

Added known case that is causing errors with currently released version

https://linear.app/netlify/issue/FRB-1721/sizecalculation-return-invalid-expect-positive-integer

### DIFF
--- a/src/run/storage/request-scoped-in-memory-cache.cts
+++ b/src/run/storage/request-scoped-in-memory-cache.cts
@@ -65,7 +65,6 @@ const estimateBlobKnownTypeSize = (
 }
 
 const estimateBlobSize = (valueToStore: BlobType | null | Promise<unknown>): PositiveNumber => {
-  let knownTypeFailed = false
   let estimatedKnownTypeSize: number | undefined
   let estimateBlobKnownTypeSizeError: unknown
   try {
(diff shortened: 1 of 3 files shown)


src/run/storage/request-scoped-in-memory-cache.cts
```diff
@@ -25,44 +25,71 @@ export function setInMemoryCacheMaxSizeFromNextConfig(size: unknown) {
...
+    )
   }
-  let knownKindFailed = false
+  if (valueToStore.value?.kind === 'ROUTE' || valueToStore.value?.kind === 'APP_ROUTE') {
+    return BASE_BLOB_SIZE + valueToStore.value.body.length
+  }
+}
+
+const estimateBlobSize = (valueToStore: BlobType | null | Promise<unknown>): PositiveNumber => {
+  let knownTypeFailed = false
```
orinokai, line 68: Technically, you could omit this var now that you're using `estimateBlobKnownTypeSizeError`. It's only used in the warning message below, which could be replaced by checking if `estimateBlobKnownTypeSizeError` is not `undefined`, but it's not a blocker, just something to consider next time you're touching the code here.

orinokai (dismissed):

src/run/storage/request-scoped-in-memory-cache.cts
pieh: I'll adjust now - it's unlikely this will be done as follow up

pieh: https://github.com/opennextjs/opennextjs-netlify/pull/2819/commits/f1ab6281390205b32b23960a79306dc33cb2d5ad dropped that extra unneded variable. Also added calculated size to warning just for it to have a bit more information

orinokai (approved):

PR #2847, opened 2025-04-15
fix: cookies set in middleware accessible during the same request

mrstork:
## Description

These changes make it possible to set a cookie in middleware and have that value be accessible at render time during the same request. This should also address a couple e2e tests failures we see in the report.

Part of FRB-1387
Fixes FRB-1748
Fixes https://github.com/opennextjs/opennextjs-netlify/issues/2669

### Tests

Used variants of the following setup:
```
// page.tsx
export default async function Page() {
  const myCookies = await cookies();

  if (!myCookies.has("nextjs-cookie")) {
    throw new Error("Cookie not found");
  }

  return (
    <div>
      <h1>Page</h1>
      <p>Generated by create next app</p>
      <p>Cookie: {myCookies.get("nextjs-cookie")?.value}</p>
    </div>
  );
}
```

```
// middleware.ts
export async function middleware(request: NextRequest) {
  const nextRes = NextResponse.next();

  nextRes.cookies.set("nextjs-cookie", "Hello from the cookie", {
    maxAge: 60 * 60 * 24 * 7,
  });

  return nextRes;
}
```

### DIFF
--- a/edge-runtime/lib/middleware.ts
+++ b/edge-runtime/lib/middleware.ts
@@ -72,7 +72,9 @@ export function mergeMiddlewareCookies(middlewareResponse: Response, lambdaReque
     const regex = new RegExp(/,(?!\s)/)
 
     middlewareCookies.split(regex).forEach((entry) => {
-      const [cookie] = entry.split(';')
+      // Extra directives within a cookie are joined on separated by "; "
+      // See: https://github.com/vercel/next.js/blob/0edb1123066a010eff2aac274f948ca2c6e85c0f/packages/next/src/compiled/%40edge-runtime/cookies/index.js#L32-L47
+      const [cookie] = entry.split('; ')
       const [name, value] = cookie.split('=')
       mergedCookies[name] = value
     })
--- a/edge-runtime/lib/middleware.ts
+++ b/edge-runtime/lib/middleware.ts
@@ -1,6 +1,7 @@
 import type { Context } from '@netlify/edge-functions'
 
 import type { ElementHandlers } from '../vendor/deno.land/x/htmlrewriter@v1.0.0/src/index.ts'
+import { getCookies } from '../vendor/deno.land/std@0.175.0/http/cookie.ts'
 
 type NextDataTransform = <T>(data: T) => T
 
@@ -57,4 +58,20 @@ export const addMiddlewareHeaders = async (
     }
   })
   return response
+}
+
+export function mergeMiddlewareCookies(middlewareResponse: Response, request: Request) {
+  let mergedCookies = getCookies(request.headers)
+  const middlewareCookies = middlewareResponse.headers.get('x-middleware-set-cookie') || ''
+  const regex = new RegExp(/,(?!\s)/) // commas that are not followed by whitespace
+
+  middlewareCookies.split(regex).forEach((entry) => {
+    const [cookie] = entry.split(';')
+    const [name, value] = cookie.split('=')
+    mergedCookies[name] = value
+  })
+
+  return Object.entries(mergedCookies)
+    .map((kv) => kv.join('='))
+    .join('; ')
 }
--- a/edge-runtime/lib/middleware.ts
+++ b/edge-runtime/lib/middleware.ts
@@ -60,16 +60,20 @@ export const addMiddlewareHeaders = async (
   return response
 }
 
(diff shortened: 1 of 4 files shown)


edge-runtime/lib/middleware.ts
```diff
@@ -58,3 +59,23 @@ export const addMiddlewareHeaders = async (
   })
   return response
 }
+
+// This serves the same purpose as the mergeMiddlewareCookies in Next.js but has been customized to our domain
+// See: https://github.com/vercel/next.js/blob/6e4495f8430eab33b12cd11dffdd8e27eee6e0cf/packages/next/src/server/async-storage/request-store.ts#L78-L105
+export function mergeMiddlewareCookies(middlewareResponse: Response, lambdaRequest: Request) {
+  let mergedCookies = getCookies(lambdaRequest.headers)
+  const middlewareCookies = middlewareResponse.headers.get('x-middleware-set-cookie')
+  const regex = new RegExp(/,(?!\s)/) // commas that are not followed by whitespace
```
pieh, line 68: is this relying on some implementation detail of how multiple set-cookie are implemented in Next.js?

because this will not work if I add test like this:
```
  await t.step('should support multiple cookies being set in middleware (version 2)', async () => {
    const request = new Request('https://www.test-url.com')
    const response = new Response()
    const headersHelper = new Headers()
    headersHelper.append(
      'set-cookie',
      `oatmeal=raisin; Path=/; Expires=Wed, 23 Apr 2025 13:37:43 GMT; Max-Age=604800`,
    )
    headersHelper.append(
      'set-cookie',
      `peanut=butter; Path=/; Expires=Wed, 23 Apr 2025 13:37:43 GMT; Max-Age=604800`,
    )
    response.headers.set(MIDDLEWARE_HEADER, headersHelper.get('set-cookie')!)
    const result = mergeMiddlewareCookies(response, request)
    assertEquals(result, 'oatmeal=raisin; peanut=butter')
  })
```

because constructing `set-cookie` this way would produce cookies seperated by `, ` which will make this pattern not split on it and result in single cookie being set.

I think it would be probably better to just vendor https://github.com/vercel/next.js/blob/da31ac70afbaf4ddd620ca06e5dbd55ce2d17255/packages/next/src/server/web/utils.ts#L34-L108 for cookie splitting - even if this is not great - at least this is how Next.js handles it, so we would be inline with their handling

And test case added for multiple cookies with `, ` seperated (I would suggest hardcoding set-cookie string in the test - almost copy of "should support multiple cookies being set in middleware" one that you already add - just have space after cookie seperator comma (I used programmatic things in my code above mostly to show this being valid seperator, but we probably should not rely on implementation detail like that that could change in the test) and/or maybe one with multiple cookies with `expires` that does contain comma

pieh (changes requested):

edge-runtime/lib/response.ts
```diff
@@ -197,9 +203,9 @@ export const buildResponse = async ({
       logger.withFields({ rewrite_url: rewrite }).debug('Rewrite url is same as original url')
       return
     }
-    res.headers.set('x-middleware-rewrite', relativeUrl)
+    edgeResponse.headers.set('x-middleware-rewrite', relativeUrl)
     request.headers.set('x-middleware-rewrite', target)
-    return addMiddlewareHeaders(context.rewrite(target), res)
+    return addMiddlewareHeaders(context.rewrite(target), edgeResponse)
```
pieh, line 208: I'm not sure, but possibly here we would also need to handle middleware cookies as well? next.js test fixture doesn't seem to have case for setting cookies for rewrites, so maybe we should not touch it in this PR - just flagging it (so this comment act as potential paper trail in case there was some future report about cookies not being handled for rewrites)

mrstork: Recreated the bug quite easily with the following middleware, so decided to fix it here
```
const nextRes = NextResponse.rewrite(url)
nextRes.cookies.set("nextjs-cookie", "rewrite");
return nextRes;
```

pieh (approved):

edge-runtime/lib/middleware.ts
mrstork: Left comments in code with explanations/links to implementation details around the `x-middleware-set-cookie` header

PR #2851, opened 2025-04-16
test: wait for background work to finish in integration tests

pieh:
## Description

Since `waitUntil` was shipped, our integration test's `invokeFunction` was not awaiting background work which resulted in some flakiness - specifically visible on Windows runners. This restore previous behavior

https://linear.app/netlify/issue/FRB-1714/flaky-test-purge-api-is-not-used-when-unstable-cache-cache-entry-gets

### DIFF
--- a/tests/utils/fixture.ts
+++ b/tests/utils/fixture.ts
@@ -396,6 +396,12 @@ export async function invokeFunction(
       process.env[key] = environment[key]
     })
 
+    let resolveInvocation, rejectInvocation
+    const invocationPromise = new Promise((resolve, reject) => {
+      resolveInvocation = resolve
+      rejectInvocation = reject
+    })
+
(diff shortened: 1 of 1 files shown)


tests/utils/fixture.ts
```diff
@@ -405,8 +411,20 @@ export async function invokeFunction(
...
+        // before resolving the promise to allow background work to finish
+        if (error) {
+          rejectInvocation(error)
+        } else {
+          resolveInvocation()
+        }
+      },
     })) as LambdaResponse
 
+    await invocationPromise
```
pieh, line 426: possibly instead of awaiting here we should return this promise, but before https://github.com/netlify/serverless-functions-api/pull/358 landed we were basically waiting for stream to close here and we were blocking stream closing to ensure background work

PR #2862, opened 2025-04-22
feat: support 'use cache'

pieh:
## Description

This is work in progress to support upcoming [`'use cache'` directive](https://nextjs.org/docs/app/api-reference/directives/use-cache)

Notes:
The primary need for custom Netlify cache handler implementation (the new one, that is used by `'use cache'` directive) is that default implementation is in-memory only. While it seems like at least `default` handler for `use cache` doesn't need entries to be actually persisted - it still need to support tags (see [`cacheTag`](https://nextjs.org/docs/app/api-reference/functions/cacheTag)) properly and in serverless environment it means that we can't use in-memory only tag manifest like the default implementation does and will need to check with some persisted source of truth.

Review notes:
This PR contains also changes from following refactor PRs that it depends on.
- https://github.com/opennextjs/opennextjs-netlify/pull/2871
- https://github.com/opennextjs/opennextjs-netlify/pull/2872

To check just "use cache" specific changes without polluting diff with refactors use this [compare view](https://github.com/opennextjs/opennextjs-netlify/compare/tmp-just-use-cache-base...michalpiechowiak/frb-1405-next15-use-cache-not-persisted?w=1)

## Tests

Primarily integration tests are being added for various `use cache` scenarios:
 - Is response cacheable on cdn? (static/dynamic page)
 - What `use cache` is used on (data fetching function, non-page react component, page react component, whole page module)

And test check behavior when executing against same serverless function instance, different serverless function instances (default cache is in-memory only, so to test behaviors it forces use of integration tests as you can't ensure hitting same or different serverless function in e2e), checking behaviors after tag invalidation (also depending on wether same serverless instance is used or not and that it should work the same in both cases because tags are shared, while cache entries themselves are not) and finally checking that expiry behavior works.

Tests failing on their own in https://github.com/opennextjs/opennextjs-netlify/actions/runs/14674525410/job/41267496397 (comment about most of tests passing - lot of test cases there do NOT test primary difference between Next.js's default cache handler and Netlify's implementation - only test related to revalidating tags on different lambda that is checked later - this is somewhat intentional because the whole cache handler is implemented here and doesn't reuse Next.js code - this is intentional to make sure any potentially future changes to default cache handler don't break any assumptions - and because we don't actually share code I added test cases just for overall cache handler behavior to make sure it works as expected)

https://linear.app/netlify/issue/FRB-1405/[next15]-use-cache-not-persisted

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "msw": "^2.0.7",
     "netlify-cli": "^20.1.1",
     "next": "^15.0.0-canary.28",
+    "next-with-cache-handler-v2": "npm:next@15.3.0-canary.13",
     "os": "^0.1.2",
     "outdent": "^0.8.0",
     "p-limit": "^5.0.0",
@@ -89,13 +90,18 @@
     "uuid": "^10.0.0",
     "vitest": "^3.0.0"
   },
+  "overrides": {
+    "react": "19.0.0-rc.0",
+    "react-dom": "19.0.0-rc.0"
+  },
   "clean-package": {
     "indent": 2,
     "remove": [
       "clean-package",
       "dependencies",
       "devDependencies",
-      "scripts"
+      "scripts",
+      "overrides"
     ]
   }
 }
--- a/src/run/handlers/tags-handler.cts
+++ b/src/run/handlers/tags-handler.cts
@@ -26,6 +26,27 @@ async function getTagRevalidatedAt(
 }
 
 /**
+ * Get the most recent revalidation timestamp for a list of tags
+ */
+export async function getMostRecentTagRevalidationTimestamp(tags: string[]) {
+  if (tags.length === 0) {
+    return 0
+  }
+
+  const cacheStore = getMemoizedKeyValueStoreBackedByRegionalBlobStore({ consistency: 'strong' })
+
+  const timestampsOrNulls = await Promise.all(
+    tags.map((tag) => getTagRevalidatedAt(tag, cacheStore)),
+  )
+
+  const timestamps = timestampsOrNulls.filter((timestamp) => timestamp !== null)
+  if (timestamps.length === 0) {
+    return 0
+  }
+  return Math.max(...timestamps)
+}
+
+/**
  * Check if any of the tags were invalidated since the given timestamp
  */
 export function isAnyTagStale(tags: string[], timestamp: number): Promise<boolean> {
--- a/src/run/handlers/use-cache-handler.ts
+++ b/src/run/handlers/use-cache-handler.ts
@@ -1,0 +1,270 @@
+import { Buffer } from 'node:buffer'
+
+import { LRUCache } from 'lru-cache'
+import type {
+  CacheEntry,
+  // only supporting latest variant (https://github.com/vercel/next.js/pull/76687)
+  // first released in v15.3.0-canary.13
+  CacheHandlerV2 as CacheHandler,
+} from 'next-with-cache-handler-v2/dist/server/lib/cache-handlers/types.js'
+
+import { getLogger } from './request-context.cjs'
+import {
+  getMostRecentTagRevalidationTimestamp,
+  isAnyTagStale,
+  markTagsAsStaleAndPurgeEdgeCache,
+} from './tags-handler.cjs'
+import { getTracer } from './tracer.cjs'
+
+// copied from default implementation
+// packages/next/src/server/lib/cache-handlers/default.ts
+type PrivateCacheEntry = {
+  entry: CacheEntry
+
+  // For the default cache we store errored cache
+  // entries and allow them to be used up to 3 times
+  // after that we want to dispose it and try for fresh
+
+  // If an entry is errored we return no entry
+  // three times so that we retry hitting origin (MISS)
+  // and then if it still fails to set after the third we
+  // return the errored content and use expiration of
+  // Math.min(30, entry.expiration)
+
+  isErrored: boolean // pieh: this doesn't seem to be actually used in the default implementation
+  errorRetryCount: number // pieh: this doesn't seem to be actually used in the default implementation
+
+  // compute size on set since we need to read size
+  // of the ReadableStream for LRU evicting
+  size: number
+}
+
+type CacheHandleLRUCache = LRUCache<string, PrivateCacheEntry>
+type PendingSets = Map<string, Promise<void>>
+
+const LRU_CACHE_GLOBAL_KEY = Symbol.for('nf-use-cache-handler-lru-cache')
+const PENDING_SETS_GLOBAL_KEY = Symbol.for('nf-use-cache-handler-pending-sets')
+const cacheHandlersSymbol = Symbol.for('@next/cache-handlers')
+const extendedGlobalThis = globalThis as typeof globalThis & {
+  // Used by Next Runtime to ensure we have single instance of
+  //  - LRUCache
+  //  - pending sets
+  // even if this module gets copied multiple times
+  [LRU_CACHE_GLOBAL_KEY]?: CacheHandleLRUCache
+  [PENDING_SETS_GLOBAL_KEY]?: PendingSets
+
+  // Used by Next.js to provide implementation of cache handlers
+  [cacheHandlersSymbol]?: {
+    RemoteCache?: CacheHandler
+    DefaultCache?: CacheHandler
+  }
+}
+
+function getLRUCache(): CacheHandleLRUCache {
+  if (extendedGlobalThis[LRU_CACHE_GLOBAL_KEY]) {
+    return extendedGlobalThis[LRU_CACHE_GLOBAL_KEY]
+  }
+
+  const lruCache = new LRUCache<string, PrivateCacheEntry>({
+    max: 1000,
+    maxSize: 50 * 1024 * 1024, // same as hardcoded default in Next.js
+    sizeCalculation: (value) => value.size,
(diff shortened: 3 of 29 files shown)


package.json
```diff
@@ -77,6 +77,7 @@
     "msw": "^2.0.7",
     "netlify-cli": "^20.1.1",
     "next": "^15.0.0-canary.28",
+    "next-next": "npm:next@^15.3.1-canary.7",
```
pieh, line 80: This is to get types for new `use cache` handler. I didn't want to touch `next` version we already use to not increase scope of this PR, as there is a lot of `next` types issues to solve in general.

Maybe this alias should be named better to indicate it's for `use cache` specifically (at least as of now)

src/build/functions/server.ts
```diff
@@ -129,6 +133,25 @@ const writeHandlerFile = async (ctx: PluginContext) => {
...
+      // see https://github.com/vercel/next.js/pull/76687 first released in v15.3.0-canary.13
+      enableUseCacheHandler: ctx.nextVersion
+        ? satisfies(ctx.nextVersion, '>=15.3.0-canary.13', {
+            includePrerelease: true,
+          })
+        : false,
+    } satisfies RunConfig),
+    'utf-8',
+  )
+}
```
pieh, line 153: this is to make `run-config.json` file to also contain next-runtime toggles and not just Next.js config to allow us to conditionally enable some setups - in this case conditionally setup our `use cache` handler depending on version because using implementation in this PR that does not implement V1 variant is resulting in crashes due to some methods from V1 interface not being implemented

src/run/handlers/tags-handler.cts
```diff
@@ -25,6 +25,27 @@ async function getTagRevalidatedAt(
...
+    tags.map((tag) => getTagRevalidatedAt(tag, cacheStore)),
+  )
+
+  const timestamps = timestampsOrNulls.filter((timestamp) => timestamp !== null)
+  if (timestamps.length === 0) {
+    return 0
+  }
+  return Math.max(...timestamps)
+}
+
```
pieh, line 48: To support interface of `use cache` handler we need method that will return most recent tag expiration timestamp .. or `0` if none of tags were ever revalidated (as opposed to just checking if tag is stale that is used in response cache handler)

src/run/handlers/use-cache-handler.ts
```diff
@@ -0,0 +1,270 @@
...
+  // after that we want to dispose it and try for fresh
+
+  // If an entry is errored we return no entry
+  // three times so that we retry hitting origin (MISS)
+  // and then if it still fails to set after the third we
+  // return the errored content and use expiration of
+  // Math.min(30, entry.expiration)
+
+  isErrored: boolean // pieh: this doesn't seem to be actually used in the default implementation
+  errorRetryCount: number // pieh: this doesn't seem to be actually used in the default implementation
```
pieh, line 35: Those are only ever set, but never actually read, so possibly those should be removed. This interface is copied as-is from Next.js default implementation, so I did want to include this intentionally here for awareness

src/run/handlers/use-cache-handler.ts
```diff
@@ -0,0 +1,270 @@
...
+
+  const pendingSets = new Map()
+  extendedGlobalThis[PENDING_SETS_GLOBAL_KEY] = pendingSets
+  return pendingSets
+}
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const tmpResolvePendingBeforeCreatingAPromise = () => {}
+
+export const NetlifyDefaultUseCacheHandler = {
```
pieh, line 92: This is mostly copy & paste from Next.js default cache handler except for tags handling

src/run/handlers/use-cache-handler.ts
```diff
@@ -0,0 +1,270 @@
...
+    return extendedGlobalThis[PENDING_SETS_GLOBAL_KEY]
+  }
+
+  const pendingSets = new Map()
+  extendedGlobalThis[PENDING_SETS_GLOBAL_KEY] = pendingSets
+  return pendingSets
+}
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const tmpResolvePendingBeforeCreatingAPromise = () => {}
```
mrstork, line 90: This is our temp variable, or was this also copy pasted?

src/run/handlers/use-cache-handler.ts
```diff
@@ -0,0 +1,270 @@
...
+          })
+          return undefined
+        }
+
+        const { entry } = privateEntry
+        const ttl = (entry.timestamp + entry.revalidate * 1000 - Date.now()) / 1000
+        if (ttl < 0) {
+          // In-memory caches should expire after revalidate time because it is
+          // unlikely that a new entry will be able to be used before it is dropped
+          // from the cache.
```
mrstork, line 123: How do we handle the "stale while revalidate" case?

mrstork, line 123: How do we handle the "stale while revalidate" case?

src/run/handlers/use-cache-handler.ts
```diff
@@ -0,0 +1,270 @@
...
+            .withFields({ cacheKey, ttl, status: 'STALE' })
+            .debug(`[NetlifyDefaultUseCacheHandler] get result`)
+          span.setAttributes({
+            cacheStatus: 'expired, discarded',
+            ttl,
+          })
+          return undefined
+        }
+
+        if (await isAnyTagStale(entry.tags, entry.timestamp)) {
```
mrstork, line 134: Learning - If any tag is stale, then the entry is stale? So if I have a long chain of tags, if for example `_n_t_/layout` becomes stale then every page using that tag will be stale (since it's often the top level tag for an entry)?

src/run/handlers/use-cache-handler.ts
pieh: it was originally same as in  https://github.com/vercel/next.js/blob/0eb78fd629858858a5a31591b3b3de294762f870/packages/next/src/server/lib/cache-handlers/default.ts#L101, just our lintings rule didn't like defining this temporary no-op function in the same place:

image

and to me, the rule mede sense so I did move it to outer scope

the purpose of tmp function there that we re-assign in the `new Promise` callback in the first place is to make typescript happy, because otherwise `resolvePending` is either nullable or we have to force a type, so none of the options are really great and I just thought the way it currently is was the least worst option

src/run/handlers/use-cache-handler.ts
pieh: https://github.com/vercel/next.js/blob/0eb78fd629858858a5a31591b3b3de294762f870/packages/next/src/server/lib/cache-handlers/default.ts#L65-L75

with the default cache handler there is not

src/run/handlers/use-cache-handler.ts
pieh: Yes, that's correct - so if you mark layout as stale - then all pages that use that layout are stale and should be regenerated on next opportunity

src/run/handlers/use-cache-handler.ts
pieh: I should clear this up - https://github.com/opennextjs/opennextjs-netlify/blob/11c7164f9a00109516df870caceb3576d352a507/src/run/handlers/use-cache-handler.ts#L19-L20 refers to most of this module (not just immediate surrounding) - but I should probably do bigger explainer there and do some separation from code immediately after this comment (and also link to some specific version of it with github link with SHA so it's clear what version this was originally modeled after)

tests/integration/use-cache.test.ts
```diff
@@ -0,0 +1,429 @@
...
+      useCacheLocationPathSegment,
+      useCacheTagPrefix,
+    } of [
+      {
+        useCacheLocationLabel: "'use cache' in data fetching function",
+        useCacheLocationPathSegment: 'use-cache-data',
+        useCacheTagPrefix: 'data',
+        expectedCachingBehaviorWhenUseCacheRegenerates: {
+          // getData function has 'use cache' so it should report same generation time, everything else is dynamically regenerated on each request
+          getDataTimeShouldShouldBeEqual: true,
```
mrstork, line 180: Don't believe the double `Should` is intentional

src/build/functions/server.ts
pieh: This was moved since initially commented on, but leaving comment up for visibility as same change is still in this PR, just in slightly different place (to reduce amount of changes it introduces)

src/run/handlers/use-cache-handler.ts
pieh: This was removed in https://github.com/opennextjs/opennextjs-netlify/pull/2862/commits/8a23ceaaccc81e993c207ba07c2159eeb95a6baf

src/run/handlers/use-cache-handler.ts
pieh: Hopefully added more clear description in https://github.com/opennextjs/opennextjs-netlify/pull/2862/commits/8a23ceaaccc81e993c207ba07c2159eeb95a6baf

tests/integration/use-cache.test.ts
pieh: fixed typo in https://github.com/opennextjs/opennextjs-netlify/pull/2862/commits/e18b9b8766ce3b73b3e824198308dc0e39dde7cf 🤞

mrstork (approved):

PR #2866, opened 2025-04-22
fix: enforce a single edge entrypoint for next edge handler

mrstork:
## Tests

1. TODO


src/build/functions/edge.ts
```diff
@@ -160,21 +160,25 @@ const getHandlerName = ({ name }: Pick<NextDefinition, 'name'>): string =>
 const buildHandlerDefinition = (
   ctx: PluginContext,
   { name, matchers, page }: NextDefinition,
-): Array<ManifestFunction> => {
-  const fun = getHandlerName({ name })
-  const funName = name.endsWith('middleware')
+): ManifestFunction & IntegrationsConfig => {
```
mrstork, line 163: See: https://github.com/netlify/edge-functions-bootstrap/pull/417

PR #2871, opened 2025-04-24
test: refactor serverless invocations so in-process and sandbox implementation use shared code and ability to run multiple invocations in same sandbox

pieh:
## Description

This splits off changes needed for https://github.com/opennextjs/opennextjs-netlify/pull/2862 to be more self contained and to make things easier to review in chunks. It refactors serverless invocation helper for integration tests - it unify handling between `invokeFunction` and `invokeSandboxedFunction` to reuse same core code of invoking function - it moves needed parts to common JS file (because sandboxed version running in child process can't use TS, like main test process can - but for preserving type checking equiavalent jsdocs and `// @ts-check` was added). It also adds ability to keep sandbox running to allow invoking it multiple times while it preserve it's state

Notes for review:
 - If full diff is too much - I tried to split it in somewhat reasonable commits that could be reviewed in sequence
 - I suggest hiding whitespace changes (especially when reviewing commit by commit) 

## Tests

This is changing integration test helper - all existing tests should pass

Part of FRB-1405


tests/utils/next-version-helpers.mjs
```diff
@@ -6,7 +6,7 @@ import fg from 'fast-glob'
 import { coerce, gt, gte, satisfies, valid } from 'semver'
 import { execaCommand } from 'execa'
 
-const FUTURE_NEXT_PATCH_VERSION = '14.999.0'
+const FUTURE_NEXT_PATCH_VERSION = '15.999.0'
```
pieh, line 9: unrelated to rest - it just impacts some conditional assertions depending on next versions when running tests locally and not setting `RESOLVED_NEXT_VERSION` env var (like CI does) - the idea of this const is to always use latest version conditionals when running locally as this is most likely what you'd test with

tests/utils/fixture.ts
```diff
@@ -508,48 +410,140 @@ export async function invokeEdgeFunction(
...
+ * Load function in child process and execute single invocation
+ */
+export async function invokeSandboxedFunction(
+  ctx: FixtureTestContext,
+  options: FunctionInvocationOptions = {},
+) {
+  const { invokeFunction, exit } = await loadSandboxedFunction(ctx, options)
+  const result = await invokeFunction(options)
+  exit()
+  return result
```
pieh, line 548: this ensures existing tests using `invokeSandboxedFunction` continue to work as they used to

tests/utils/lambda-helpers.mjs
```diff
@@ -0,0 +1,197 @@
...
+/**
+ * @param {FixtureTestContext} ctx
+ * @param {LoadFunctionOptions} options
+ */
+export async function loadFunction(ctx, { env } = {}) {
+  const restoreEnvironment = temporarilySetEnv(ctx, env)
+
+  const { handler } = await import(
+    'file:///' + join(ctx.functionDist, SERVER_HANDLER_NAME, '___netlify-entry-point.mjs')
+  )
```
pieh, line 121: note that this changed order versus current implementation so that env vars are set BEFORE importing serverless handler

this is because some of the imported code evaluate env vars at import time - so it's pretty helpful to set them up early (for example when using `NEXT_PRIVATE_DEBUG_CACHE`)

tests/utils/lambda-helpers.mjs
```diff
@@ -0,0 +1,194 @@
+// @ts-check
+
+// this is not TS file because it's used both directly inside test process
+// as well as child process that lacks TS on-the-fly transpilation
+
+import { join } from 'node:path'
+import { BLOB_TOKEN } from './constants.mjs'
+import { execute as untypedExecute } from 'lambda-local'
+
+const SERVER_HANDLER_NAME = '___netlify-server-handler'
```
pieh, line 10: this is copied from https://github.com/opennextjs/opennextjs-netlify/blob/e96cd1831d0a6b0f6429e08a77127a48467de56a/src/build/plugin-context.ts#L22 (for the serverless invocation in same process it was imported from there, but it was previously already copied for sandboxed invocation because it can't import TS, so not fully new that this is duplicated)

tests/utils/sandbox-child.mjs
```diff
@@ -1,109 +1,53 @@
...
+ * @type {import('./lambda-helpers.mjs').InvokeFunction | undefined}
+ */
+let invokeFunctionImpl
+
+process.on(
+  'message',
+  /**
+   * @param {any} msg
+   */
+  async (msg) => {
```
pieh, line 18: This wasn't typed before, but adding explicit note was needed after adding `// @ts-check` to get at least partial type checking here and is this is not used directly in tests I didn't want to spend too much time on making details of sandbox fully typed

tests/utils/sandbox-child.mjs
```diff
@@ -13,16 +18,29 @@ process.on(
   async (msg) => {
     if (msg?.action === 'exit') {
       process.exit(0)
+    } else if (msg?.action === 'loadFunction') {
+      const [ctx, options] = msg.args
+
+      invokeFunctionImpl = await loadFunction(ctx, options)
+
+      if (process.send) {
```
serhalp, line 26: when is this false? is this expected?

serhalp (approved):
LGTM. I didn't think we'd ever actually get around to consolidating these. Thanks! 🙌🏼

tests/utils/sandbox-child.mjs
pieh: typescript `@types/node` report it as nullable in general which forces some checking (due to added `// @ts-check` directive)

generally it won't exist in parent process or if there is no IPC bridge setup (this is to send IPC messages to parent process), but in our current usage it will actually always be defined

tests/utils/lambda-helpers.mjs
```diff
@@ -0,0 +1,181 @@
+// @ts-check
+
+// this is not TS file because it's used both directly inside test process
+// as well as child process that lacks TS on-the-fly transpilation
```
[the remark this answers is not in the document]
pieh: We could try, but ideally this is one that matches what vitest is using and I wasn't sure it was worth the effort given TS-ish jsdocs are possibility. I've done something like that in https://github.com/gatsbyjs/gatsby/pull/32120 (you can grep for `ts-register` there, tho that exact thing with `@babel/register` I wouldn't try to use here). But in gatsby case this child processes handling was part of core framework functionality so full type safety there was much more important, while here it might just result in our own test helper problems as we don't use this in actual next runtime

PR #2872, opened 2025-04-25
refactor: move tags handling from cache-handler module to dedicated tags-handler to allow for reuse

pieh:
## Description

This splits off changes needed for https://github.com/opennextjs/opennextjs-netlify/pull/2862 to be more self contained and to make things easier to review in chunks.

This moves tags handling to dedicated module instead of handling all of it in existing cache handler. On it's own this should increase seperation and readability as existing cache handler already is pretty complex and handle many concerns so this just moves one of those concerns out and also sets things up so tags handling can be reused for `use cache` handlers 

## Tests

Existing tests using tags and tag purges should still pass. There should be no functional changes happening.

Part of FRB-1405


src/run/handlers/tags-handler.cts
```diff
@@ -0,0 +1,146 @@
...
+ */
+async function lastTagRevalidationTimestamp(
+  tag: string,
+  cacheStore: MemoizedKeyValueStoreBackedByRegionalBlobStore,
+): Promise<number | null> {
+  const tagManifest = await cacheStore.get<TagManifest>(tag, 'tagManifest.get')
+  if (!tagManifest) {
+    return null
+  }
+  return tagManifest.revalidatedAt
```
mrstork, line 25: just dreaming - If we used `undefined` instead of `null` we could do `tagManifest?.revalidatedAt` and write less code

src/run/handlers/cache.cts
```diff
@@ -427,70 +423,17 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
         if (requestContext?.didPagesRouterOnDemandRevalidate) {
           // encode here to deal with non ASCII characters in the key
           const tag = `_N_T_${key === '/index' ? '/' : encodeURI(key)}`
-          const tags = tag.split(/,|%2c/gi).filter(Boolean)
-
-          if (tags.length === 0) {
-            return
-          }
 
           getLogger().debug(`Purging CDN cache for: [${tag}]`)
```
mrstork, line 427: If tags are empty we now log this line, so I think we might want to be moving this log line over to the new function as well

src/run/handlers/tags-handler.cts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+async function doRevalidateTag(tags: string[]): Promise<void> {
+  getLogger().withFields({ tags }).debug('NetlifyCacheHandler.revalidateTag')
+
+  if (tags.length === 0) {
+    return
+  }
+
+  const data: TagManifest = {
```
mrstork, line 111: Maybe `initialTagData`?

src/run/handlers/tags-handler.cts
```diff
@@ -0,0 +1,146 @@
...
+    revalidatedAt: Date.now(),
+  }
+
+  const cacheStore = getMemoizedKeyValueStoreBackedByRegionalBlobStore({ consistency: 'strong' })
```
mrstork, line 115: 👍

mrstork (approved):

src/run/handlers/tags-handler.cts
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Get timestamp of the last revalidation for a tag
+ */
+async function lastTagRevalidationTimestamp(
+  tag: string,
+  cacheStore: MemoizedKeyValueStoreBackedByRegionalBlobStore,
+): Promise<number | null> {
+  const tagManifest = await cacheStore.get<TagManifest>(tag, 'tagManifest.get')
+  if (!tagManifest) {
+    return null
```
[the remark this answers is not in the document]
pieh: pros & cons

returning a number may simplify types and possibly remove the need to do specific `null` checking, but I think it's actually better to return non-number here exactly to force consumer of this function to consider what should be done when there is no value so it forces to take a second how to handle scenario if there is no tag manifest as this should prevent bugs that are result of oversight if someone would assume there is always a value here

Next.js `use cache` handler wants to return `0` in cases like that - https://github.com/vercel/next.js/blob/071e31705f4451bdeb8ada4126a0da9c49d11761/packages/next/src/server/lib/cache-handlers/types.ts#L102-L108 so maybe we should follow this, but because this is used not just for new cache handler I'd prefer more verbose handling to ensure that consumers handling don't have potentially incompatible oversights

src/run/handlers/tags-handler.cts
pieh: I don't think producing less code lines is good goal - I think making the code the most readable is

src/run/handlers/cache.cts
pieh: yup, good catch - I should have moved it (plus it will be beneficial overall as future callers will also inherit this log so all edge cache purges should produce it)

src/run/handlers/tags-handler.cts
pieh: that's not just initial - this is the only way we are setting when tag was revalidated last - so we do this on first tag invalidation but also on every subsequent one (also this was just copied as is)

maybe it should just be named `tagManifest` instead to be very verbose what it represents because `data` is just ... not telling what it is exactly

mrstork: I know it was simply copied over, but I was searching for a slightly better name. Happy with a rename to `tagManifest` as well.

mrstork (approved):

mrstork (approved):

PR #2881, opened 2025-04-29
fix: add a fallback for loadManifest import path

mrstork:
## Description

Resolves failures in next@canary integration tests

## Tests

Ran the code changes with `test all versions` tag to ensure failing canary integration tests are now resolved

Fixes FRB-1770

### DIFF
--- a/src/run/handlers/cache.cts
+++ b/src/run/handlers/cache.cts
@@ -176,7 +176,13 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
         typeof cacheControl !== 'undefined')
     ) {
       try {
-        const { loadManifest } = await import('next/dist/server/load-manifest.js')
+        let loadManifest
+        try {
+          // Starting in 15.4.0-canary.10 loadManifest was relocated (https://github.com/vercel/next.js/pull/78358)
+          ({ loadManifest } = await import('next/dist/server/load-manifest.external.js'))
+        } catch {
+          ({ loadManifest } = await import('next/dist/server/load-manifest.js'))
+        }
         const prerenderManifest = loadManifest(
           join(this.options.serverDistDir, '..', 'prerender-manifest.json'),
         ) as PrerenderManifest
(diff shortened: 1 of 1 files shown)


src/run/handlers/cache.cts
```diff
@@ -176,7 +176,15 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
...
-        const { loadManifest } = await import('next/dist/server/load-manifest.js')
+        let loadManifest
+        try {
+          // @ts-expect-error Starting in 15.4.0-canary.10 loadManifest was relocated (https://github.com/vercel/next.js/pull/78358)
+          // eslint-disable-next-line @typescript-eslint/no-extra-semi, n/no-missing-import, import/no-unresolved
+          ;({ loadManifest } = await import('next/dist/server/load-manifest.external.js'))
+        } catch {
+          // eslint-disable-next-line @typescript-eslint/no-extra-semi
+          ;({ loadManifest } = await import('next/dist/server/load-manifest.js'))
+        }
```
pieh, line 187: now that there 2 potential locations - it would be good to memoize `loadManifest` after first successful import, because successful imports put modules in cache so next time we call `await import` for same module is just cache lookup so previous code only paid price of trying to load a module once at first try

but if `next/dist/server/load-manifest.external.js` would not exist for used version - each time we call this function it would be attempted to load it again and then falling back to actually working import (until Next.js change again heh)

pieh (dismissed):

src/run/handlers/cache.cts
```diff
@@ -165,6 +166,30 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
...
+    } catch {
+      const { loadManifest } = await import('next/dist/server/load-manifest.js')
+      this.prerenderManifest = Promise.resolve(
+        loadManifest(prerenderManifestPath) as PrerenderManifest,
+      )
+    }
+
+    return this.prerenderManifest
+  }
+
```
pieh, line 192: This would need to not be part of class instance.

I think this would also be overall good because `loadManifest` call would no longer happen on each call to `injectEntryToPrerenderManifest` (`loadManifest` seems to use memoization itself https://github.com/vercel/next.js/blob/c3429682aa910eb2b5ddd1e761c8ec8cfaa4bb04/packages/next/src/server/load-manifest.external.ts#L40-L43 , but we then no longer would need to rely on this if it ever change to avoid redoing same work for subsequent requests)

pieh (approved):

PR #2891, opened 2025-05-01
fix: ensure internal `x-middleware-set-cookie` header is not passed on to lambda

mrstork:
## Description

Middleware cookies depend on the `x-middleware-set-cookie header` to work, however Next does not expect this header to be passed on after the middleware. This change ensures the value is not set when sending the request to the lambda.

## Tests

```
./run-local-test.sh test/e2e/app-dir/app-middleware

PASS default test/e2e/app-dir/app-middleware/app-middleware.test.ts (234.184 s)
  app-dir with middleware
    ✓ should filter correctly after middleware rewrite (3697 ms)
    ✓ retains a link response header from the middleware (335 ms)
    ✓ should be possible to modify cookies & read them in an RSC in a single request (1121 ms)
    ✓ should respect cookie options of merged middleware cookies (847 ms)
    ✓ should omit internal headers for middleware cookies (249 ms)
    ✓ should ignore x-middleware-set-cookie as a request header (175 ms)
    ✓ should be possible to read cookies that are set during the middleware handling of a server action (754 ms)
    Mutate request headers for Serverless Functions
      ✓ Adds new headers (238 ms)
      ✓ Deletes headers (167 ms)
      ✓ Updates headers (139 ms)
      ✓ Supports draft mode (116 ms)
    Mutate request headers for Edge Functions
      ✓ Adds new headers (341 ms)
      ✓ Deletes headers (151 ms)
      ✓ Updates headers (118 ms)
      ✓ Supports draft mode (117 ms)
    Mutate request headers for next/headers
      ✓ Adds new headers (200 ms)
      ✓ Deletes headers (182 ms)
      ✓ Updates headers (177 ms)
      ✓ Supports draft mode (136 ms)
  app dir - middleware without pages dir
    ✓ Updates headers (445 ms)
  app dir - middleware with middleware in src dir
    ✓ works without crashing when using RequestStore (346 ms)
```

Focus is on `should omit internal headers for middleware cookies` test

Fixes FRB-1774


tests/test-config.json
```diff
@@ -377,6 +377,15 @@
       "tests": [
         "router autoscrolling on navigation bugs Should apply scroll when loading.js is used"
       ]
+    },
+    {
+      "file": "test/e2e/app-dir/app-middleware/app-middleware.test.ts",
+      "reason": "Relies on access to environment variables set on the edge",
```
mrstork, line 383: See: https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/get-edge-preview-props.ts

edge-runtime/lib/middleware.ts
```diff
@@ -67,6 +67,10 @@ export function mergeMiddlewareCookies(middlewareResponse: Response, lambdaReque
   const middlewareCookies = middlewareResponse.headers.get('x-middleware-set-cookie')
 
   if (middlewareCookies) {
+    // Next expects internal headers to be omitted when cookies are set by the middleware
+    // See: https://github.com/vercel/next.js/blob/005db43079c7b59fd8c2594e8362761dc4cb3211/test/e2e/app-dir/app-middleware/app-middleware.test.ts#L197-L207
+    middlewareResponse.headers.delete('x-middleware-set-cookie')
```
pieh, line 72: Does final response reaching client (browser/curl etc) would have actual `set-cookie` with this removal? from what I can see `mergeMiddlewareCookies` only sets `cookie` header:
https://github.com/opennextjs/opennextjs-netlify/blob/7795157ec4ab1270acb063abb28ac6fc5ad423a2/edge-runtime/lib/response.ts#L245-L248

so I wonder if this change as-is wouldn't result in clients no longer actually receiving `set-cookie` headers?

tests/test-config.json
pieh: I think there is something we can do to actually make it work and possibly this would also handle some other cases that rely on env vars like that:

This sets some "PreviewProps" at build time with a bit of random values
https://github.com/vercel/next.js/blob/f3f0a03e820eecaa15a8067791e930f58d4a03c4/packages/next/src/build/index.ts#L1108-L1113

And this looks like it tries to set them in built middleware
https://github.com/vercel/next.js/blob/f3f0a03e820eecaa15a8067791e930f58d4a03c4/packages/next/src/build/webpack-config.ts#L2009-L2022

Then those env vars seems to be defined in `middleware-manifest.json` - for example like this:
```
{
  "version": 3,
  "middleware": {
    "/": {
      "files": [
        "server/edge-runtime-webpack.js",
        "server/middleware-f3956634.js",
        "server/middleware-2971c211.js",
        "server/middleware-7fce854e.js",
        "server/middleware-89adb3be.js",
        "server/middleware-7874eb72.js",
        "server/middleware-27161c75.js",
        "server/middleware-d1744647.js",
        "server/middleware-16b16fa0.js",
        "server/middleware-8bfe7224.js",
        "server/middleware-b9f70e77.js",
        "server/middleware-377fed06.js"
      ],
      "name": "middleware",
      "page": "/",
      "matchers": [
        {
          "regexp": "^(?:\\/(_next\\/data\\/[^/]{1,}))?\\/test(?:\\/((?:[^\\/#\\?]+?)(?:\\/(?:[^\\/#\\?]+?))*))?(\\.json)?[\\/#\\?]?$",
          "originalSource": "/test/:path*"
        }
      ],
      "wasm": [],
      "assets": [],
      "env": {
        // we have env vars to set, but we don't set them
        "__NEXT_BUILD_ID": "jz-SjGjUVmhXJqhfMaRsg",
        "NEXT_SERVER_ACTIONS_ENCRYPTION_KEY": "TNX5YdWxrvsdPSUR7O84uW1PWttdF3O8O2eWT4G1VpU=",
        "__NEXT_PREVIEW_MODE_ID": "94e82426a39aa9b691f468ba3e0a61a4",
        "__NEXT_PREVIEW_MODE_ENCRYPTION_KEY": "d9975864b9708a047ca1cc8d1baf3c86f98ec4cecad4c98959c8e8ade0ed9574",
        "__NEXT_PREVIEW_MODE_SIGNING_KEY": "152fc2b4ce081946d22431113018c8be384692e2f90af6b31fa61769bf1e4834"
      }
    }
  },
  "functions": {},
  "sortedMiddleware": [
    "/"
  ]
}
```

And the part that we would need to do is to ensure that we set those `process.env.X` values from this manifest somewhere in https://github.com/opennextjs/opennextjs-netlify/blob/main/src/build/functions/edge.ts as we don't seem to handle `env` portion of manifest currently. We do handle `matchers`, `name`, `files` and `wasm`, but not `env` (and also not `assets` which does have https://github.com/opennextjs/opennextjs-netlify/issues/2697 for it)

mrstork: Thank you! Changes made in d2d41fa9f6aaed56d76ebf03744a70528c54844b

edge-runtime/lib/middleware.ts
mrstork: Setting the `set-cookie` header is done in `addMiddlewareHeaders` which happens right after the highlighted code.
https://github.com/opennextjs/opennextjs-netlify/blob/7795157ec4ab1270acb063abb28ac6fc5ad423a2/edge-runtime/lib/middleware.ts#L53-L59

Side note - Since `x-middleware-set-cookie` is in `INTERNAL_HEADERS` it gets filtered out from the request as soon as it reaches the lambda, so this change is pretty much just for response header cleanliness and the e2e test.

pieh (approved):

PR #2912, opened 2025-05-07
ci: remove `--timings` flag for Next.js repo e2e tests

serhalp:
## Description

This doesn't seem to provide any value for us. It just adds some test run time and instability to our suite, since this ends up attempting to read a `test-timings.json` file from disk (fails every time) and then makes a request to fetch timings data from a private Vercel endpoint, which sometimes fails with a 403 for some reason.

### Documentation

N/A

## Tests

N/A

It was added here without explanation: https://github.com/netlify/next-runtime-minimal/pull/334/commits/7ea73a2c7c6be6a898bdc6174e09a8198848a1cb


mrstork (approved):

mrstork:
Removed the tag so that the tests don't run again when we update the branch

PR #2935, opened 2025-05-19
test: add test cases for unexpected permanent cdn cache control cases

pieh:
## Tests

Adding some tests that seems to catch problematic cases of not cacheable route handler getting cdn cache control set

https://linear.app/netlify/issue/FRB-1707/next-runtime-is-sending-cdn-cache-control-headers-unexpectedly


src/build/content/static.test.ts
```diff
@@ -62,10 +65,7 @@ async function readDirRecursive(dir: string) {
   return paths
 }
 
-let failBuildMock: Mock<
-  Parameters<PluginContext['utils']['build']['failBuild']>,
-  ReturnType<PluginContext['utils']['build']['failBuild']>
->
+let failBuildMock: Mock<PluginContext['utils']['build']['failBuild']>
```
pieh, line 68: Unrelated, but as some point `Vitest.Mock` type changed and instead of requiring separate params and return type, it now wants just mocked function type so while I was here I adjusted type

src/build/content/static.ts
```diff
@@ -27,21 +27,23 @@ export const copyStaticContent = async (ctx: PluginContext): Promise<void> => {
...
 
     const fallbacks = ctx.getFallbacks(await ctx.getPrerenderManifest())
+    const fullyStaticPages = await ctx.getFullyStaticHtmlPages()
 
     try {
       await mkdir(destDir, { recursive: true })
       await Promise.all(
         paths
-          .filter((path) => !paths.includes(`${path.slice(0, -5)}.json`))
+          .filter((path) => !path.endsWith('.json') && !paths.includes(`${path.slice(0, -5)}.json`))
```
pieh, line 36: just small optim to return false early for json and avoid array search, doesn't change results, because if `path` ends with `.json` then paths will also include `${path.slice(0, -5)}.json` as it will be same path

src/build/content/static.ts
```diff
@@ -27,21 +27,23 @@ export const copyStaticContent = async (ctx: PluginContext): Promise<void> => {
...
             const html = await readFile(join(srcDir, path), 'utf-8')
             verifyNetlifyForms(ctx, html)
 
             const isFallback = fallbacks.includes(path.slice(0, -5))
+            const isFullyStaticPage = !isFallback && fullyStaticPages.includes(path)
 
             await writeFile(
               join(destDir, await encodeBlobKey(path)),
-              JSON.stringify({ html, isFallback } satisfies HtmlBlob),
+              JSON.stringify({ html, isFullyStaticPage } satisfies HtmlBlob),
```
pieh, line 46: we are no longer storing wether html blob is fallback, but we still use it to determine `isFullyStaticPage`

tests/fixtures/simple/app/api/slow-not-cacheable-with-html-read/route.js
```diff
@@ -0,0 +1,22 @@
...
+
+export async function GET() {
+  // This adds intentional delay here to make it more likely to hit some next-server
+  // initialization side-effects such as preloading page entries
+  // and trying to assert that side-effects do NOT impact the response.
+  // There is no way to force problematic side-effect scenario to happen without
+  // modifying the next internals.
+  // See https://github.com/vercel/next.js/blob/592401bb7fec83079716b2c9b090db580a63483f/packages/next/src/server/next-server.ts#L321-L327
+  // which starts NOT awaited async work
+  await new Promise((resolve) => setTimeout(resolve, 5_000))
```
pieh, line 12: without some kind of sleep in function, it was VERY hit'n'miss wether issue was reproducable - with this sleep, at this locally it was always reprodicable - but because the whole issue is timing/async - this is not 100% test scenario, but I couldn't figure out anything better without doing some kind of modifications to next.js code itself

tests/utils/fixture.ts
```diff
@@ -191,6 +192,20 @@ export async function runPluginStep(
...
+    redirects: [],
+  }
+
+  // load netlify.toml if it exists
+  if (existsSync(join(ctx.cwd, 'netlify.toml'))) {
+    const resolvedNetlifyConfig = await resolveNetlifyConfig({ cwd: ctx.cwd })
+    if (resolvedNetlifyConfig.config) {
+      netlifyConfig = resolvedNetlifyConfig.config
+    }
+  }
```
pieh, line 207: added this to support one of added test route handlers that reads non-nextjs produced html file and for that to work, we need to pass netlifyConfig produced from `netlify.toml` which instructs bundling to include some external files

tests/utils/lambda-helpers.mjs
```diff
@@ -143,7 +143,7 @@ export async function loadFunction(ctx, { env } = {}) {
         flags: flags ?? DEFAULT_FLAGS,
       },
       lambdaFunc: { handler },
-      timeoutMs: 4_000,
+      timeoutMs: 10_000,
```
pieh, line 146: bumped function invocation timeout to allow for 5s sleep in added test scenarios

src/run/next.cts
```diff
@@ -96,9 +104,16 @@ export async function getMockedRequestHandler(...args: Parameters<typeof getRequ
           const relPath = relative(resolve('.next/server/pages'), path)
           const file = await cacheStore.get<HtmlBlob>(relPath, 'staticHtml.get')
           if (file !== null) {
-            if (!file.isFallback) {
+            if (file.isFullyStaticPage) {
```
pieh, line 107: isFullyStaticPage will be false for fallbacks, but also additionally it somewhat inverts the check to only set `usedFsReadForNonFallback` for KNOWN static pages instead of not setting for fallbacks (but still setting it for everything else even if it's not pages router page)

package.json
```diff
@@ -51,6 +51,7 @@
     "@fastly/http-compute-js": "1.1.5",
     "@netlify/blobs": "^8.2.0",
     "@netlify/build": "^32.2.0",
+    "@netlify/config": "^23.0.1",
```
pieh, line 54: Added dev dependency is used only in integration tests - it's not being used otherwise

eduardoboucas (approved):

PR #2936, opened 2025-05-19
fix: handle case of env not being defined for edge functions

pieh:
## Description

Follow up to https://github.com/opennextjs/opennextjs-netlify/pull/2891 + adjustments to e2e setups after recent urls changes

### DIFF
--- a/src/build/functions/edge.ts
+++ b/src/build/functions/edge.ts
@@ -132,9 +132,11 @@ const copyHandlerDependencies = async (
 
   const outputFile = join(destDir, `server/${name}.js`)
 
-  // Prepare environment variables for draft-mode (i.e. __NEXT_PREVIEW_MODE_ID, __NEXT_PREVIEW_MODE_SIGNING_KEY, __NEXT_PREVIEW_MODE_ENCRYPTION_KEY)
-  for (const [key, value] of Object.entries(env)) {
-    parts.push(`process.env.${key} = '${value}';`)
+  if (env) {
+    // Prepare environment variables for draft-mode (i.e. __NEXT_PREVIEW_MODE_ID, __NEXT_PREVIEW_MODE_SIGNING_KEY, __NEXT_PREVIEW_MODE_ENCRYPTION_KEY)
+    for (const [key, value] of Object.entries(env)) {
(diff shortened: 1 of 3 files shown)


src/build/functions/edge.ts
```diff
@@ -132,9 +132,11 @@ const copyHandlerDependencies = async (
...
   const outputFile = join(destDir, `server/${name}.js`)
 
-  // Prepare environment variables for draft-mode (i.e. __NEXT_PREVIEW_MODE_ID, __NEXT_PREVIEW_MODE_SIGNING_KEY, __NEXT_PREVIEW_MODE_ENCRYPTION_KEY)
-  for (const [key, value] of Object.entries(env)) {
-    parts.push(`process.env.${key} = '${value}';`)
+  if (env) {
+    // Prepare environment variables for draft-mode (i.e. __NEXT_PREVIEW_MODE_ID, __NEXT_PREVIEW_MODE_SIGNING_KEY, __NEXT_PREVIEW_MODE_ENCRYPTION_KEY)
+    for (const [key, value] of Object.entries(env)) {
+      parts.push(`process.env.${key} = '${value}';`)
+    }
```
pieh, line 139: https://github.com/opennextjs/opennextjs-netlify/actions/runs/15105688788/job/42453967192#step:14:1785

next@^13 did not have `env` defined (it was added in https://github.com/vercel/next.js/commit/dcff078936c72f43d2a2e37a7899b1a6f631855a ), so we need to check if it's defined first

tests/netlify-deploy.ts
```diff
@@ -148,7 +148,7 @@ export class NextDeployInstance extends NextInstance {
 
     try {
       const deployUrlRegex = new RegExp(
-        /https:\/\/app\.netlify\.com\/sites\/(?<siteName>[^\/]+)\/deploys\/(?<deployID>[0-9a-f]+)/gm,
+        /https:\/\/app\.netlify\.com\/(sites|projects)\/(?<siteName>[^\/]+)\/deploys\/(?<deployID>[0-9a-f]+)/gm,
```
pieh, line 151: it's `projects` now, but kept old one in there in case there would be any kind of revert on that

eduardoboucas (approved):

PR #2943, opened 2025-05-21
chore: add e2e report skip entry for renamed `basepath.test.ts`

serhalp:
## Description

This test suite file was moved in https://github.com/vercel/next.js/pull/77630.

This results in it showing up as failed in new reports: https://682e414a03ba12f55c5f417b--runtime-e2e-report.netlify.app/.

Example run here:
https://github.com/opennextjs/opennextjs-netlify/actions/runs/15163298543#summary-42664185436.

I also clarified the `reason` field.

We could do something gross to expect the Netlify default 404 page or configure a custom 404 page that mimics Vercel's... but I don't think it's worth it.

### Documentation

N/A

## Tests

N/A

### DIFF
--- a/tests/test-config.json
+++ b/tests/test-config.json
@@ -199,8 +199,18 @@
       ]
     },
     {
+      "//": "This was the test file name before v15.3.0-canary.27 / v15.3.0",
       "file": "test/e2e/basepath.test.ts",
-      "reason": "Hard-coded Vercel error message",
+      "reason": "Hard-coded expectation of Vercel's default 404 page",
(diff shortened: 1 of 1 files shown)


tests/test-config.json
```diff
@@ -199,8 +199,18 @@
       ]
     },
     {
+      "//": "This was the test file name before v15.3.0-canary.27 / v15.3.0",
```
mrstork, line 202: Clever!

mrstork (approved):

PR #2959, opened 2025-06-11
test: print cli output when timing out

pieh:
## Description

To aid with debugging vercel/next tests that are just timing out without any meaningful output this captures the output and will print it out if `destroy()` is called before deploy finished to figure out if something is just slow or we are sometimes hanging on something.

See example https://github.com/opennextjs/opennextjs-netlify/actions/runs/15591799921/job/43913715730#step:20:82 (from test run on tmp commit that was intentionally timing out)

### DIFF
--- a/tests/netlify-deploy.ts
+++ b/tests/netlify-deploy.ts
@@ -154,6 +154,8 @@ export class NextDeployInstance extends NextInstance {
     const deployRes = await deployResPromise
 
     if (deployRes.exitCode !== 0) {
+      // clear deploy output to avoid printing it again in destroy()
+      this._deployOutput = ''
       throw new Error(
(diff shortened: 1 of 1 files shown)


tests/netlify-deploy.ts
```diff
@@ -168,7 +183,7 @@ export class NextDeployInstance extends NextInstance {
         require('console').log(`Logs: ${buildLogsUrl}`)
       }
     } catch (err) {
-      console.error(err)
+      require('console').error(err)
```
pieh, line 186: not related to main change

this error was not being printed out before - there is some weird console patching going on - so adjusting this to be same as other logs in case those might hide some important detail for us that would help debugging

serhalp (approved):

PR #2976, opened 2025-06-25
test: loose some assertions for fallback: true e2e tests to allow nonbreaking next@canary changes

pieh:
## Description

Follow up to https://github.com/opennextjs/opennextjs-netlify/pull/2974 that adjust assertions for currently failing tests now that `fallback: true` cache-control was fixed in https://github.com/vercel/next.js/pull/80865

### DIFF
--- a/tests/e2e/page-router.test.ts
+++ b/tests/e2e/page-router.test.ts
@@ -157,12 +157,15 @@ test.describe('Simple Page Router (no basePath, no i18n)', () => {
 
         const fallbackWasServed =
           useFallback && headers1['cache-status'].includes('"Next.js"; fwd=miss')
-        expect(headers1['netlify-cache-tag']).toBe(
-          fallbackWasServed ? undefined : `_n_t_${encodeURI(pagePath).toLowerCase()}`,
-        )
+        if (!fallbackWasServed) {
+          expect(headers1['netlify-cache-tag']).toBe(`_n_t_${encodeURI(pagePath).toLowerCase()}`)
+        }
         expect(headers1['netlify-cdn-cache-control']).toBe(
           fallbackWasServed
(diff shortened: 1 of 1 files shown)


tests/e2e/page-router.test.ts
```diff
@@ -157,12 +157,15 @@ test.describe('Simple Page Router (no basePath, no i18n)', () => {
 
         const fallbackWasServed =
           useFallback && headers1['cache-status'].includes('"Next.js"; fwd=miss')
-        expect(headers1['netlify-cache-tag']).toBe(
-          fallbackWasServed ? undefined : `_n_t_${encodeURI(pagePath).toLowerCase()}`,
-        )
+        if (!fallbackWasServed) {
+          expect(headers1['netlify-cache-tag']).toBe(`_n_t_${encodeURI(pagePath).toLowerCase()}`)
+        }
```
pieh, line 162: fallback does get cache tag with newest canary, but as long as it's not cacheable, we don't actually care about cache tag

tests/e2e/page-router.test.ts
```diff
@@ -157,12 +157,15 @@ test.describe('Simple Page Router (no basePath, no i18n)', () => {
...
         expect(headers1['netlify-cdn-cache-control']).toBe(
           fallbackWasServed
-            ? undefined
+            ? // fallback should not be cached
+              nextVersionSatisfies('>=15.4.0-canary.95')
+              ? `private, no-cache, no-store, max-age=0, must-revalidate, durable`
+              : undefined
```
pieh, line 168: Both of those mean "no-cache" in practice so those are generally the same in fact that fallback should not be cached

Just note that same change is applied 3 times in this PR, but will just comment on first case

PR #2984, opened 2025-06-27
test: add locale assertions for middleware handling

pieh:
## Tests

1. TODO

https://github.com/opennextjs/opennextjs-netlify/issues/2973
https://linear.app/netlify/issue/FRB-1919/version-5113-doesnt-seem-to-pass-the-correct-locale-anymore-from

### DIFF
--- a/tests/e2e/edge-middleware.test.ts
+++ b/tests/e2e/edge-middleware.test.ts
@@ -63,11 +63,13 @@ test.describe('json data', () => {
       describeLabel: 'NextResponse.next() -> getServerSideProps page',
       selector: 'NextResponse.next()#getServerSideProps',
       jsonPathMatcher: '/link/next-getserversideprops.json',
+      assertLocale: true,
     },
     {
       describeLabel: 'NextResponse.next() -> getStaticProps page',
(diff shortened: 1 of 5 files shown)


tests/fixtures/middleware-i18n/middleware.js
```diff
@@ -17,11 +20,9 @@ export async function middleware(request) {
...
 
   if (url.pathname.startsWith('/link/rewrite-me')) {
-    const rewriteUrl = new URL(
-      url.pathname.replace('/link/rewrite-me', '/link/rewrite-target'),
-      url,
-    )
-    return NextResponse.rewrite(rewriteUrl, {
+    // use NextUrl to rewrite the URL as it does preserve the locale
+    url.pathname = url.pathname.replace('/link/rewrite-me', '/link/rewrite-target')
+    return NextResponse.rewrite(url, {
```
pieh, line 25: Note: previous rewrite was just losing locale url due to using just `NextURL.pathname` to construct new `URL` effectively losing locale context in the process. 

This change keeps using `NextURL` and just mutate `pathname` portion of it which preserve `locale` context and resulting in seemingly working as expected rewrites

PR #2985, opened 2025-07-01
test: adjust assertions to work with validateRSCRequestHeaders enabled

pieh:
## Description

`next@canary` now are automatically opting into `validateRSCRequestHeaders` (note currently it seems like this is enabled in canary only - https://github.com/vercel/next.js/pull/80954 so possibly won't be applied to future stable releases and will remain canary-only)

This does result in additional 307 redirects for middleware redirects ... and rewrites (!). 

In this PR I adjusted our currently failing tests for cache poisoning to work with that additional redirect

### DIFF
--- a/tests/e2e/edge-middleware.test.ts
+++ b/tests/e2e/edge-middleware.test.ts
@@ -373,7 +373,11 @@ test.describe('RSC cache poisoning', () => {
   test('Middleware rewrite', async ({ page, middleware }) => {
     const prefetchResponsePromise = new Promise<Response>((resolve) => {
       page.on('response', (response) => {
-        if (response.url().includes('/test/rewrite-to-cached-page')) {
+        if (
+          (response.url().includes('/test/rewrite-to-cached-page') ||
+            response.url().includes('/caching-rewrite-target')) &&
+          response.status() === 200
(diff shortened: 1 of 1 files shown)


tests/e2e/edge-middleware.test.ts
```diff
@@ -373,7 +373,11 @@ test.describe('RSC cache poisoning', () => {
   test('Middleware rewrite', async ({ page, middleware }) => {
     const prefetchResponsePromise = new Promise<Response>((resolve) => {
       page.on('response', (response) => {
-        if (response.url().includes('/test/rewrite-to-cached-page')) {
+        if (
+          (response.url().includes('/test/rewrite-to-cached-page') ||
+            response.url().includes('/caching-rewrite-target')) &&
```
pieh, line 378: Note - the redirect that happens due to `validateRSCRequestHeaders` ( https://github.com/vercel/next.js/blob/21ff892d05e43644ebfd990d03d7f1742f9c46a7/packages/next/src/server/base-server.ts#L2086-L2098 ) result in seeing redirect with location being rewrite target

mrstork (approved):

PR #2987, opened 2025-07-05
Glob all server files instead of specific subdirectories

timneutkens:
## Tests

1. TODO

### DIFF
--- a/src/build/content/server.ts
+++ b/src/build/content/server.ts
@@ -99,13 +99,10 @@ export const copyNextServerCode = async (ctx: PluginContext): Promise<void> => {
         : ctx.nextDistDir
     const destDir = join(ctx.serverHandlerDir, nextFolder)
 
-    const paths = await glob(
-      [`*`, `server/*`, `server/chunks/*`, `server/edge-chunks/*`, `server/+(app|pages)/**/*.js`],
-      {
-        cwd: srcDir,
-        extglob: true,
-      },
-    )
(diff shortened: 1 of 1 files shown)


timneutkens:
Only matching `chunks/**/*` would work yeah 👍

tests/fixtures/hello-world-turbopack/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "hello-world-app",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "postinstall": "next build --turbopack",
```
pieh, line 6: Is it possible to set `--turbopack` via `next.config` instead of CLI toggle?

While the setup is sufficient right now for the added integration test, it would not use turbopack if we tried to use in deploy/e2e tests due to this setup defaulting to `next build` https://github.com/opennextjs/opennextjs-netlify/blob/main/tests/utils/create-e2e-fixture.ts#L142 and fixture definition not specyfing otherwise, which I'm happy to adjust for future proofing / avoiding confusion in the future..

I tried shortly following `--turbopack` toggle handling and seems to me like not possible to toggle it with `next.config` based on handling here https://github.com/vercel/next.js/blob/259f9a62203a2af128d429e34742fbe847c5d7a4/packages/next/src/cli/next-build.ts#L85-L103 which only check CLI toggles or `IS_TURBOPACK_TEST` env var, but figured worth checking

PR #3009, opened 2025-07-14
fix: handle edge runtime pages and middleware for turbopack builds

pieh:
## Description

Follow up to https://github.com/opennextjs/opennextjs-netlify/pull/2987

This will attempt to adjust handling to fix:
 - [x] edge runtime pages
 - [x] middleware

Previous PR did adjust setup for nodejs pages, but that is not sufficient for full coverage

## Tests

Expanded turbopack fixture and tests to test edge runtime pages and middleware

### DIFF
--- a/src/build/content/server.ts
+++ b/src/build/content/server.ts
@@ -105,6 +105,7 @@ export const copyNextServerCode = async (ctx: PluginContext): Promise<void> => {
         `server/*`,
         `server/chunks/**/*`,
         `server/edge-chunks/**/*`,
+        `server/edge/chunks/**/*`,
         `server/+(app|pages)/**/*.js`,
       ],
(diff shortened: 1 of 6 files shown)


src/build/content/server.ts
```diff
@@ -105,6 +105,7 @@ export const copyNextServerCode = async (ctx: PluginContext): Promise<void> => {
         `server/*`,
         `server/chunks/**/*`,
         `server/edge-chunks/**/*`,
+        `server/edge/chunks/**/*`,
```
pieh, line 108: needed for edge runtime pages when building with turbopack

`server/edge-chunks` seems to contain things like assets and wasm files (?) for middleware for non-turbopack builds

PR #3015, opened 2025-07-24
Update test-e2e.yml

pieh:
## Tests

1. TODO

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -63,11 +63,11 @@ jobs:
             echo "total=16" >> $GITHUB_OUTPUT
           elif [ "${{ github.event_name }}" == "pull_request" ]; then
             VERSION_SELECTORS=[\"latest\"]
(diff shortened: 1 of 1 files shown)


.github/workflows/test-e2e.yml
```diff
@@ -205,6 +205,7 @@ jobs:
           # one job may wait for deploys in other jobs (only one deploy may be in progress for
           # a given alias at a time), resulting in cascading timeouts.
           DEPLOY_ALIAS: vercel-next-e2e-${{ matrix.version_spec.selector }}-${{ matrix.group }}
+          NODE_OPTIONS: --import /home/runner/work/opennextjs-netlify/opennextjs-netlify/opennextjs-netlify/skip-integrations.mjs
```
pieh, line 208: there are sometimes failure related to fetching extensions, this might be due to doing a lot of deploys for the same site/account and hitting rate limiting or something, it's not quite clear - this for now just mocks responses to be empty to avoid those failures and see if that helps at all

PR #3018, opened 2025-07-30
[wip] Feat/node middleware support

pieh:
## Description

veeery wip, trying to assess state of support after doing some minimal manual tests for few cases which seemed to be working

## Tests

1. TODO

### DIFF
--- a/src/build/functions/edge.ts
+++ b/src/build/functions/edge.ts
@@ -194,11 +194,170 @@ export const clearStaleEdgeHandlers = async (ctx: PluginContext) => {
 }
 
 export const createEdgeHandlers = async (ctx: PluginContext) => {
+  // Edge middleware
   const nextManifest = await ctx.getMiddlewareManifest()
+  // Node middleware
+  const functionsConfigManifest = await ctx.getFunctionsConfigManifest()
+
   const nextDefinitions = [...Object.values(nextManifest.middleware)]
   await Promise.all(nextDefinitions.map((def) => createEdgeHandler(ctx, def)))
 
   const netlifyDefinitions = nextDefinitions.flatMap((def) => buildHandlerDefinition(ctx, def))
+
+  if (functionsConfigManifest?.functions?.['/_middleware']) {
+    const middlewareDefinition = functionsConfigManifest?.functions?.['/_middleware']
+    const entry = 'server/middleware.js'
+    const nft = `${entry}.nft.json`
+    const name = 'node-middleware'
+
+    // await copyHandlerDependencies(ctx, definition)
+    const srcDir = join(ctx.standaloneDir, ctx.nextDistDir)
+    // const destDir = join(ctx.edgeFunctionsDir, getHandlerName({ name }))
+
+    const fakeNodeModuleName = 'fake-module-with-middleware'
+
+    const fakeNodeModulePath = ctx.resolveFromPackagePath(join('node_modules', fakeNodeModuleName))
+
+    const nftFilesPath = join(ctx.nextDistDir, nft)
+    const nftManifest = JSON.parse(await readFile(nftFilesPath, 'utf8'))
+
+    const files: string[] = nftManifest.files.map((file: string) => join('server', file))
+    files.push(entry)
+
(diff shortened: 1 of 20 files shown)


src/run/next.cts
```diff
@@ -101,7 +105,7 @@ export async function getMockedRequestHandler(...args: Parameters<typeof getRequ
         // only try to get .html files from the blob store
         if (typeof path === 'string' && path.endsWith('.html')) {
           const cacheStore = getMemoizedKeyValueStoreBackedByRegionalBlobStore()
-          const relPath = relative(resolve('.next/server/pages'), path)
+          const relPath = relative(resolve(nextConfig.distDir, 'server/pages'), path)
```
pieh, line 108: not related to Node.js Middleware support, but this fixes serving static files if non-default `distDir` is used (test setup for Node.js middleware variants is hitting those conditions)

src/build/plugin-context.ts
```diff
@@ -14,6 +14,7 @@ import type { PrerenderManifest, RoutesManifest } from 'next/dist/build/index.js
 import type { MiddlewareManifest } from 'next/dist/build/webpack/plugins/middleware-plugin.js'
 import type { PagesManifest } from 'next/dist/build/webpack/plugins/pages-manifest-plugin.js'
 import type { NextConfigComplete } from 'next/dist/server/config-shared.js'
+import type { FunctionsConfigManifest } from 'next-with-cache-handler-v2/dist/build/index.js'
```
mrstork, line 17: No change needed - I continue to be a little worried about this way of getting types, but I still have no better solution to offer.

edge-runtime/lib/cjs.ts
```diff
@@ -0,0 +1,330 @@
...
+
+  if (!hookedIn) {
+    // @ts-expect-error - private untyped API
+    const original_resolveFilename = Module._resolveFilename.bind(Module)
+    // @ts-expect-error - private untyped API
+    Module._resolveFilename = (...args) => {
+      let target = args[0]
+      let isRelative = args?.[0].startsWith('.')
+
+      if (isRelative) {
```
mrstork, line 283: How do absolute paths work? Would an import have to start with `/virtual-root/...`?

src/build/plugin-context.ts
pieh: Yeah, I don't have really answer for it other than possibly adoption of Adapters API once it become stable could stabilize types situation as this will be public API with proper semver considerations (tho even with Adapters API if we will look to support multiple Next.js major versions with same adapter, we likely would run in similar problem again)

src/build/functions/edge.ts
```diff
@@ -161,26 +194,119 @@ const copyHandlerDependencies = async (
...
+  def: EdgeOrNodeMiddlewareDefinition,
 ): Array<ManifestFunction> => {
-  const functionHandlerName = getHandlerName({ name })
-  const functionName = name.endsWith('middleware')
-    ? 'Next.js Middleware Handler'
-    : `Next.js Edge Handler: ${page}`
-  const cache = name.endsWith('middleware') ? undefined : ('manual' as const)
+  const functionHandlerName = getHandlerName({ name: def.name })
+  const functionName = 'Next.js Middleware Handler'
+  const cache = def.name.endsWith('middleware') ? undefined : ('manual' as const)
```
mrstork, line 306: Perhaps we could move this into `EdgeOrNodeMiddlewareDefinition` i.e. `def.cacheName` or similar

src/build/functions/edge.ts
```diff
@@ -1,13 +1,43 @@
...
+  'matchers'
+>
+
+function nodeMiddlewareDefinitionHasMatcher(
+  definition: NodeMiddlewareDefinitionWithOptionalMatchers,
+): definition is NodeMiddlewareDefinition {
+  return Array.isArray(definition.matchers)
+}
+
+type EdgeOrNodeMiddlewareDefinition = {
```
mrstork, line 25: Thoughts on keeping the two types separate for now and using something like `EdgeMiddlewareDefinition | NodeMiddlewareDefinition` downstream?

PR #3043, opened 2025-08-11
test: adjust assertions for /_not-found being possibly added to prerender manifest

pieh:
## Description

To account for https://github.com/vercel/next.js/pull/82199

## Tests

1. TODO

### DIFF
--- a/tests/utils/next-version-helpers.mjs
+++ b/tests/utils/next-version-helpers.mjs
@@ -24,6 +24,20 @@ export function nextVersionSatisfies(condition) {
   return satisfies(checkVersion, condition, { includePrerelease: true }) || version === condition
 }
 
+export function isNextCanary() {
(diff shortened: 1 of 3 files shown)


tests/utils/next-version-helpers.mjs
```diff
@@ -24,6 +24,20 @@ export function nextVersionSatisfies(condition) {
...
+}
+
+export function shouldHaveAppRouterNotFoundInPrerenderManifest() {
+  // https://github.com/vercel/next.js/pull/82199
+
+  // The canary versions are out of band, as there stable/latest patch versions higher than base of canary versions
+  // and change was not backported to stable versions
+  return nextVersionSatisfies('>=15.4.2-canary.33') && isNextCanary()
+}
+
```
pieh, line 40: This is really annoying and frankly bad check that will likely need to be adjusted once that change is released in stable, but to me it feels better than doing some kind of filtering in tests themselves. At least it will be somewhat explicit of what full list of generated blobs are (...depending on next version :/ )

PR #3081, opened 2025-08-27
test: skip after Next.js repo tests

pieh:
## Description

See inline comment for context

### DIFF
--- a/tests/test-config.json
+++ b/tests/test-config.json
@@ -420,6 +420,10 @@
         "app-dir action size limit invalid config should respect the size set in serverActions.bodySizeLimit for plaintext fetch actions should error for requests that exceed the size limit",
         "app-dir action size limit invalid config should respect the size set in serverActions.bodySizeLimit for multipart fetch actions should error for requests that exceed the size limit"
       ]
+    },
+    {
+      "file": "test/e2e/app-dir/next-after-app-deploy/index.test.ts",
+      "reason": "Common test setup in this test suite is making 2 requests to page with 1 hour revalidation asserting it's the same, but on Netlify the first fetch will serve content we mark as stale, so second one will serve freshly generated one and those won't match."
     }
   ],
   "failures": [


tests/test-config.json
```diff
@@ -420,6 +420,10 @@
         "app-dir action size limit invalid config should respect the size set in serverActions.bodySizeLimit for plaintext fetch actions should error for requests that exceed the size limit",
         "app-dir action size limit invalid config should respect the size set in serverActions.bodySizeLimit for multipart fetch actions should error for requests that exceed the size limit"
       ]
+    },
+    {
+      "file": "test/e2e/app-dir/next-after-app-deploy/index.test.ts",
+      "reason": "Common test setup in this test suite is making 2 requests to page with 1 hour revalidation asserting it's the same, but on Netlify the first fetch will serve content we mark as stale, so second one will serve freshly generated one and those won't match."
```
pieh, line 426: We are failing on this step https://github.com/vercel/next.js/blob/589065d7884456180b991d4bd57da0e5bdb7fbff/test/e2e/app-dir/next-after-app-deploy/index.test.ts#L44-L56 which is shared by all tests in this test suite

Note that if we would run same "dummy" request as one that is done for PPR case, those tests do pass.

The whole thing likely happens because we mark cache entries generated at build time as stale (if they have `revalidate` set to something not permanent): 
https://github.com/opennextjs/opennextjs-netlify/blob/8202a9a6e0e19da20a31b0e87c8f8b90d06af707/src/build/content/prerendered.ts#L166-L167

and because of that:
 - first request serves that "stale" content generated at build time and regenerates the page in the background
 - second request serves freshly generated content
 - and content of those won't match

We do have our own e2e tests for next/after that were inspired by this test in https://github.com/opennextjs/opennextjs-netlify/blob/main/tests/e2e/after.test.ts so we will have test coverage for that particular feature still

PR #3093, opened 2025-09-02
fix: support segment data in cache entries

pieh:
## Description

For `clientSegmentCache` support. 

## Tests

Mainly relies on vercel/next.js tests like https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/segment-cache/basic/segment-cache-basic.test.ts

This is experimental feature, we will look to maybe some of our tests in the future


src/build/content/prerendered.ts
```diff
@@ -97,7 +121,10 @@ const buildAppCacheValue = async (
   if (
     !meta.status &&
     rsc.includes('NEXT_NOT_FOUND') &&
-    !meta.headers['x-next-cache-tags'].includes('/@')
+    !(
+      typeof meta.headers?.['x-next-cache-tags'] === 'string' &&
+      meta.headers?.['x-next-cache-tags'].includes('/@')
+    )
```
pieh, line 127: Note here - I typed `meta` (see line 80) above, and because `headers` is typed as `OutgoingHttpHeaders` (and not `Record<string,string>`) it required some additional type guard here to satisfy type checking

src/shared/cache-types.cts
```diff
@@ -44,17 +45,19 @@ type IncrementalCachedAppPageValueForMultipleVersions = Omit<
   'kind'
 > & {
   kind: 'APP_PAGE'
-}
+} & Pick<IncrementalCachedAppPageValueForNewVersions, 'segmentData'>
```
pieh, line 48: This is terrible (stitching type from different node versions, but if I just tried to use types for newer version, all hell broke lose, so this, while awful, felt like only reasonable option without getting stuck in fixing bunch of types

PR #3103, opened 2025-09-04
ci: don't install fixture's deps if next version requierment is not satisfied

pieh:
## Description

We are running tests on node 18 right now. Next@16 will be bumping minimal node version to 20.9 and will be failing builds if it's not met.

We do have some fixtures that require next@canary (which already require node 20.9) and if the next version constraint is not met for the fixture - we do install and build fixture, just with whatever is in package.json instead of version we ask to test.

This skips the installation and build for fixtures that do not meet next version constraints so that we could continue to test with node 18 on next@<16

## Tests

Not really adding tests, but I did run:
  - our own integration/e2e against next@latest (usual PR runs)
  - our own integration/e2e against next@canary in https://github.com/opennextjs/opennextjs-netlify/actions/runs/17492270525
  - vercel/next.js e2e tests for latest and canary - https://github.com/opennextjs/opennextjs-netlify/actions/runs/17492016738

I think they all work as expected now:
 - correct next version used (and fixtures that need newer version are skipped)
 - node@20 for next@canary and node@18 for everything else so we continue to test node@18 which is our min node version against next versions that support it

There are some failures there, but I think it's normal flakiness, I didn't want to restart them because it's the setup I wanted to assert on, not results on the tests themselves


tests/prepare.mjs
```diff
@@ -60,10 +60,12 @@ const promises = fixtures.map((fixture) =>
     })
     await Promise.all(publishDirectories.map((dir) => rm(dir, { recursive: true, force: true })))
 
-    if (NEXT_VERSION !== 'latest') {
-      await setNextVersionInFixture(cwd, NEXT_VERSION, {
-        logPrefix: `[${fixture}] `,
-      })
```
pieh, line 66: Note - the similar check is now done inside `setNextVersionInFixture` to avoid doing any actual changes for `latest`, but still allow to figure out if NEXT_VERSION satisfies potential fixture constraints like those:

https://github.com/opennextjs/opennextjs-netlify/blob/5e2813227aa439fc39bad993c51329833f502ea1/tests/fixtures/ppr/package.json#L15-L19

https://github.com/opennextjs/opennextjs-netlify/blob/5e2813227aa439fc39bad993c51329833f502ea1/tests/fixtures/middleware-node-runtime-specific/package.json#L15-L19

Note - those are not "dependencies" - those our own definitions that control which next versions are applicable for this fixture (for example for node middleware fixtures, we only allow next versions that actually support that, etc)

mrstork (approved):

PR #3129, opened 2025-09-16
fix: produce relative redirect location url for same origin

pieh:
## Description

This is not really fixing something currently not behaving correctly, but brings our middleware redirect handling closer to behavior on Vercel where "local" redirect doesn't result in `location` redirect header being absolute

### DIFF
--- a/tests/integration/middleware.test.ts
+++ b/tests/integration/middleware.test.ts
@@ -130,7 +130,7 @@ for (const {
         expect(response.status).toBe(307)
         expect(response.headers.get('location'), 'added a location header').toBeTypeOf('string')
         expect(
-          new URL(response.headers.get('location') as string).pathname,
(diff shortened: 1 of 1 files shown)


tests/integration/middleware.test.ts
```diff
@@ -130,7 +130,7 @@ for (const {
         expect(response.status).toBe(307)
         expect(response.headers.get('location'), 'added a location header').toBeTypeOf('string')
         expect(
-          new URL(response.headers.get('location') as string).pathname,
+          new URL(response.headers.get('location') as string, 'http://n').pathname,
```
pieh, line 133: Added `base` here because relative `location` without it just fails to construct `URL` object and throws `Invalid URL`. We do check `pathname` anyway so base will be discarded

PR #3147, opened 2025-10-01
Feat/skew protection

pieh:
## Tests

1. TODO


src/build/skew-protection.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  if (enabledOrDisabledReason === EnabledOrDisabledReason.OPT_IN_ENV_VAR) {
+    console.log(
+      `Setting up Next.js Skew Protection due to NETLIFY_NEXT_SKEW_PROTECTION=${process.env.NETLIFY_NEXT_SKEW_PROTECTION} environment variable.`,
+    )
+  } else {
+    console.log('Setting up Next.js Skew Protection.')
+  }
+
+  process.env.NEXT_DEPLOYMENT_ID = process.env.DEPLOY_ID
```
pieh, line 83: Ideally instead of setting this internal Next.js env var, we could set this via `next.config.js#deploymentId` ( https://github.com/vercel/next.js/blob/bd727d6f19852cc7ffd44b0321fd35b5f88c55e3/packages/next/src/server/config-shared.ts#L1065-L1068 ), but for now we don't have blessed way to mutate config

src/build/skew-protection.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  await mkdir(dirname(ctx.skewProtectionConfigPath), {
+    recursive: true,
+  })
+  await writeFile(
+    ctx.skewProtectionConfigPath,
+    JSON.stringify(
+      {
+        patterns: ['.*'],
+        sources: [
```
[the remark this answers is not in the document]
pieh: https://github.com/opennextjs/opennextjs-netlify/pull/3147/commits/92cfc9a67a3765a426612bf7a99bf32013132a71 (also dropped `null, 2` `JSON.stringify` params while using this opportunity as that was mainly done before for manual checking that everything is fine)

src/build/skew-protection.test.ts
```diff
@@ -0,0 +1,338 @@
...
+      const result = shouldEnableSkewProtection(mockCtx)
+
+      expect(result).toEqual({
+        enabled: true,
+        enabledOrDisabledReason: 'on-env-var',
+      })
+    })
+  })
+
+  describe('environment variable opt-out', () => {
```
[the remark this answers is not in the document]
pieh: https://github.com/opennextjs/opennextjs-netlify/pull/3147/commits/b2f988323a1a57ebb0d992b0dda6e4727c74f4d7?diff=split&w=1 grouped all env var handling (and kept opt-in, opt-out groups - I find it better for readability)

tests/e2e/skew-protection.test.ts
```diff
@@ -0,0 +1,574 @@
...
+            console.log(
+              `💾 Fixture and deployed site have been persisted. To clean up automatically, run tests without the 'E2E_PERSIST' environment variable.`,
+            )
+
+            return
+          }
+
+          await deployA.cleanup()
+          await deployB.cleanup()
+          await deleteSite(siteId)
```
mrstork, line 130: micro-optimization - if we're deleting the site, I don't think we need to delete both deploys first

pieh: Individual deploy cleanups here are to handle local fixture deletion
https://github.com/opennextjs/opennextjs-netlify/blob/184b506fa366fe9a9bb9683410955691cfc9c9c6/tests/utils/create-e2e-fixture.ts#L306-L313

PR #3180, opened 2025-10-15
fix: support revalidateTag SWR in 'use cache' handler

pieh:
## Description

Support `revalidateTag` SWR profiles for `use cache` handler. This is follow up to https://github.com/opennextjs/opennextjs-netlify/pull/3173 which added support for regular cache handler

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -5,11 +5,11 @@ on:
   pull_request:
     branches: [main]
   schedule:
-    - cron: '0 3 * * *' # Run every day at 3am UTC
+    - cron: "0 3 * * *" # Run every day at 3am UTC
   workflow_dispatch:
     inputs:
       versions:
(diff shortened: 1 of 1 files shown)


tests/fixtures/use-cache/app/next-cache.ts
```diff
@@ -0,0 +1,20 @@
...
+const NextCache = NextCacheTyped as any
+
+export const cacheLife: any =
+  'cacheLife' in NextCache
+    ? NextCache.cacheLife
+    : 'unstable_cacheLife' in NextCache
+      ? NextCache.unstable_cacheLife
+      : () => {
+          throw new Error('both unstable_cacheLife and cacheLife are missing from next/cache')
+        }
```
pieh: This was changed in https://github.com/vercel/next.js/pull/84877

The `cacheTag` handling is just future proofing (it's still `unstable_cacheTag` as of now)

PR #3181, opened 2025-10-16
test: deflake and another next 16 compat update for fixtures

pieh:
## Tests

1. TODO

### DIFF
--- a/src/run/handlers/request-context.cts
+++ b/src/run/handlers/request-context.cts
@@ -67,18 +67,28 @@ export function createRequestContext(request?: Request, context?: Context): Requ
     logger.debug('[NetlifyNextRuntime] Background revalidation request')
   }
 
+  async function recursivelyWaitForBackgroundWork() {
+    let settledPromisesCount = 0
+    while (settledPromisesCount < backgroundWorkPromises.length) {
+      const currentPromiseCount = backgroundWorkPromises.length
+      await Promise.allSettled(backgroundWorkPromises)
+      settledPromisesCount = currentPromiseCount
+    }
+  }
+
   return {
     isBackgroundRevalidation,
     captureServerTiming: request?.headers.has('x-next-debug-logging') ?? false,
     trackBackgroundWork: (promise) => {
-      if (context?.waitUntil) {
-        context.waitUntil(promise)
-      } else {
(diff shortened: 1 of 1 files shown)


tools/fetch-retry.mjs
```diff
@@ -0,0 +1,41 @@
...
+
+globalThis.fetch = async (...args) => {
+  let backoff = 100
+  for (let attempt = 1; attempt <= NUM_RETRIES; attempt++) {
+    try {
+      return await originalFetch.apply(globalThis, args)
+    } catch (error) {
+      let shouldRetry = false
+      // not ideal, but there is no error code for that
+      if (error.message === 'fetch failed' && attempt < NUM_RETRIES) {
```
pieh: this is happening very randomly and does not seem related to any specific fetch target:
 - it might be happening in fetch that CLI is doing to fetch extensions and result in confusing error about site not being linked (even tho we specifically pass site_id to cli command in our e2e tests)
- it might be happening in data fetches in fixtures (both when preparing fixtures for integration tests and when building and deploying for e2e tests)

it seems more related to runners and hitting potential limits there

tests/utils/create-e2e-fixture.ts
```diff
@@ -542,12 +542,18 @@ export const fixtureFactories = {
...
+      },
     }),
   nxIntegratedDistDir: () =>
     createE2EFixture('nx-integrated', {
       packagePath: 'apps/custom-dist-dir',
       buildCommand: 'nx run custom-dist-dir:build',
       publishDirectory: 'dist/apps/custom-dist-dir/dist',
+      env: {
+        NX_ISOLATE_PLUGINS: 'false',
+      },
```
pieh: nx on next canary starting failing with errors mentioned in https://github.com/nrwl/nx/issues/27040 (but it's not exactly the same, because issue mention `postinstall`, while for us errors were happening at `next build` time, so despite that issue being closed with a fix, I think we are hitting similar problem in different place). I tried `NX_ISOLATE_PLUGINS` which was suggested as workaround there and that seemed to work here as well

pieh:
I'll close this as some adjustments were merged separately and some are included in https://github.com/opennextjs/opennextjs-netlify/pull/3180

PR #3190, opened 2025-10-21
fix: support skipProxyUrlNormalize

pieh:
## Description

`skipMiddlewareUrlNormalize` next.config option is being deprecated (as part of overall `middleware` -> `proxy` rename) and is being replaced by `skipProxyUrlNormalize`:
```
⚠ `skipMiddlewareUrlNormalize` is deprecated. Please use `skipProxyUrlNormalize` instead in next.config.js.
```

This adds support for the renamed config option as we need specific handling for it in our middleware/proxy handling.

Added integration tests asserting that `nextUrl` in middleware/proxy is not normalized when `skipMiddlewareUrlNormalize` / `skipProxyUrlNormalize` is used. Note that while we supported `skipMiddlewareUrlNormalize` for long time, we didn't actually have test for it 😱 , so test is added for both now deprecated and new config option

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@netlify/build": "^35.1.7",
         "@netlify/config": "^24.0.4",
         "@netlify/edge-bundler": "^14.5.5",
-        "@netlify/edge-functions": "^2.17.1",
+        "@netlify/edge-functions": "^3.0.0",
         "@netlify/edge-functions-bootstrap": "^2.14.0",
         "@netlify/eslint-config-node": "^7.0.1",
         "@netlify/functions": "^4.3.0",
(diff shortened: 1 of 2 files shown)


tests/fixtures/proxy-i18n-skip-normalize/proxy.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    response.headers.set('x-next-url-pathname', request.nextUrl.pathname)
+
+    return response
+  }
+}
+
+export const config = {
+  runtime: 'nodejs',
+}
```
pieh: It seems like Proxy will be Node.js only (tho it's not as of now), so I do add this explicitly here in hope the test setup won't break when Next.js ships the change to make Proxy Node.js only. If that change end up not shipping, I'll adjust the setup to run in both Edge and Node.js runtimes as we do for most of our middleware fixtures

PR #3195, opened 2025-10-22
ci: adjustments after next@16 stable release

pieh:
## Description

All of these are adjustments to tests and test setup:

1. We were checking if we are testing `canary` when deciding to use node@20 (which is now minimun node version needed for next@16), but now with next@16 stable release this no longer make sense to do, so now we finally check major of resolved next version to decide node version
2. Adjustment for test fixture added yesterday for Next.js changes shipped after fixture was introduced
3. Adjust some next version constraint check for a behavior that was next@canary only until 16.0.0 was released

### DIFF
--- a/tests/utils/next-version-helpers.mjs
+++ b/tests/utils/next-version-helpers.mjs
@@ -41,8 +41,10 @@ export function shouldHaveAppRouterNotFoundInPrerenderManifest() {
 export function shouldHaveAppRouterGlobalErrorInPrerenderManifest() {
   // https://github.com/vercel/next.js/pull/82444
 
-  // this is not used in any stable version yet
-  return isNextCanary() && nextVersionSatisfies('>=15.5.1-canary.4')
+  return (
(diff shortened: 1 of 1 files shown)


tests/utils/next-version-helpers.mjs
```diff
@@ -41,8 +41,10 @@ export function shouldHaveAppRouterNotFoundInPrerenderManifest() {
 export function shouldHaveAppRouterGlobalErrorInPrerenderManifest() {
   // https://github.com/vercel/next.js/pull/82444
 
-  // this is not used in any stable version yet
-  return isNextCanary() && nextVersionSatisfies('>=15.5.1-canary.4')
+  return (
+    (isNextCanary() && nextVersionSatisfies('>=15.5.1-canary.4')) ||
+    nextVersionSatisfies('>=16.0.0')
+  )
```
pieh: Another change that graduated from canary-only to be enabled in stable releases (this is used in our tests only to make correct assertions, it doesn't change actual support for it)

PR #3218, opened 2025-10-29
perf: use conditional blob gets if same blob was fetched before for previous requests

pieh:
## Description

This introduces usage of conditional blob get requests if we fetched or stored given blob key on previous requests in given lambda instance. This allows to avoid fetching blob data if blob didn't change since last time given lambda.

This builds on top of memoization we already do within same request introduced in https://github.com/opennextjs/opennextjs-netlify/pull/2777

It does add global blobs map with weak refs to entries stored in our current LRU cache as a mechanism to offload memory management to what we already have and not having to do complex tracking of when global entries should be disposed.

Existing tests for memoized storage were updated and expanded to tests conditional blob gets scenario

### DIFF
--- a/src/run/storage/storage.test.ts
+++ b/src/run/storage/storage.test.ts
@@ -90,7 +90,7 @@ describe('getMemoizedKeyValueStoreBackedByRegionalBlobStore', () => {
     const store = getMemoizedKeyValueStoreBackedByRegionalBlobStore()
     const get1 = await store.get(TEST_KEY, OTEL_SPAN_TITLE)
 
-    expect(mockedStore.getWithMetadata, 'Blobs should be requested').toHaveBeenCalledTimes(1)
+    expect(mockedStore.getWithMetadata, 'Blobs should be requested').toHaveBeenCalledOnce()
     expect(get1, 'Expected blob should be returned').toBe(TEST_DEFAULT_VALUE)
 
     const get2 = await store.get(TEST_KEY, OTEL_SPAN_TITLE)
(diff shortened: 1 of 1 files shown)


src/run/storage/storage.cts
```diff
@@ -25,16 +25,48 @@ export const getMemoizedKeyValueStoreBackedByRegionalBlobStore = (
...
+          // if we don't get blob (null) or etag for some reason is missing,
+          // we still want to store resolved blob value so that it could be reused
+          // within the same request
+          inMemoryCache.set(key, blob)
+        }
+
+        span?.setAttributes({
+          etag: result?.etag,
+          reusingPreviouslyFetchedBlob: shouldReuseMemoizedBlob,
+          status: blob ? (shouldReuseMemoizedBlob ? 'Hit, no change' : 'Hit') : 'Miss',
```
pieh: this is pretty much modeling 304, 200, 404 response status codes from underlying blobs response

hopefully this is temporary until we can nicely "merge" next-runtime produced span with blobs span without losing next.js specific context (primarly logical cache key vs blob key which is encoded and the context of blobs get - is it for tag manifest, static html blob or cachehandler entry)

src/run/storage/storage.test.ts
```diff
@@ -151,55 +244,72 @@ describe('getMemoizedKeyValueStoreBackedByRegionalBlobStore', () => {
...
-      mockedStore.get.mockClear()
+      expect(mockedStore.getWithMetadata, 'Value should be read from blobs').toHaveBeenCalledTimes(
+        1,
+      )
+      mockedStore.getWithMetadata.mockClear()
 
       await store.get('heavy-route-1', OTEL_SPAN_TITLE)
-      expect(mockedStore.get, 'Value should be read from memory').toHaveBeenCalledTimes(0)
-      mockedStore.get.mockClear()
+      expect(mockedStore.getWithMetadata, 'Value should be read from memory').toHaveBeenCalledTimes(
```
mrstork: There is a `toHaveBeenCalledOnce` if you'd like

PR #3222, opened 2025-10-31
fix: add Buffer shim if is not globally available, adjust process and AsyncLocalStorage shims to only be applied when not available

pieh:
## Description

We discovered some problems when trying to bump deno version in https://github.com/opennextjs/opennextjs-netlify/pull/3221 which seemed to require adding `Buffer` global shim. While doing so I figured it might make sense to only try to apply global shims conditionally if they things are not globally available already


edge-runtime/shim/node.js
```diff
@@ -1,16 +1,21 @@
...
+}
+
+if (typeof AsyncLocalStorage === 'undefined') {
+  globalThis.AsyncLocalStorage = (await import('node:async_hooks')).AsyncLocalStorage
+}
 
-globalThis.AsyncLocalStorage = AsyncLocalStorage
+if (typeof Buffer === 'undefined') {
+  globalThis.Buffer = (await import('node:buffer')).Buffer
+}
```
pieh: this is fun one - Buffer is not globally defined in deno 1.39.0, 1.46.3 or 2.2.4 (from tested versions), but is available in 2.4.2

PR #3262, opened 2025-11-19
fix: incorrect path of middleware nft in setups using base directory

pieh:
## Description

Details in https://github.com/opennextjs/opennextjs-netlify/pull/3262#pullrequestreview-3482035279

## Tests

Added test case for monorepo setup that is using base directory instead of package path and use proxy / node middleware

### DIFF
--- a/src/build/functions/edge.ts
+++ b/src/build/functions/edge.ts
@@ -210,7 +210,7 @@ const copyHandlerDependenciesForNodeMiddleware = async (ctx: PluginContext) => {
 
   const entry = 'server/middleware.js'
   const nft = `${entry}.nft.json`
-  const nftFilesPath = join(process.cwd(), ctx.distDir, nft)
+  const nftFilesPath = join(ctx.publishDir, nft)
   const nftManifest = JSON.parse(await readFile(nftFilesPath, 'utf8'))
 
   const files: string[] = nftManifest.files.map((file: string) => join('server', file))


tests/utils/create-e2e-fixture.ts
```diff
@@ -349,24 +357,27 @@ export async function deploySiteWithBuildbot(
   newZip.addLocalFolder(isolatedFixtureRoot, '', (entry) => {
     if (
       // don't include node_modules / .git / publish dir in zip
-      entry.startsWith('node_modules') ||
-      entry.startsWith('.git') ||
+      entry.includes('node_modules') ||
+      entry.includes('.git') ||
```
pieh: new fixture has nested directory with `node_modules`, so this ensures we skip zipping those

tests/utils/create-e2e-fixture.ts
```diff
@@ -248,7 +256,7 @@ async function installRuntime(
       env['YARN_ENABLE_SCRIPTS'] = 'false'
       break
     case 'pnpm':
-      command = `pnpm add file:${join(isolatedFixtureRoot, packageName)} ${
+      command = `pnpm add file:${relativePathToPackage} ${
```
pieh: relative path needed for buildbot to actually find tarballed runtime in zip file - this possibly will have to be replicated for other package managers if we will need to use buildbot for those (I only did `pnpm` here, `npm` already worked when buildbot builds were added for skew protection)

PR #3279, opened 2025-11-26
fix: de-duplicate blob operation spans

mrstork:
## Description

Depends on https://github.com/netlify/primitives/pull/544

## Tests

1. Install the `@netlify/blobs` package from the changes linked above
2. Deploy and produce a trace
3. Observe that blob actions no longer have duplicate spans

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "@netlify/edge-functions-bootstrap": "^2.17.1",
         "@netlify/eslint-config-node": "^7.0.1",
         "@netlify/functions": "^5.1.0",
-        "@netlify/otel": "^5.0.0",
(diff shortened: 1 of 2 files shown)


src/run/storage/storage.cts
pieh: We can remove:
 - `blobKey` (duplicate with https://github.com/netlify/primitives/blob/52252659d4809f8d9157739431674edb44e960a5/packages/blobs/src/store.ts#L323 )
 - `previousEtag` (duplicate with https://github.com/netlify/primitives/blob/52252659d4809f8d9157739431674edb44e960a5/packages/blobs/src/store.ts#L327)

pieh (approved):

PR #3282, opened 2025-11-27
fix: handle node_modules in standalone's dist dir

pieh:
## Description

Adjustments to handle recent e2e failures on next@canary ( https://github.com/opennextjs/opennextjs-netlify/actions/runs/19727042798 ).

Some details:

It seems like imports in pages router with turbopack builds now generate additional `node_modules` directory with symlinks.

image

We currently don't handle `standalone/.next/node_modules` (marked with red on above screenshot). We only handle `standalone/node_modules` (marked with green). This adds handling for "new" node_modules location, preserving hierarchy (so no merging of them, we will have `node_modules` and `.next/node_modules` in lambda)

It does look like https://github.com/vercel/next.js/commit/7905cb4cc3246f20145694ee490ff4ecd926541f#diff-86d6014a082abde94fef6b35552605715afe6eb9e7e467ac931d2f153cd3b934 is what introduced change. It does seem to me that that there is some kind of bug there to be honest, because we are hitting case of producing those hashed symlinks on DIRECT dependencies that we don't even declare as external in config and because of that I'm expanding our tests (following tests added in that commit) to ensure we have coverage in future as well in case it is indeed a bug that we fell into this new behavior despite seemingly not meeting criteria for it.

Additionally ... while expanding test fixture I've hit yet another unhandled edge case, where we were not copying some some top-level directories from root of standalone directory (marked directory missing in NF)

image

### DIFF
--- a/e2e-report/package-lock.json
+++ b/e2e-report/package-lock.json
@@ -18,7 +18,7 @@
         "daisyui": "^4.12.2",
         "eslint": "^9.34.0",
         "eslint-config-next": "^15.5.0",
-        "netlify-cli": "^23.12.1",
+        "netlify-cli": "^23.12.2",
         "postcss": "^8.4.38",
         "sass": "^1.77.1",
         "tailwindcss": "^3.4.4"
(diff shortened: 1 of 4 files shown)


src/build/content/server.ts
```diff
@@ -116,36 +116,53 @@ export const copyNextServerCode = async (ctx: PluginContext): Promise<void> => {
...
+      const filter = ctx.constants.IS_LOCAL ? undefined : nodeModulesFilter
+      const src = join(srcDir, 'node_modules')
+      const dest = join(destDir, 'node_modules')
+      await cp(src, dest, {
+        recursive: true,
+        verbatimSymlinks: true,
+        force: true,
+        filter,
+      })
+    }
```
pieh: This part is fix for next@canary change to ensure the "hashed" artificial `node_modules` are included in NF (see first part of description for more details including first screenshot)

PR #3349, opened 2025-12-30
fix: handle PPR shells for fully dynamic segments on Next.js 16.1.0+

### LINKED ISSUE #3344: [Bug]: ENOENT for .prefetch.rsc with cacheComponents and Next.js 16.1.1 dynamic routes
## Description

When using `cacheComponents: true` with Next.js 16.1.1, the plugin fails during build with:

```
Error: Failed assembling prerendered content for upload
ENOENT: no such file or directory, open '.next/server/app/chat/[id].prefetch.rsc'
```

## Environment

- **@netlify/plugin-nextjs**: 5.15.3
- **Next.js**: 16.1.1
- **Node**: 22.x
- **Package manager**: Bun

## next.config.ts

```ts
const nextConfig: NextConfig = {
  cacheComponents: true,
  // ... other config
};
```

## Analysis

The route `/chat/[id]` is a `"use client"` page wrapped in Suspense. The build correctly identifies it as PPR:

```
├ ◐ /chat/[id]
│ └ /chat/[id]
```

### Prerender Manifest

The prerender manifest shows:
```json
{
  "experimentalPPR": true,
  "renderingMode": "PARTIALLY_STATIC",
  "prefetchDataRoute": null,
  "dataRoute": "/chat/[id].rsc"
}
```

Note: `prefetchDataRoute: null` - Next.js explicitly does NOT generate `.prefetch.rsc` files.

### Actual Files Generated

```
.next/server/app/chat/
├── [id].html (empty)
├── [id].meta
├── [id].segments/
│   ├── _full.segment.rsc
│   ├── _head.segment.rsc
│   ├── _index.segment.rsc
│   ├── _tree.segment.rsc
│   └── chat.segment.rsc
```

**No `.rsc` or `.prefetch.rsc` files exist at the route level.** Next.js 16 with `cacheComponents` uses the new `.segments/` directory structure instead.

## Expected Behavior

Since `renderingMode: "PARTIALLY_STATIC"` and `prefetchDataRoute: null`, the plugin should:
1. Not require `.rsc` or `.prefetch.rsc` files
2. Use the `.segments/` directory data instead

PR #3275 was supposed to make RSC files optional for PPR pages, but this case is still failing.

## Suspected Cause

The fix in PR #3275 checks:
```typescript
if (rscIsRequired) { throw error }
```

Where `rscIsRequired` is `false` when `renderingMode === 'PARTIALLY_STATIC'`.

However, the error is still being thrown, suggesting either:
1. The `renderingMode` isn't being correctly read from the manifest
2. A different code path is being hit that doesn't have the optional RSC logic
3. The new `.segments/` format requires additional handling

## Reproduction

1. Create a Next.js 16.1.1 app with `cacheComponents: true`
2. Add a dynamic route like `/chat/[id]` with a `"use client"` page
3. Deploy to Netlify

## Workaround

Currently none found that preserves `cacheComponents: true`.

serhalp:
## Description

The previous fix (https://github.com/opennextjs/opennextjs-netlify/pull/3275) made `.rsc` files optional for entries in `manifest.routes` with `renderingMode === 'PARTIALLY_STATIC'`. However, actual PPR shells are processed through another code path that wasn't updated in this way.

[When using Cache Components, routes _with_ or _without_ a `generateStaticParams` behave differently](https://nextjs.org/docs/app/api-reference/file-conventions/dynamic-routes#with-cache-components). The former have no concrete prerendered pages (no known params at build time), only the shell. Such routes have a `fallback` pointing to the shell, but since the shell code path still requires `.rsc` files to be found, the build fails on Next.js `^16.1.0 || ^16.0.2-canary.29`, versions that [no longer emit these files](https://github.com/vercel/next.js/pull/86100).

### Documentation

https://nextjs.org/docs/app/api-reference/file-conventions/dynamic-routes#with-cache-components

## Tests

This PR adds integration tests for this scenario. They were failing before this fix.

There's also a repro here: https://github.com/serhalp/repro-nextjs-netlify-3344

Fixes #3344

### DIFF
--- a/tests/fixtures/ppr/app/dynamic-params/[id]/page.js
+++ b/tests/fixtures/ppr/app/dynamic-params/[id]/page.js
@@ -1,0 +1,44 @@
+import { Suspense } from 'react'
+import { connection } from 'next/server'
+
+// no `generateStaticParams` export
+
+async function getData(id) {
+  await connection()
+  const res = await fetch(`https://api.tvmaze.com/shows/${id}`, {
+    next: {
+      tags: [`show-${id}`],
+    },
+  })
+  await new Promise((res) => setTimeout(res, 3000))
+  return res.json()
+}
+
+async function Content({ id }) {
+  const data = await getData(await id)
+
+  return (
+    <>
+      <h1>Dynamic Page (dynamic params): {id}</h1>
+      <dl>
+        <dt>Show</dt>
+        <dd>{data.name}</dd>
+        <dt>Param</dt>
+        <dd>{await id}</dd>
+        <dt>Time</dt>
+        <dd data-testid="date-now">{new Date().toISOString()}</dd>
+      </dl>
+    </>
+  )
+}
+
+// This is a dynamic page (segment) where params are NOT statically generated (dynamic, at request time)
+export default async function DynamicPageWithDynamicParams({ params }) {
+  return (
+    <main>
+      <Suspense fallback={<div>loading...</div>}>
+        <Content id={params.then(({ id }) => id)} />
+      </Suspense>
+    </main>
+  )
+}
--- a/tests/integration/simple-app.test.ts
+++ b/tests/integration/simple-app.test.ts
@@ -384,6 +384,7 @@ test<FixtureTestContext>('cacheable route handler is cached on cdn (revalidate=1
 test<FixtureTestContext>('rewrites to external addresses dont use compression', async (ctx) => {
   await createFixture('simple', ctx)
   await runPlugin(ctx)
+
   const page = await invokeFunction(ctx, {
     url: '/rewrite-no-basepath',
     headers: { 'accept-encoding': 'gzip' },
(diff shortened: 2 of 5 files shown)


tests/fixtures/ppr/app/dynamic-params/[id]/page.js
```diff
@@ -0,0 +1,44 @@
...
+  await new Promise((res) => setTimeout(res, 3000))
+  return res.json()
+}
+
+async function Content({ id }) {
+  const data = await getData(await id)
+
+  return (
+    <>
+      <h1>Dynamic Page (dynamic params): {id}</h1>
```
Comment (serhalp) [inline tests/fixtures/ppr/app/dynamic-params/[id]/page.js]: As [documented here](https://nextjs.org/docs/app/api-reference/file-conventions/dynamic-routes#with-cache-components), `await params` must be wrapped in a Suspense boundary when using Cache Components and no `generateStaticParams`, otherwise the build fails with https://nextjs.org/docs/messages/blocking-route.

AFAICT, this is not strictly related to the bug this PR is fixing. It's just another incidental difference between dynamic segments with/without `generateStaticParams`.

tests/fixtures/ppr/app/dynamic-params/[id]/page.js
```diff
@@ -0,0 +1,44 @@
...
+    </>
+  )
+}
+
+// This is a dynamic page (segment) where params are NOT statically generated (dynamic, at request time)
+export default async function DynamicPageWithDynamicParams({ params }) {
+  return (
+    <main>
+      <Suspense fallback={<div>loading...</div>}>
+        <Content id={params.then(({ id }) => id)} />
```
Comment (serhalp) [inline tests/fixtures/ppr/app/dynamic-params/[id]/page.js]: 🤷🏼 pattern recommended by the docs: https://nextjs.org/docs/messages/blocking-route#params-and-searchparams

tests/integration/simple-app.test.ts
```diff
@@ -399,17 +400,19 @@ test.skipIf(
...
-      '/2',
       shouldHaveAppRouterNotFoundInPrerenderManifest() ? undefined : '/404',
-      isExperimentalPPRHardDeprecated() ? undefined : '/[dynamic]',
+      isExperimentalPPRHardDeprecated() ? undefined : '/static-params/[id]',
       shouldHaveAppRouterGlobalErrorInPrerenderManifest() ? '/_global-error' : undefined,
       shouldHaveAppRouterNotFoundInPrerenderManifest() ? '/_not-found' : undefined,
+      '/dynamic-params/[id]',
       '/index',
+      '/static-params/1',
+      '/static-params/2',
```
serhalp: these are `/1` and `/2` renamed for clarity

PR #3476, opened 2026-04-08
fix: set cache tag for getStaticProps's redirect

pieh:
## Description

When `getStaticProps` returns `redirect` property, it is using `REDIRECT` cache kind. We were completely skipping handling for this which resulted in:
 - no cache tags setting (so those routes could never on-demand revalidate)
 - `CacheHandler.get` would return `null` so would act as a "miss" and would each time regenerate when actual invocation happened. It would still be cached on CDN, but the "related" route (the "html"/json pair) would rerun instead of serving frtom cache and also rewrite cases would also regenerate.
 
This adds support for `REDIRECT` cache kind to be on par with Page Router actual pages

## Tests

Added integration test for veryfying that cache tags are set and e2e test testing flow of producing redirect, using `res.revalidate()` on it and regenerating slightly modified redirect (different target and adding `revalidate` property so also testing that CDN cache-control is being handled properly)

Fixes FRB-2168


tests/e2e/dynamic-cms.test.ts
```diff
@@ -105,4 +113,142 @@ test.describe('Dynamic CMS', () => {
...
+    for (const { label, contentKey, expectedCacheTag, urlPath, pathToRevalidate, isPageData } of [
+      {
+        label: 'Invalidates getStaticProps redirect from durable cache (implicit default locale)',
+        urlPath: `/content/html-implicit-default-locale-${postFix}`,
+        contentKey: `html-implicit-default-locale-${postFix}`,
+        expectedCacheTag: `_n_t_/en/content/html-implicit-default-locale-${postFix}`,
+        isPageData: false,
+      },
+      {
+        label: 'Invalidates 404 html from durable cache (explicit default locale)',
```
pieh: ops, I need to update test names here - the structure was copied from the other test in this test suite

pieh: Renamed in https://github.com/opennextjs/opennextjs-netlify/pull/3476/commits/98c2f7fb40c43be81bba7c1942f04cb3b847917d, without other changes, so while tests are still running now - they _should_ pass like they did on previous commit

src/run/handlers/cache.cts
```diff
@@ -399,6 +401,14 @@ export class NetlifyCacheHandler implements CacheHandlerForMultipleVersions {
             },
           }
         }
+        case 'REDIRECT': {
```
serhalp, line 404: Seems like https://typescript-eslint.io/rules/switch-exhaustiveness-check/ would have caught this much earlier!

serhalp (approved):
💯

src/run/handlers/cache.cts
pieh: It would help, yeah. 

We actually do have one more unhandled here with those changes - `IMAGE` - but as far as I can tell this is for Image CDN, which should make it not applicable, because we offload that to our Image CDN. I'll look to try follow up with that eslint rule and adding explicit case for `IMAGE` so it's clear that this is not being handled here, not because it was missed, but because it's not applicable.

I do want to mention that even with that rule and handling all those types - the exact shape comes from whatever Next.js version user is using - we have this abonimation trying to create unions of types from multiple Next.js versions, but setting it up is a bit manual - both type merging and having having multiple Next.js versions installed so types are importable (with npm aliases like https://github.com/opennextjs/opennextjs-netlify/blob/4c7847b8dc05ebf2ea07e8ade8fdb029e23210e1/package.json#L88-L90) so I doubt this will be ever something very reliable, but that's not an excuse not to try to improve it to extent that is possible at least.

PR #3479, opened 2026-04-16
fix: don't optimize same-url rewrite if response headers need to be applied

pieh:
## Description

We have few cases of optimizations in middleware handling where we allow request to continue with usual flow for cases when middleware does not need to apply any modifications to response. This is because it's more efficient for Netlify Edge Function to `return undefined` vs `return context.next()` (the origin response does NOT need to go back to EF for the `undefined` / no modifications needed case). 

However checks we do to determine if modification is needed were not covering case that require response header additions which was resulting in unexpected cases of response headers that should be added by middleware missing. This expands the checks to make sure we don't do optimized version of handling for cases when response headers set in middleware need to be applied.

## Tests

Initial failing test cases added in PR. Need to be expanded to cover all relevant cases being addressed with changes.


tests/fixtures/middleware/middleware-shared.ts
```diff
@@ -83,6 +83,14 @@ const getResponse = (request: NextRequest) => {
...
   }
 
+  if (request.nextUrl.pathname === '/test/rewrite-to-self') {
+    return NextResponse.rewrite(new URL('/test/rewrite-to-self', request.url), {
+      request: {
+        headers: requestHeaders,
+      },
+    })
+  }
+
```
pieh, line 93: This diff might be confusing because it doesn't seem like we add any headers here, but all this is wrapped in shared setup for all the tested cases that adds request/response headers - https://github.com/opennextjs/opennextjs-netlify/blob/8090604519245cbb352b37e6569e7d720aab3736/tests/fixtures/middleware/middleware-shared.ts#L9-L32

pieh:
While not handled yet, because I need to dig into it - I am marking there is some similar potential problem here https://github.com/opennextjs/opennextjs-netlify/blob/00a2779ea121ad5186b075ebce3b4ba053b0ddd2/edge-runtime/lib/response.ts#L238-L245

Just I'm not sure yet why/when this is applied and need to dig into the rationale behind this redirect case handling even existing. Originally it was added as part of https://github.com/opennextjs/opennextjs-netlify/commit/402caaea0d5d8f2b2e09c92307ad7e9eab3b99d4 (without tests :( which is the troubling part to figure out context)

pieh:
PPR test failing in `
Run tests / test (5, ubuntu-latest, canary)` is unrelated and being addressed in https://github.com/opennextjs/opennextjs-netlify/pull/3485

tests/integration/middleware.test.ts
```diff
@@ -37,7 +37,7 @@ for (const {
     isNodeMiddleware: false,
     label: 'Edge runtime middleware',
   },
-  hasNodeMiddlewareSupport()
+  false && hasNodeMiddlewareSupport()
```
serhalp: 👁️ 👃🏼 👁️

serhalp (dismissed):
LGTM, great find!

tests/integration/middleware.test.ts
pieh: Sigh, this was supposed to be `.only` for table tests used just for iteration. But this is not being picked up as actual `.only` would be heh.

Will fix

pieh: Fixed in https://github.com/opennextjs/opennextjs-netlify/pull/3479/changes/7d97f9a6f255e24d119ad18f64a01a15502a3688

serhalp (approved):

PR #3489, opened 2026-04-20
Add GitHub Actions workflow for Node.js with Webpack

Kelleretoro:
## Tests

1. TODO


Kelleretoro (commented):
ok

Kelleretoro (commented):
ok

PR #3506, opened 2026-05-07
fix: matching on data routes when middleware matcher is using allowed paths

pieh:
## Tests

Added fixture and test for cases when "allow list" paths are used for middleware/proxy matching


src/build/functions/edge.ts
```diff
@@ -80,13 +93,18 @@ const augmentMatchers = (
...
+                .replace(
+                  '(?:\\/((?!_next\\/)[^/.]{1,}))',
+                  `(?:\\/((?!_next\\/)(${i18NConfig.locales.join('|')}){1,}))?`,
+                ),
             }
           : matcher,
-        {
-          ...matcher,
-          regexp: pathToRegexp(matcher.originalSource).source,
-        },
```
pieh, line 89: This was only there for the case when i18n locale prefix was not present in request. But `pathToRegexp` approach was not matching on data routes.

So instead change above that marks locale part in originally produced regexp as option takes care of what this tried to do

serhalp (approved):

PR #3510, opened 2026-05-10
fix: mark prerendered content as less stale now that expire setting is enforced for marking for blocking re-render

pieh:
## Description

Adjust route-level cache handling after https://github.com/vercel/next.js/pull/93211

Main things is that "expire" timing is now enforced. We've been doing trick if using `lastModfied` "a year ago" (for prerendered content) or `lastModified: -1` (later in runtime) to make Next.js treat cache entries as stale, but now this no longer works for `next@canary` and upcoming `16.3` version and we need to calculate or fake things to correctly mark as stale when we should.

## Tests

No test changes. This PR fixes currently failing next@canary tests (for example see failures in https://github.com/opennextjs/opennextjs-netlify/actions/runs/27339057998/job/80770835521 when tests did run against `main` branch)


Kelleretoro (approved):

Kelleretoro (approved):

Kelleretoro (approved):

serhalp (approved):
LGTM

PR #3512, opened 2026-05-11
fix: vary RSC responses by auth inputs

dunfe:
## Summary

- Adds request cookie names to `Netlify-Vary` for RSC/flight requests so cacheable RSC payloads vary by session/auth cookies instead of sharing one cache object across users.
- Adds `authorization` to the `Netlify-Vary` header list for RSC requests that include an `Authorization` header.
- Adds unit coverage for authenticated RSC vary behavior.

## Why

RSC/flight payloads can contain user-specific data when an App Router page reads `cookies()`/auth state. If a page or middleware makes that RSC response cacheable, the existing default `Netlify-Vary` only varies on Next/RSC routing headers and preview cookies. That can allow a cached `text/x-component` payload for one authenticated user to be reused for another user with the same RSC route/cache key.

This is a defensive hardening layer: for RSC requests that arrive with cookies or authorization, include those auth inputs in the Netlify cache key.

## Test plan

- `npm run test:unit -- src/run/headers.test.ts`
- `npm run lint -- --no-cache src/run/headers.ts src/run/headers.test.ts`
- `npm run typecheck`
- `npm run build`

### DIFF
--- a/src/run/headers.test.ts
+++ b/src/run/headers.test.ts
@@ -60,6 +60,42 @@ describe('headers', () => {
         )
       })
 
+      test('with RSC request cookies added to avoid sharing user-specific flight data', () => {
+        const headers = new Headers()
+        const request = new Request(defaultUrl, {
+          headers: {
+            RSC: '1',
+            Cookie: 'session=user-a; theme=dark',
+          },
+        })
+        vi.spyOn(headers, 'set')
+
+        setVaryHeaders(headers, request, defaultConfig)
+
+        expect(headers.set).toBeCalledWith(
+          'netlify-vary',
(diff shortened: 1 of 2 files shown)


pieh:
Can you describe scenario where authenticated requests would get cached?

AFAIK of you try to make use of request headers / cookies it should opt response into not cacheable response ( https://nextjs.org/docs/app/glossary#request-time-apis#request-time-apis )

dunfe:
Closing this PR because the issue is specific to the legacy Netlify Next Runtime v4 RSC data routing path. Runtime v5 does not reproduce in the controlled repro, so I moved the patch to the v4 runtime repository instead.

PR #3513, opened 2026-05-11
fix(v4): skip RSC data rewrites for authenticated requests

dunfe:
## Summary

- Updates the legacy Runtime v4 `rsc-data` edge router to skip `.rsc` data-route rewrites for RSC requests that include `Cookie` or `Authorization` headers.
- Keeps anonymous/static RSC requests on the existing rewrite path.
- Adds Deno tests for anonymous rewrite behavior and authenticated bypass behavior.

## Why

Runtime v4 generates an `rsc-data` edge function that rewrites App Router RSC requests matching prerendered static/dynamic routes to `.rsc` data routes. A broad App Router SSG catch-all route can match a concrete dynamic authenticated page such as `/dashboard`, causing an authenticated RSC request to be rewritten to a cacheable `.rsc` path.

If that RSC payload contains user-specific data and is later cached by the CDN/ODB path, subsequent users can receive the first user's `text/x-component` payload. Authenticated RSC requests should therefore continue to the origin route instead of being rewritten to static RSC data routes.

Runtime v5 does not reproduce this issue in the controlled repro; this patch is scoped to the legacy Runtime v4 RSC data-routing path.

## Demo repro

Patched Runtime v4 deployable demo repo:

https://github.com/dunfe/netlify-next-v4-patched-cache-repro

This repo keeps the same vulnerable app shape as the original repro:

- `next@14.2.35`
- App Router SSG catch-all: `● /[...slug]`
- Dynamic authenticated dashboard: `ƒ /dashboard`
- `SIMULATE_CONTENTFUL_FAILURE=1`
- Middleware still sets cacheable headers for `/dashboard`

The only intended change is that it installs a locally packed patched Runtime v4 plugin tarball:

```json
"@netlify/plugin-nextjs": "file:vendor/netlify-plugin-nextjs-4.41.6-rsc-auth-rewrite-fix.tgz"
```

Local verification for the demo repo:

```txt
Using Next.js Runtime - v4.41.6

Functions:
- ___netlify-handler
- ___netlify-odb-handler
- _ipx

Edge Functions:
- next_middleware
- rsc-data
```

The generated Runtime v4 `rsc-data` edge function contains the patch:

```ts
if (request.headers.has('cookie') || request.headers.has('authorization')) {
  return
}
```

Expected behavior for the patched demo:

- User A's RSC request to `/dashboard` returns `User A`.
- User B's RSC request to `/dashboard` returns `User B`.
- User B should not receive User A's cached `text/x-component` payload.

Original vulnerable repro and root-cause report:

https://github.com/dunfe/netlify-next-first-user-cache-repro

### DIFF
--- a/packages/runtime/src/templates/edge-shared/rsc-data.test.ts
+++ b/packages/runtime/src/templates/edge-shared/rsc-data.test.ts
@@ -1,0 +1,88 @@
+import { assertEquals } from 'https://deno.land/std@0.167.0/testing/asserts.ts'
+import { describe, it } from 'https://deno.land/std@0.167.0/testing/bdd.ts'
+import { getRscDataRouter, PrerenderManifest } from './rsc-data.ts'
+
+const manifest: PrerenderManifest = {
+  version: 3,
+  routes: {},
+  dynamicRoutes: {
+    '/[...slug]': {
+      routeRegex: '^/(.+?)(?:/)?$',
+      fallback: null,
+      dataRoute: '/[...slug].rsc',
+      dataRouteRegex: '^/(.+?)\\.rsc$',
+    },
+  },
+  notFoundRoutes: [],
+}
+
+const createContext = () => {
+  let rewrittenTo: string | undefined
+
+  return {
+    context: {
+      rewrite: (target: string) => {
+        rewrittenTo = target
+        return new Response('rewritten')
+      },
+    },
+    get rewrittenTo() {
+      return rewrittenTo
+    },
+  }
+}
+
+describe('getRscDataRouter', () => {
+  it('rewrites anonymous RSC requests that match prerendered dynamic routes', () => {
+    const router = getRscDataRouter(manifest)
+    const rewriteContext = createContext()
+
+    const response = router(
+      new Request('https://example.netlify.app/dashboard', {
+        headers: {
+          RSC: '1',
+        },
+      }),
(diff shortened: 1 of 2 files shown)


pieh:
I'm not sure if this is actually properly fixable in V4. Issues like this is one of the reasons we did complete rework of our Next.js integration into how it's handled in V5, because approach that V4 was taking was basically unmaintable for the way App Router works (it was mostly fine for Pages Router where it's very clear at build time which routes should be cacheable and which should not be). 

Using existance of cookies as escape hatch from routing to cacheable might work (it's been a while so don't remember exact details on ODB vs SSR handlers in v4, which might require more changes than that), but it also mean it would opt out of cdn cachingroutes that might read cookies in the browser and not server so this would make handling worse for that case. Any setups with client analytics could no longer cache on CDN even if cookies are used for unrelated to auth reasons. So this doesn't seems like correct approach to me and it basically goes back to setup in V4 being basically incompatible with App Router.

We also don't apply fixes and adjustments to v4 anymore other than ensuring it can keep working with the way it works as-is on our platform for legacy applications, so it's rather unlikely that we would merge anything that changes behaviors of it.

dunfe:
Thanks for the detailed explanation.
I agree this PR is probably not the right direction. I’ll close this PR and treat V5 as the correct path forward for this class of issues.

PR #3550, opened 2026-07-27
test: adjust assertion version gates

pieh:
## Description

Follow up to https://github.com/opennextjs/opennextjs-netlify/pull/3547 - version gate was not correct as Next.js was changing logic over-time there and was not caught earlier ( see https://github.com/opennextjs/opennextjs-netlify/pull/3548 failing on older Next.js versions)

## Tests

1. TODO


hrishikesh-k (approved):

raquelpn1979-tech:
Svelte específicamente en este proyecto

PR #3553, opened 2026-07-29
test: add the adapter e2e suite workflow

hrishikesh-k:
Adds CI infrastructure (.github/workflows/adapter-e2e.yml, scripts/e2e-*.sh) to run the adapter
e2e suite against Next.js built from source, including forks carrying test/harness changes not
yet submitted upstream. This surfaced the fixes below, each covered by a new unit test.

fix: don't crash on server actions over the body size limit

feat: add `_next/static/service-worker/*` header handling

fix: keep non-ascii blob keys consistent between build and runtime

fix: await the Node middleware require in the generated edge handler template


src/build/plugin-context.ts
```diff
@@ -473,6 +473,10 @@ export class PluginContext {
     if (!this.#shells) {
       this.#shells = Object.entries(prerenderManifest.dynamicRoutes).reduce(
         (shells, [route, meta]) => {
+          // `meta.fallbackSourceRoute` marks routes skipped during export because they have
+          // unpopulated fallback root params. They exist only so Next.js can match requests and
+          // reuse the source route's cached shell, so no `.html`/`.meta` files are written for
+          // them. Only `meta.fallback` represents a real shell we should read here.
```
pieh, line 479: https://github.com/opennextjs/opennextjs-netlify/pull/3553/changes/52804468ae26706ae4906ba693dfe937267b968f The change I did there that was reverted in that commit might not be correct, but it was seemingly fixing some issue in Next e2e tests - the fact it was fixing it might have been accidental and the fix was not actually correct.

I'll try to dig into history to find what the issue (or at least next.js e2e test) was it.

I think it's fine to not have it here (especially if it was breaking something else)

.github/workflows/adapter-e2e.yml
```diff
@@ -0,0 +1,388 @@
...
+
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 25
+          path: nextjs
+          # `inputs` is empty on a push trigger, so both fall back to their
+          # dispatch defaults. Swapping these two literals is how you point a
+          # branch at a fork without touching anything else.
+          repository: ${{ inputs.nextjsRepo || 'pieh/next.js' }}
+          ref: ${{ inputs.nextjsRef || 'e2e-tests-adjustments' }}
```
pieh: I used my fork to test my not yet PRed next.js repo adjustments, but for things that would land in `main` - probably we want to point actual Next.js repo (and `canary` branch) by default

.github/workflows/adapter-e2e.yml
```diff
@@ -0,0 +1,388 @@
+# reference: https://nextjs.org/docs/app/api-reference/adapters/testing-adapters
+
+name: test-e2e-deploy
+
+on:
+  push:
```
pieh: Not sure if we want to run those on all pushes - was good trigger when iterating on changes in PR-less branch (because workflow_dispatch wouldn't be possible until workflow is in `main`). But for now I would probably drop the `push` and keep just `workflow_dispatch`?

Maybe later use `schedule` as well, but probably not yet?

.github/workflows/adapter-e2e.yml
```diff
@@ -0,0 +1,388 @@
...
+    # Compiling next-swc from source dominates this job: ~25-40min cold, a few
+    # minutes once the cargo cache is warm.
+    timeout-minutes: 90
+    env:
+      # Whether to test next.js from this checkout instead of the published npm
+      # release. `inputs` is empty on a push, and a plain `inputs.buildNextjs ||
+      # 'true'` fallback would be wrong here: an explicitly UNCHECKED box is
+      # `false`, which is falsy, so it would silently build anyway. Hence the
+      # event check — dispatch obeys the box, everything else builds.
+      BUILD_NEXTJS: ${{ github.event_name != 'workflow_dispatch' || inputs.buildNextjs }}
```
pieh: forcing Next.js build (instead using whatever is published to npm) was mainly to be able to test my unshipped Next.js changes - for workflow_dispatch it probably make sense to add a toggle and not default to always building from source - it takes some extra time to do that (both rust/native turbopack + node.js package for main `next` package) that I think _could_ bo omitted if not required (unless we would want to force that due to quirks with some package not being actually published either because of unlucky timing OR canary versions not actually being published)

pieh:
https://github.com/opennextjs/opennextjs-netlify/blob/main/.github/workflows/test-e2e.yml (our "old" next.js e2e test setup) need to figure out what to do with this one - it's no longer working - so at very least we should at least disable the workflow to not waste github actions on it. 

There are some some things that are not captured in new workflow yet - like collecting data for our proper e2e report. Not sure if we will want to reuse it as-is, so just flagging that some thought will need to be put what to do about it.

Probably we should not touch it in this PR and have it handled separately to not blow scope of this one.

pieh:
We should make sure to use https://github.com/googleapis/release-please#what-if-my-pr-contains-multiple-fixes-or-features or https://github.com/googleapis/release-please#how-can-i-fix-release-notes so we have separate entries for each of those fixes in this PR

Also we probably should not have a note about adapter e2e and our release - this part is more of a `ci`/`chore`/`test` than something that should land in changelog / release notes of our package

pieh:
> * fix: only add the `_next/static/service-worker/*` header rule when a build actually has a service worker

this is technically a fix against what was only in this branch - we currently don't have any of that in `main`, so probably should be described more as something about adding `_next/static/service-worker/*` handling

> fix: only treat PPR routes with their own generated shell as shells

related to https://github.com/opennextjs/opennextjs-netlify/pull/3553#discussion_r3674528977 - currently this PR doesn't actually change anything vs handling we have in `main` (my change was reverted and just comment was added which is fine) - so probably we don't want to actually have note about this (and if some change is still actually needed - we can do this as standalone change after this lands)

.github/workflows/adapter-e2e.yml
hrishikesh-k: Dropped `push`, `workflow_dispatch` only now

.github/workflows/adapter-e2e.yml
hrishikesh-k: With `push` gone, the existing `buildNextjs` toggle (already defaulting to `false`) is now the only path, so dispatch runs no longer default to building from source

.github/workflows/adapter-e2e.yml
hrishikesh-k: Dropped the push-trigger-only fallback to `pieh/next.js`/`e2e-tests-adjustments` entirely (since `push` is gone - see below), so the only path left is the `workflow_dispatch` inputs, which already default to `vercel/next.js`/`canary`.

hrishikesh-k:
Thanks! I have actioned all that needed an action (I hope 😅). Let me know if I missed something or did something incorrectly!

pieh:
I think the `routesManifest` read for the service worker headers _might_ be breaking static exports handling 🤔 
 - https://github.com/opennextjs/opennextjs-netlify/actions/runs/30458463207/job/90598101369?pr=3553#step:14:1248
 - (not sure if it's same reason, but looks probable) https://github.com/opennextjs/opennextjs-netlify/actions/runs/30458463207/job/90598101502?pr=3553#step:14:281
 
The question is wether it would be correct to skip trying to read that manifest for static exports? If that feature can't be used there - than probably that's easiest. But if it's expected that this feature should work for static exports - we might need to switch to checking if `.next/static/service-worker` exist instead of using manifest?

hrishikesh-k:
Thanks for catching that 😅 you were right. I have now switched to reading the dir instead.

pieh (approved):
🙏

PR #3574, opened 2026-09-15
fix: preserve initURL

pieh:
## Description

When middleware rewrites a request, the edge function forwards the rewrite target as the request URL to the server handler (that URL is the CDN cache key). Next.js derives its `initURL` request meta from `req.url`, so anything built from `initURL` ended up with the internal rewrite target instead of the URL the client requested:

- Next.js 16.3 enables `experimental.validateRSCRequestHeaders` by default on stable builds (vercel/next.js#93367). With it, RSC requests that lack a matching `_rsc` param get a 307 whose `Location` is built from `initURL`. For a page behind a middleware rewrite the browser was redirected to the internal path if `?_rsc` param was lost along the way, which then showed up in the address bar (and 404'd on reload when middleware guards the internal paths). Setting the flag to `false` avoids the redirect but drops the additional RSC cache-poisoning protection, so it isn't a fix (we should be generally good with cache poisoning because we should vary on all relevant headers, but it relies on implementation details that might change in future next versions, so keeping additional layer of protection seems like a good idea).
- `request.url` in Route Handlers and other `initURL` consumers (server action redirects, `req.url`-derived URLs in pages router) had the same problem on all versions.

Fix:

- The edge function sets `x-next-public-url` to the original request URL when it rewrites.
- The server handler pre-populates `initURL` from that header as a locked accessor, since Next.js assigns `initURL` from `req.url` unconditionally in several places. Only path and query are taken from the header, the origin is always the real request's, so a client-supplied header can't do anything beyond requesting that path directly.
- `Netlify-Vary` includes `x-next-public-url` for responses not served from Next.js' own cache, so dynamic responses that depend on the public URL aren't shared between different public URLs rewriting to the same target. Cached (`x-nextjs-cache`) responses are rendered without the request and keep being shared.

## Tests

Added test case with example of setup with redirect losing `?_rsc` param to a location that middleware rewrites (from public to internal path). That hits the case of rsc validation that produce yet another redirect (and this is the redirect that needs to keep original pre-middleware rewrite location)


edge-runtime/lib/response.ts
```diff
@@ -227,6 +227,9 @@ export const buildResponse = async ({
     }
     edgeResponse.headers.set('x-middleware-rewrite', relativeUrl)
     request.headers.set('x-middleware-rewrite', target)
+    // the server handler receives the rewrite target as its URL (that's the CDN cache key), so it
+    // needs the URL the client actually requested to be able to produce public-facing URLs
+    request.headers.set('x-next-public-url', request.url)
```
hrishikesh-k: Is this client-spoofable by any chance? 🤔

src/run/handlers/server.ts
```diff
@@ -58,6 +58,48 @@ const disableFaultyTransferEncodingHandling = (res: ComputeJsOutgoingMessage) =>
...
+/**
+ * When the edge middleware rewrites, the function receives the rewrite target as its URL and the
+ * URL the client requested in a header. Next.js keeps the latter as the `initURL` request meta
+ * (public-facing redirects, `request.url` in Route Handlers, ...) and sets it from `req.url`
+ * unconditionally in more than one place, so a plain pre-set would be overwritten. A locked
+ * accessor survives those writes. Only the path and query are taken from the header, the origin is
+ * always the one of the actual request, so a client sending the header can't gain anything it
+ * couldn't already do by requesting that path directly.
+ */
+const lockInitURL = (req: object, request: Request) => {
```
hrishikesh-k: This along with the above is potentially trusting the header whenever it's present, without checking whether a rewrite actually happened. Not sure if that's a real surface area for any issue though.

edge-runtime/lib/response.ts
pieh: I now attach request id together with public url in single header and only trust public url if request id in that header and one that lambda sees matches. request id is provided by platform so this should ensure anti-tampering

hrishikesh-k (approved):

